Issue #1, opened 2021-08-24
Add postgres integration to backend with 'go-pg'


jgwest:
- Add https://github.com/go-pg/pg to go.mod of 'backend'
- Next, write a function that uses go-pg to connect to our database (see above URL for examples of how to do this)

The backend folder is VERY barebones at the moment... it's just a very simple module based on a tiny example from the go-restful project, just enough for us to get started. Feel free to create new Go packages,  move stuff around in there, etc.

samyak-jn:
Hey Hi, I'll give this a try, and hence assigning this issue to myself! :taco:

Issue #3, opened 2021-08-24
Add go linting and gosec to project


jgwest:
Add go linting and gosec to project (to backend and cluster agent)

See: https://github.com/securego/gosec

And see:
- https://github.com/argoproj/argo-cd/blob/c7e532941fc0c75b10d8e3a2cf2372941a3c1edc/Makefile#L317

PR #15, opened 2021-09-03
Adding Postgres Integration with `go-pg` and Minor Refactoring (solves #1)

samyak-jn:
Hi, :hugs: 

This is a WIP PR in regards to solve issue #1 

The code is in the very initial stage, might need a lot of refactoring in the later on versions, for now, this PR does Postgres integration with `go-pg` along with the basic refactoring in such a manner that:

All the config files (such as Postgres integration et al) for in config/ dir, and while the API routes endpoints go in routes/ dir.
```
                        backend
                        ├── config
                        │   └── postgres-integration.go
                        └── routes
                            └── restful-basic-authentication.go
```


backend/config/postgres-integration.go
```diff
@@ -0,0 +1,44 @@
...
+// function to connect to Postgres
+func Connect2db() *pg.DB {
+	fmt.Println("Connection Initiated")
+	opts := &pg.Options{
+		Network:               "",
+		Addr:                  "localhost:5432",
+		User:                  "postgres",
+		Password:              "gitops",
+		Database:              "managed-gitops-postgres",
+		ApplicationName:       "",
```
jgwest, line 20: My guess is that we can remove these lines:
```go
		Network:               "",
```
and
```go
		ApplicationName:       "",
		DialTimeout:           0,
		ReadTimeout:           0,
		WriteTimeout:          0,
		MaxRetries:            0,
		RetryStatementTimeout: false,
		MinRetryBackoff:       0,
		MaxRetryBackoff:       0,
		PoolSize:              0,
		MinIdleConns:          0,
		MaxConnAge:            0,
		PoolTimeout:           0,
		IdleTimeout:           0,
		IdleCheckFrequency:    0,
```

backend/config/postgres-integration.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/go-pg/pg/v10"
+)
+
+// function to connect to Postgres
+func Connect2db() *pg.DB {
```
jgwest, line 12: I think this can change this to `connectToDatabase()`, as numbers are not normally used as verbs in methods names in Go.

jgwest (changes requested):
Looks great, thanks @samyak-jn 

A couple tweaks requested, you may also need to do this:
- `make fmt` (to run go fmt)
- `go mod tidy` (to tidy up the mod files)
With #3 I've added GitHub actions (based on what I'm using in the ApplicationSet repo) to check that these are now correct.

backend/config/postgres-integration.go
samyak-jn: Yes, it would be clean to have this removed if it was not required! :hugs: 
Done, done!

jgwest (approved):
LGTM!

Issue #4, opened 2021-08-24
Add REST endpoints to Go backend


jgwest:
- Implement the REST endpoints from https://docs.google.com/document/d/1e1UwCbwK-Ew5ODWedqp_jZmhiZzYWaxEvIL-tqebMzo/edit#heading=h.mzse2565h9bp using `go-restful`
- See  https://github.com/emicklei/go-restful for examples

This work includes defining structs for the JSON values returned by the endpoints:
```go

package api

// struct corresponding to the JSON respresentation of the User
type User struct {
	// JSON representation of fields from DB
}

// struct corresponding to the JSON representation of the Application
type Application {

	// JSON representation of fields from DB
	
	// name
	// spec contents
	// etc
	
	// (same fields as DB)
}

// etc

```

In order to verify the endpoints are working as expected, [Postman](https://www.postman.com/product/rest-client/) or [a similar utility](https://alternativeto.net/software/postman/?license=opensource) can be used.

samyak-jn:
Hey hi, I would like to give this issue a try, assigning myself! :taco: 
Please let me know in case of any conflict :hugs:

jgwest:
Done :tada:

PR #18, opened 2021-09-09
Adding REST routes to the go backend (Solves: #4)

samyak-jn:
Hello! :hugs:

This is a WIP PR in regards to solve issue #4 

For now, this pull request is in work in progress, and has commits regarding the endpoints of the operation module only. The implementation for the managed environments and applications needs to be done.
Along with the code, the tests are also delivered with the respective modules :test_tube:


jgwest:
@samyak-jn 

Following up from today's discussion, here's some pseudocode for the go structs we might use for managed services:
```go
package routes

type ManagedEnvironmentListResponse struct {
	Entries []ManagedEnvironmentListEntry `json:"entries"`
}

type ManagedEnvironmentListEntry struct {
	Name             string `json:"name"`
	URL              string `json:"url"`
	ConnectionStatus string `json:"connectionStatus"`
}

// Handle the GET to /api/v1/managedenvironment
// This is used by the 'List Managed Application Environments' UI and/or CLI
func handleListManagedEnvironments() ManagedEnvironmentListResponse {
	ret := ManagedEnvironmentListResponse{
		Entries: []ManagedEnvironmentListEntry{
			// With another isssue, we would fill this in with data from the database,
			// but for now can just leave it blank, or use fake data.
		},
	}
	// (...)
	return ret
}

type ManagedEnvironmentSingleEntry struct {
	Name              string `json:"name"`
	URL               string `json:"url"`
	KubeConfig        string `json:"config"`
	KubeConfigContext string `json:"configContext"`
	ConnectionStatus  string `json:"connectionStatus"`
}

// handle the GET to /api/v1/managedenvironment/{id}
func handleGetASpecificManagementEnvironment(id string) ManagedEnvironmentSingleEntry {
	ret := ManagedEnvironmentSingleEntry{
		// Normally we would read data from the database, and then insert it into this
		// struct, but for now just leave it blank or use fake data.
		// (...)
	}
	return ret
}

type ManagedEnvironmentPostEntry struct {
	Name              string `json:"name"`
	URL               string `json:"url"`
	KubeConfig        string `json:"config"`
	KubeConfigContext string `json:"configContext"`
}

// handle the POST to /api/v1/managedenvironment
func handlePostManagedEnvironment(request ManagedEnvironmentPostEntry) string {
	// We would store the 'request' object from the user in the database,
	// then return the UUID of the operation.
	// But for now, just leave it blank.

	return ""
}
```

samyak-jn:
You are awesome @jgwest! :hugs: 
This is so helpful, thank you! :tada:

samyak-jn:
Hi,
As per our discussion, this PR has been raised, handles basic architecture for creating routes!
Furthermore, in upcoming issues, once we know the functionality as well, we might have better tests written for the same!

Thank you for all the help :100:

backend/routes/application/application.go
```diff
@@ -0,0 +1,97 @@
...
+	"net/http"
+	"time"
+
+	restful "github.com/emicklei/go-restful/v3"
+)
+
+// Creating a REST layer for application - ApplicationResource
+
+type Application struct {
+	Id, Name string
```
jgwest, line 25: You can add json tags to these fields (Id, Name), like this:
https://github.com/argoproj-labs/applicationset/blob/0edd1527bb73fa3c8e8578a1a6124e71412f5403/api/v1alpha1/applicationset_types.go#L45

This will allow `json.Marshal` to convert this struct to/from a JSON string.

(And likewise with the other resources)

backend/routes/application/application.go
```diff
@@ -0,0 +1,97 @@
...
+		log.Fatal(err)
+	}
+}
+
+// GET info of operations depening upon the id
+func (a ApplicationResource) findApplication(request *restful.Request, response *restful.Response) {
+	id := request.PathParameter("application-id")
+	app := a.applications[id]
+	if len(app.Id) == 0 {
+		response.AddHeader("Content-Type", "text/plain")
```
jgwest, line 63: We should also return error code 404 here, if the application isn't found. This is the standard HTTP error code that communicates that the resource is not available (with REST, the vast majority of errors are communicated via [HTTP codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)).

backend/routes/application/application.go
samyak-jn: Yes, totally makes sense, corrected at relevant implementations!

backend/routes/application/application.go
samyak-jn: Yes, but if you check here at [L64](https://github.com/samyak-jn/managed-gitops/blob/d590aabc24810b13a0e03af2bb332b49ac7bc6ad/backend/routes/application/application.go#L64), there is an implementation of writing to an Error String which is returning `http.StatusNotFound` which is basically is

> const http.StatusNotFound untyped int = ### 404

Do you want something else, apart from this?

jgwest: Ah :sweat_smile:,  missed that, thanks!

backend/routes/application/application.go
```diff
@@ -0,0 +1,97 @@
...
+	wsContainer.Router(restful.CurlyRouter{})
+	a := ApplicationResource{map[string]Application{}}
+	a.Register(wsContainer)
+
+	log.Print("The server is up, and listening to port 8090 on your host.")
+	server := &http.Server{Addr: ":8090", Handler: wsContainer}
+	log.Fatal(server.ListenAndServe())
+}
+
+func waitForServerUp(serverURL string) error {
```
jgwest, line 89: In order to reduce code duplication, we should move the `waitForServerUp` function into a test package (something like `(root)/util/test`), and there should only be one copy of it within the codebase (it appears it is currently duplicated in multiple different files).

backend/routes/application/application.go
```diff
@@ -0,0 +1,97 @@
...
+			log.Fatal(err)
+		}
+	}
+}
+
+// Add function to start up the server, running against dedicated port
+// Usage of CurlyRouter is done because of the efficiency while using wildcards and expressions
+func RunRestfulCurlyRouterServer() {
+	wsContainer := restful.NewContainer()
+	wsContainer.Router(restful.CurlyRouter{})
```
jgwest, line 80: I think we should update the startup logic such that all of the resources should be initialized by, and started from, a single function, for example:
```go
func RunRestfulCurlyRouterServer() {
	wsContainer := restful.NewContainer()
	wsContainer.Router(restful.CurlyRouter{})
	
	// Register first resource
	a.Register(ApplicationResource{map[string]Application{}})

	// Register second resource
	o := OperationResource{map[string]Operation{}}
	o.Register(wsContainer)
	
	// Register nth resource
	// (...)

	
	log.Print("The server is up, and listening to port 8090 on your host.")
	server := &http.Server{Addr: ":8090", Handler: wsContainer}
	log.Fatal(server.ListenAndServe())
}
```	
Eg rather than having each resource have its own server listen function.

Then, this function should be called from `main.go` at the root of the application, so that it can serve requests for all API routes on startup.

jgwest (changes requested):
LGTM, a few more tweaks requested and then good to go.

backend/routes/application/application.go
samyak-jn: Absolutely, by having all at one place (say main.go) it will synchonise as we will be having one server to listen to! I'll make the necessary changes! :)

samyak-jn:
Hi, the relevant changes have been made, I have one request though, can we not squash and merge this PR, all the commits defined are having a detailed commit message of what it is doing, and is very easy to cherry-pick or open something up to see what it is doing instead of having all as one! Just a suggestion from my end!

jgwest (approved):
Looks great, thanks @samyak-jn!

Issue #9, opened 2021-08-24
Implement 'List Managed Clusters' page (w/o backend connection)


jgwest:
Using PatternFly and React, implement this page in the `frontend` component.
See [product UI page](https://docs.google.com/document/d/1w_TeKlW17U6DOCOtzN-0uXSJXwwuuj2CeWCOTim7wAI/edit) for description of what this page should contain.

Since the backend is not implemented yet, use fake data on the page as needed (rather than querying a backend). For example, on a page table containing a list of applications, populate the table with some fake application data.

jgwest:
:tada:

PR #24, opened 2021-09-28
Adds a page for List Managed Cluster Area (Solves #9)

samyak-jn:
Hello! :hugs:

This is a PR in regards to solve issue #9, it handles a UI that gives a very basic idea of how the front end for List Managed Cluster would look like, with the basic dummy data implementation!

Below is the short glimpse of how it looks like:

Screenshot from 2021-09-28 16-33-00

With filter for URL:
Screenshot from 2021-09-28 16-32-35


jgwest (approved):
Looks really nice, I think your React skills easily surpass mine :tada: :smile: . Thanks!

Issue #8, opened 2021-08-24
Implement Argo CD Load Test


jgwest:
In order to allow us to decide how much memory/CPU to provision to give to Argo CD, we need to create a load test that uses Argo CD to create a bunch of cluster resources.

Create a Go unit test that does the following:
-  Spins up 1 Argo CD instance (by following the Argo CD install instructions), then waits for it to be installed (all expected resources are running).
    - Probably the easiest way to spin up an Argo CD instance from inside Go is to call `kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml` using [golang's os Exec function](https://pkg.go.dev/os/exec).
- Next, it adds new Applications to Argo CD (by creating the `Application` CR within the Argo CD namespace)
    - For examples of interaction with Kubernetes client API, see what we do in ApplicationSet repo. Search for uses of `argov1alpha1.Application`
- This will require us to find real world applications that we can use to test this.  
    - See https://github.com/argoproj/argo-cd/issues/4392 for 3 examples. 
    - See https://github.com/redhat-cop/gitops-catalog for more examples (but say away from the ones that require OpenShift)
- Use large applications like the prometheus operator, not tiny applications like the guestbook. For an example of a large application, apply the following in the namespace:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: jgw-prometheus
  namespace: argocd
spec:
  destination:
    namespace: jgw
    server: https://kubernetes.default.svc
  project: default
  source:
    helm:
      version: v2
    path: appset-examples/cluster-addon/prometheus-operator
    repoURL: https://github.com/jgwest/argocd-example-apps.git
```
- Determine pass/validation criteria (what does it mean for a load test to pass? no pod restarts?), and implement within test itself
- For, use applications that do not require OpenShift (for example, no openshift-specific operators)

See https://github.com/redhat-appstudio/managed-gitops/tree/main/utilities/load-test for barebones project to get started from.

The test should be its own standalone module under `https://github.com/redhat-appstudio/managed-gitops/utilities/load-test`. Why? Because it is likely to be useful by other projects (gitops operator, argo cd upstream), and keeping it separate makes this easy. (At a future time we might move it into its own repo)

See here for my previous (outdated) thoughts on how this would look: https://issues.redhat.com/browse/GITOPS-895

PR #25, opened 2021-10-04
[WIP] Add Argo CD Load Test (Solves: #8)

samyak-jn:
Hello! :hugs:

This is a WIP PR in regards to solve issue #8 

For now, this pull request is in work in progress, and has commits consisting of relevant functions that are needed for the load test to show correct resource info which is a necessity for the tests to pass. Whilst, writing these tests, and exploring  about the same there were bits that was discussed to how this all validates and work. Once done, I'll provide a comment/link to verify those points for everyone!


samyak-jn:
The PR is still in WIP, but requesting for the review to go through all the used utility and tests.
For now, there are two tests integrated!
TEST 1: Deploy 100 guestbook in different namespace
TEST 2:  Deploy Prometheus (heavy application) + 20 Guestbook Application (small application)

ToDos:
Figuring out how helm works. And, finally create cert-manager and external-dns application object to integrate with TEST 3 (Deploy all heavy application in one go)

samyak-jn:
A basic snap of output of how the test runs, and what information it depicts! :t-rex: 

```
=== RUN   TestNGuestbook
Testing for 100 Guestbook applications


Pods Memory Resource before creation:

Pod Name                                Container Name                  Memory Usage (in Ki)
argocd-redis-5b6967fdfc-kbvks           redis                            11580
argocd-dex-server-5dc4df4967-pfbzt      dex                              32528
argocd-repo-server-5f9b5d9584-wcs2g     argocd-repo-server               98456
argocd-application-controller-0         argocd-application-controller    90940
argocd-server-86dcc9f88f-b64d6          argocd-server                    61236


Pods Memory Resource after Creation:

Pod Name                                Container Name                  Memory Usage (in Ki)
argocd-dex-server-5dc4df4967-pfbzt      dex                              32528
argocd-repo-server-5f9b5d9584-wcs2g     argocd-repo-server               98456
argocd-application-controller-0         argocd-application-controller    90940
argocd-server-86dcc9f88f-b64d6          argocd-server                    61236
argocd-redis-5b6967fdfc-kbvks           redis                            11580


Difference in the Pod Memory Usage (in Ki)

Pod Name                                Container Name                  Memory Usage (in Ki)
argocd-repo-server-5f9b5d9584-wcs2g     argocd-repo-server               0
argocd-application-controller-0         argocd-application-controller    0
argocd-server-86dcc9f88f-b64d6          argocd-server                    0
argocd-redis-5b6967fdfc-kbvks           redis                            0
argocd-dex-server-5dc4df4967-pfbzt      dex                              0
100 guestbook applications are created, deployed and deleted successfully!
--- PASS: TestNGuestbook (41.20s)
```

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -2,38 +2,355 @@ package loadtest
...
-	if err != nil {
-		t.Errorf("error, %v", err)
+	// Check if the Pod is getting refreshed or not
+	if reflect.DeepEqual(PodC, PodRestartcount(namespace)) != true {
+		t.Errorf("error, %v", "The Pods are getting refreshed!")
 		return
 	}
 
-	for _, app := range appList.Items {
+	fmt.Print("\n\nPods Memory Resource after Creation:\n\n")
```
jgwest, line 77: We should move the pod memory resource check after the `wait.Poll`, which I talk about below.

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -2,38 +2,355 @@ package loadtest
...
+	memoryPost := Get_pod_info(kubeconfig, master, namespace, podName)
+	PodInfoParse(memoryPost)
+
+	fmt.Print("\n\nDifference in the Pod Memory Usage (in Ki)\n\n")
+	delta := PodMemoryDiff(memoryInit, memoryPost)
+	PodInfoParse(delta)
+
+	// less use of generic sleep statements in automated tests
```
jgwest, line 85: I've found the example I wanted to give you, which is a way to avoid using sleep statements in tests:
https://github.com/redhat-developer/gitops-operator/blob/d9b4c4844b39f78683cd30525993e9af19929069/test/helper/application_status.go#L164

(this is a pattern I recommend using for all test code, as an alternative to use sleep statements)

Ultimately it looks like this:
```go
err := wait.Poll( /* time to wait between retries*', /* max time to wait */, func() (bool, err) 

    // return true if the thing we were waiting for has happened

    // if an error occurs, 'return false, error', which will end the loop.

    return false // this will keep the loop going, until true is returned

}) 
if err != nil {
   // fail the test
}
```

So `wait.Poll` will keep checking whether a particular condition is met, and will timeout with an error it is not. 

In this case, we probably want to wait for all the Argo CD applications to be processed by Argo CD, before check the pod memory usage.

In order to tell if Argo CD has finished processing the Application, we can look in the `(app variable).Status.Health.Status` field, which will be empty (`""`) if the Argo CD hasn't yet processed the Application resource.

So that might look like this:
```go
err := wait.Poll( /* time to wait between retries*', /* max time to wait */, func() (bool, err) {

// list all Argo CD applications in the namespace
    // for each application, check if (app variable).Status.Health.Status
        // if status is empty, return false

return true
})
```
So this will keep waiting until Argo CD has processed all the Applications  (and thus no need to use a sleep statement and guess how long it will take.) This leads to more reliable tests.

utilities/load-test/loadtest/utility.go
```diff
@@ -61,8 +70,106 @@ type E2EFixtureK8sClient struct {
...
+		panic(err)
+	}
+
+	if podName != "" {
+		// To get memory info of the specific pod passed as an argument
+		podMetrics, err := mc.MetricsV1beta1().PodMetricses(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+		if err != nil {
+			panic(err)
+		}
+		podResource[(podMetrics.ObjectMeta.Name + "\t" + podMetrics.Containers[0].Name)] = podMetrics.Containers[0].Usage["memory"].ToUnstructured()
```
jgwest, line 138: It looks like you are calling `ToUnstructed()` here;  if you instead _don't_ call `toUnstructured`, it means you shouldn't have to parse it as a string later.

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -2,38 +2,355 @@ package loadtest
...
+// 	// if err_delCertMan != nil {
+// 	// 	t.Errorf("error, %v", err_delCertMan)
+// 	// 	return
+// 	// }
+// 	fmt.Println("Prometheus, External DNS and Cert Manager applications are created, deployed and deleted successfully!")
+// }
+
+func TestAllApplication(t *testing.T) {
+
+	// TEST 3: Deply Prometheus (heavy application) + 20 Guestbook Application (small application)
```
jgwest, line 248: Same changes as in `TestNGuestbook`

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -2,38 +2,355 @@ package loadtest
...
+		_, err_create := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Create(context.TODO(), guestbookApp, v1.CreateOptions{})
+		if err_create != nil {
+			t.Errorf("error, %v", err_create)
+			return
+		}
+	}
 
-	if err != nil {
-		t.Errorf("error, %v", err)
+	// Check if the Pod is getting refreshed or not
```
jgwest, line 71: Can move this after the `wait.Poll`, which I talk about below.

utilities/load-test/loadtest/utility.go
```diff
@@ -61,8 +70,106 @@ type E2EFixtureK8sClient struct {
 	AppClientset     appclientset.Interface
 }
 
-// func CheckError(err error) {
-// 	if err != nil {
-// 		debug.PrintStack()
-// 	}
-// }
+// Kubectl_apply is basically a go utility that performs kubetcl apply -n namespace -f  *yaml
+func Kubectl_apply(namespace string, URL string) {
```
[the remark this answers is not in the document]
samyak-jn: Thanks for the pointing the CamelCase best practise scenario! :D
But, a quick question on exporting, do we not need to include the argocd manifest installation in main package for the function? If we do, we might need to export it, [here](https://github.com/samyak-jn/managed-gitops/blob/5ffdf9e7b6a38d248d5f241ac4c8564c22b0df2b/utilities/load-test/main.go#L9).

PR #31, opened 2021-10-26
Use red hat supported postgresql image

sbose78:
Hi,

Didn't mean to be disruptive :) I wanted to put in something that could eventually be moved into a Red Hat supported staging/prod.

* Removed the helm chart generated manifests since they were not OpenShift friendly ( was running as a specific user, was using a bitnami image)
* Replaced the manifests with those generated from supported OpenShift Templates - now using the `registry.redhat.io/rhel8/postgresql-10:latest` with the default restricted scc.

https://github.com/openshift/library/blob/master/official/postgresql/templates/postgresql-persistent.json

### DIFF
--- a/manifests/postgresql-staging/README.md
+++ b/manifests/postgresql-staging/README.md
@@ -1,21 +1,16 @@
-# PostgreSQL GitOps Staging YAML
+# PostgreSQL Database for GitOps 
 
-For Postgresql, there are a few ways to install it on Kubernetes, so we have some options for how we want to install it to the staging environment. We could use a PostgreSQL operator, Helm chart, or just flat YAML.
+## Support
 
-Here, I'm suggesting we just use plain YAML to install PostgreSQL, but for sake of simplicity that plain YAML file will be rendered from the Postgresql Helm Chart from Bitnami.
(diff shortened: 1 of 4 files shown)


jgwest (changes requested):
Thanks @sbose78?

I'm curious, how was this YAML generated (and/or what is it based on)? Is it an automated process (as with the original script), or just copied from somewhere else?

Rather than replacing the old content, can you create a new folder containing your new content? My concern is that what you have proposed will only work on OpenShift, and as developers we need the ability to develop on our local machines w/o OpenShift. Ideally we can adapt what you have provided to work for both cases.

sbose78:
I extracted this out of the official openshift Template - which hasn't changed in ages ( other than the image ).

I wouldn't suggest developers using anything other than openshift ( I know kind is easy :) ) especially because a lot of manifests which work on non-openshift kube do not work on openshift.

jgwest:
@sbose78 - I haven't forgotten this one :smile:, I've opened a story to track the tweaks we need to make to our manifests adopt this: https://issues.redhat.com/browse/GITOPSRVCE-98 , and we'll be starting work on it shortly.

jgwest:
We've completed our move to Red Hat supported PostgreSQL 12 image based on @sbose78 PR. Thanks @sbose78!

PR #32, opened 2021-10-28
Adds initial support for retrieving metrics from an Argo CD instance (Solves: Gitops-1389)

samyak-jn:
Hello! :hugs:

This PR is in regards to solving the issue [GITOPS-1389](https://issues.redhat.com/browse/GITOPS-1389).

### DIFF
--- a/cluster-agent/go.mod
+++ b/cluster-agent/go.mod
@@ -4,7 +4,6 @@ go 1.16
 
 require (
 	github.com/argoproj/argo-cd/v2 v2.1.1
-	github.com/jgwest/managed-gitops/cluster-agent v0.0.0-20211027023712-42af25b249bf
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.13.0
 	k8s.io/apimachinery v0.21.2
--- a/cluster-agent/go.mod
+++ b/cluster-agent/go.mod
@@ -1,9 +1,10 @@
-module github.com/jgwest/managed-gitops/cluster-agent
+module github.com/redhat-appstudio/managed-gitops/cluster-agent
 
 go 1.16
 
 require (
 	github.com/argoproj/argo-cd/v2 v2.1.1
+	github.com/jgwest/managed-gitops/cluster-agent v0.0.0-20211027023712-42af25b249bf
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.13.0
 	k8s.io/apimachinery v0.21.2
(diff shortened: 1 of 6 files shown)


cluster-agent/go.mod
```diff
@@ -1,9 +1,10 @@
-module github.com/jgwest/managed-gitops/cluster-agent
+module github.com/redhat-appstudio/managed-gitops/cluster-agent
 
 go 1.16
 
 require (
 	github.com/argoproj/argo-cd/v2 v2.1.1
+	github.com/jgwest/managed-gitops/cluster-agent v0.0.0-20211027023712-42af25b249bf
```
jgwest, line 7: I'm guessing this line can be removed now that you have renamed from the old module name.

cluster-agent/go.sum
```diff
@@ -475,6 +475,8 @@ github.com/itchyny/timefmt-go v0.1.2/go.mod h1:0osSSCQSASBJMsIZnhAaF1C2fCBTJZXrn
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 h1:BQSFePA1RWJOlocH6Fxy8MmwDt+yVQYULKfN0RoTN8A=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99/go.mod h1:1lJo3i6rXxKeerYnT8Nvf0QmHCRC1n8sfWVwXF2Frvo=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
+github.com/jgwest/managed-gitops/cluster-agent v0.0.0-20211027023712-42af25b249bf h1:vj07onsiXmPR2Tsd0IktyAJTG/bOgNOwgKW1ObkAMfY=
```
jgwest, line 478: Likewise guessing these will be removed when the `jgwest` referenced is removed from go.mod.

cluster-agent/main.go
```diff
@@ -112,4 +119,9 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+
+	//Retrieving ArgoCD metrics: Application Controller, API Server, Repo Server
```
jgwest, line 123: These three lines can be removed for now (we'll add code that calls these functions to the appropriate spot with later stories).

cluster-agent/utils/retrieve_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+// GET Request to retrive API Server metrics from the route endpoint
+func GetAPIServerMetrics(routeEndpoint string) string {
+	// #nosec G107
+	resp, err := http.Get(routeEndpoint)
+	if err != nil {
```
jgwest, line 13: For these 3 Get functions, you can return `(string, error)` from the func, and instead `return "", err` if the err is non-nil (rather than just logging it via `log.Fatalln`).

cluster-agent/utils/retrieve_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+		log.Fatalln(err)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	sb := string(body)
+	log.Printf(sb)
```
jgwest, line 21: Can remove the log statements that output to console, and just return the value below

jgwest (changes requested):
Thanks @samyak-jn , some minor tweaks than good to go.

cluster-agent/go.mod
samyak-jn: Yes, totally makes sense!

jgwest (approved):
LGTM, thanks @samyak-jn!

PR #34, opened 2021-11-17
Creates Webhook and recieve the payload on route endpoints in gitops-backend(Solves: Gitops-1509)

samyak-jn:
Hi :hugs: 

This pull request is in with regards to story GITOPS-1509, it consists of a function with helps to create a webhook, and one to parse it along with the endpoints which act as a payload URL for the Webhook!


backend/routes/webhooks/webhooks.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type WebHookInfo struct {
+	Id        string // Id for the github webhook request
+	Event     string // indicates which event took place (push, starred, pull request etc)
+	Signature string // signature of the github webhook request
+	Payload   []byte // consists of all the contents within the webhook
+}
+
+func ParseWebhookInfo(request *restful.Request, response *restful.Response) {
```
jgwest, line 22: We can add an `error` as the return value for this function, and convert all the current `log.Fatalln(...)` (and others) calls to `return fmt.Errorf(...)` calls.

backend/util/createHook.go
```diff
@@ -0,0 +1,58 @@
...
+	return token, nil
+}
+
+// Parameters:
+// 	personalAccessToken : This points to the personal access token (PAT) assigned from the GitHub with the repo access atleast
+// 	authorUsername 		: The author git username (a/c to the PAT)
+// 	authorName 			: The author full name registered (a/c to the PAT)
+// 	repoName 			: The repository name on which the hook will be defined
+// 	hookURL 			: This points to the payloadURL on which github will POST request
+func CreateWebHook(personalAccessToken string, authorUsername string, authorName string, repoName string, hookURL string) {
```
jgwest, line 28: Likewise can add an `(error)` return value, and return any errors seen

jgwest (changes requested):
Looks great, a couple tweaks and good to go.

backend/routes/webhooks/webhooks.go
samyak-jn: As per the research, go-restful route function won't allow us to have an error as a return value!

The struct for the restful.RouteFunction is defined as within the module:
`type RouteFunction func(*Request, *Response)`

It shows error on manipulating details, so that was the finding. If there is any other approach to this please let me know :D

jgwest (approved):
Looks great, thanks @samyak-jn! Happy to see that adding this was straightforward, due to `go-github` library (At the beginning I was worried we would have to implement a lot of the GitHub API calls ourselves :fearful: )

PR #38, opened 2021-11-24
Database Integration to Cluster Agent (Solves: GITOPS-1378)

samyak-jn:
Hi :hugs:

This pull request is in with regards to story GITOPS-1378, it consists of relevant Unsafe (create/update) function in for Application and ApplicationState. Along with update in reconciler function which is responsible for checking the following:
- If the entry exists at neither place, do nothing
- if the entry exists in the database but not in the namespace, create application in  the namespace
- if the entry exists not in the database but in the namespace, delete application from the namespace
- if exists at both places, then compare and update the changes in the namespace

This PR is in under WIP right now!

### DIFF
--- a/cluster-agent/controllers/argoproj.io/application_controller.go
+++ b/cluster-agent/controllers/argoproj.io/application_controller.go
@@ -190,24 +190,38 @@ func (r *ApplicationReconciler) createApplicationCRFromDB(ctx context.Context,
 	// - If ApplicationState row exists no change, else create!
 
 	dbApplicationState := &db.ApplicationState{Applicationstate_application_id: string(dbAppId)}
-
 	if err := databaseQueries.UncheckedGetApplicationStateById(ctx, dbApplicationState); err != nil {
-
 		// We expect not found error here, any other error should return
 		if !apierr.IsNotFound(err) {
 			log.Error(err, "unexpected error in createApplicationCR")
 			return err
 		}
 
+		// If the ApplicationState doesn't exist, create it
 		dbApplicationState := &db.ApplicationState{
 			Applicationstate_application_id: dbApplication.Application_id,
 			Health:                          string(application.Status.Health.Status),
 			Sync_Status:                     string(application.Status.Sync.Status),
 		}
 
-		errCreateApplicationState := databaseQueries.UncheckedCreateApplicationState(ctx, dbApplicationState)
-		if errCreateApplicationState != nil {
-			return errCreateApplicationState
+		if err := databaseQueries.UncheckedCreateApplicationState(ctx, dbApplicationState); err != nil {
+			log.Error(err, "unable to create ApplicationState")
+			return err
+		}
+
+		return nil
+	}
+
+	// If the status in the database is different from what is in the Application CR, then update the database
+	if dbApplicationState.Health != string(application.Status.Health.Status) ||
+		dbApplicationState.Sync_Status != string(application.Status.Sync.Status) {
+
+		dbApplicationState.Health = string(application.Status.Health.Status)
+		dbApplicationState.Sync_Status = string(application.Status.Sync.Status)
+
+		if err := databaseQueries.UncheckedUpdateApplicationState(ctx, dbApplicationState); err != nil {
+			log.Error(err, "unable to update ApplicationState")
+			return err
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


jgwest:
The Application database table entry needs a reference to a gitopsengineinstance (which is the Argo CD instance) and the managedenvironment (which is the target cluster to deploy the application k8s resources to). I've have committed a couple of new functions you can use for this:

### To get the managed environment database entry (go-like pseudocode):
```golang
argoCDNamespace := /* Get the Namespace the Application CR is in, usually argocd */
managedEnv, err := util.UncheckedGetOrCreateManagedEnvironmentByNamespaceUID(ctx, argoCDNamespace, dbq, log) 
```


### To get the GitOps engine entry (go-like pseudocode):
```golang
// Get cluster credentials if they already exist, or create them if they don't.
// This is a hack until we have proper cluster credential support (eg GITOPS-1564)
var clusterCreds *db.ClusterCredentials
{
	var clusterCredsList []db.ClusterCredentials
	if err := dbq.UnsafeListAllClusterCredentials(ctx, &clusterCredsList); err != nil {
		return err
	}
	

	// Next, go through clusterCredsList and look for credentials that contain Host: 'host'
	// - Set clusterCreds to match if they are found
	
	if clusterCreds == nil {
		clusterCreds = db.ClusterCredentials{
			Host:                        "host",
			Kube_config:                 "kube_config",
			Kube_config_context:         "kube_config_context",
			Serviceaccount_bearer_token: "serviceaccount_bearer_token",
			Serviceaccount_ns:           "serviceaccount_ns",
		}
		if err := dbq.CreateClusterCredentials(ctx, &clusterCreds); err != nil {
			return nil, fmt.Errorf("unable to create cluster creds for managed env: %v", err)
		}
	}
}

gitopsEngineNamespace := /* get the Namespace that the 'Application' CR is in */
kubesystemNamespaceUID := /* get the UID of the 'kube-system' namespace */

var gitopsEngineInstance *db.GitopsEngineInstance

gitopsEngineInstance := util.UncheckedGetOrCreateGitopsEngineInstanceByInstanceNamespaceUID(ctx, gitopsEngineNamespace, kubesystemNamespaceUID, clusterCreds, dbq, log)
```

jgwest (changes requested):
Here's some thoughts on the WIP PR so far, and I've added some code that should help, see [my comment here](https://github.com/redhat-appstudio/managed-gitops/pull/38#issuecomment-982217804). Thanks for your patience while waiting for me to put this code together, it was a fair bit of work :muscle: .

cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -41,16 +48,171 @@ type ApplicationReconciler struct {
...
+			}
+
+			application, errConvert := dbAppToApplicationCR(dbApplication)
+			if errConvert != nil {
+				return ctrl.Result{}, errConvert
+			}
+			errAddToNamespace := r.Create(ctx, application, &client.CreateOptions{})
+			if errAddToNamespace != nil {
+				return ctrl.Result{}, errAddToNamespace
+			}
```
samyak-jn, line 100: Proposal on how to check for the case when the application does not exists in the namespace but in db!

Fetch the list of application from the database and store it in an array!
Then, while taking app.Name as source of identification from the namespace, look in the provided array, the function contains will return a bool value with the Application ID, which will help us to create those application CRs in the namespace with the help of dbAppToApplicationCR method and client create functionality!

This chunk of code still need to be tested properly, but is syntactically correct. Open to suggestions if any.

Note: this is a hack to get this resolved meanwhile, will be researching more for a concrete solution in future!

jgwest:
Some code updates pushed to @samyak-jn's feature branch:
- Rebase to latest from main
- Refactor the controller (once the cluster-agent story is in place, it will be easier to test these changes; for now I have added a bool that can be used to enable/disable this logic)
- Some additional updates to database logic
- Add logic such that if status in the database is different from what is in the Application CR, then update the database

@samyak-jn, take a look and let me know if you see any obvious issues in any of the changes I've pushed.

samyak-jn (commented):
Hi,
@jgwest, LGTM, took a look at these, no blockers, and thank you for updating these out!

jgwest:
@samyak-jn  :sunglasses:

PR #47, opened 2021-12-13
Add tests to check the sanity for the database input methods

samyak-jn:
Hi :hugs:

This pull request is in with regards to story GITOPS-1375, which checks for inputting malicious values within the database objects!

### DIFF
--- a/backend-shared/config/db/injection_test.go
+++ b/backend-shared/config/db/injection_test.go
@@ -1,0 +1,271 @@
+package db
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGitopsEngineInstanceWrongInput(t *testing.T) {
+	dbq, err := NewUnsafePostgresDBQueries(true, true)
+	if !assert.NoError(t, err) {
+		return
+	}
+	defer dbq.CloseDatabase()
+
+	ctx := context.Background()
+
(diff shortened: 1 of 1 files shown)


backend-shared/config/db/injection_test.go
```diff
@@ -0,0 +1,262 @@
...
+		EngineCluster_id:        gitopsEngineCluster.Gitopsenginecluster_id,
+	}
+
+	clusterAccess := ClusterAccess{
+		Clusteraccess_user_id:                   testClusterUser.Clusteruser_id,
+		Clusteraccess_managed_environment_id:    managedEnvironment.Managedenvironment_id,
+		Clusteraccess_gitops_engine_instance_id: gitopsEngineInstance.Gitopsengineinstance_id,
+	}
+
+	if err = dbq.CreateClusterCredentials(ctx, &clusterCredentials); err != nil {
```
jgwest, line 219: These should all use this:
```
	if !assert.NoError(t, err) {
			return
		}
```
or at least `assert.NoError(t, err)`), even if this is already tested elsewhere.

jgwest (changes requested):
Looks good, minor tweak then good to go.

samyak-jn (commented):
Made the relevant changes (use of assert) where it was left, thanks!

jgwest (approved):
:sunglasses:

PR #51, opened 2022-01-10
Create Service Account on the remote cluster (GITOPS: 1564, 1601)

samyak-jn:
Hi :hugs:,

This PR is in regards with (GITOPS: 1564, 1601), which includes creating a service account on the remote cluster. This PR is also shipped with Unit test relevant to the functions created, for now we are testing on the local kubernetes cluster clientset. For the testing purpose, the hostname is taken as 'https://kubernetes.svc.default'.


backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	ArgoCDManagerServiceAccount     = "argocd-manager"
+	ArgoCDManagerClusterRole        = "argocd-manager-role"
+	ArgoCDManagerClusterRoleBinding = "argocd-manager-role-binding"
+)
+
+var (
+	hostname                          = "https://kubernetes.svc.default"
+	ArgoCDManagerNamespacePolicyRules = []rbacv1.PolicyRule{
```
drpaneas, line 28: Using capital letters you actually export the variable, so other packages can `import hack` pkg. Is this intentional?

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+		return false, nil
+	})
+	if err != nil {
+		return "", fmt.Errorf("Failed to wait for service account secret: %v", err)
+	}
+	token, ok := secret.Data["token"]
+	if !ok {
+		return "", fmt.Errorf("Secret %q for service account %q did not have a token", secret.Name, serviceAccount)
```
drpaneas, line 153: If you are actually going to create a unit test for checking this errors, it would make your life easier when writing unit-tests, so you can easily do an equality check for a specific error condition. For example:

```go
var secretErr = errors.New("secret err message blah-blah")
```

so you can easily test it:

```go
func someFunc() {
    secretMock.EXPECT().Save(gomock.Eq(&foooBarBar)).Return(nil, secretErr)
 }
```

Of course if you are _not_ going to test this, there is no point going that extra mile. I would just like to make sure you are aware of this, as you use `fmt.Errorf` in a lot's of places in your code.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// GetServiceAccountBearerToken will attempt to get the provided service account until it
+// exists, iterate the secrets associated with it looking for one of type
+// kubernetes.io/service-account-token, and return it's token if found.
+func GetServiceAccountBearerToken(clientset kubernetes.Interface, ns string, sa string) (string, error) {
+	var serviceAccount *corev1.ServiceAccount
+	var secret *corev1.Secret
+	var err error
+	err = wait.Poll(500*time.Millisecond, 30*time.Second, func() (bool, error) {
```
drpaneas, line 130: `500` is a magic number. A magic number is a numeric literal that is not defined as a constant, but which may change, and therefore can be hard to update later. If you think we are going to re-use this in other function as well, please create a constant.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+
+func generateClientFromClusterServiceAccount(hostname string, bearerToken string) client.Client {
+	config := &rest.Config{
+		Host:        hostname,
+		BearerToken: bearerToken,
+	}
+	clientObj, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Fatal(err)
+		return nil
```
drpaneas, line 166: Make sure that in all of your `return` statements there is a blank line _before_ them, to emphasise them.

drpaneas (commented):
I didn't go through testing this locally (I need some help enabling the debugger with that). I just made some general comments around your code from the aspect of readability, tech-dept and maintenance. 

PS: Most (if not all) of my suggestions are quite opinionated based from my Golang experience in the k8s project, so feel free to ignore them completely as there is no _right_ or _wrong_ here -- it's just a matter of personal preference mixed with coding experience.

Note: Before merging this PR, we need to first to merge https://github.com/redhat-appstudio/managed-gitops/pull/54 or the CI will break.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+		return "", fmt.Errorf("Failed to wait for service account secret: %v", err)
+	}
+	token, ok := secret.Data["token"]
+	if !ok {
+		return "", fmt.Errorf("Secret %q for service account %q did not have a token", secret.Name, serviceAccount)
+	}
+	return string(token), nil
+}
+
+func generateClientFromClusterServiceAccount(hostname string, bearerToken string) client.Client {
```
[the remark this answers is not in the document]
samyak-jn: Your statement seems very valid here, though it is currently not being used by another function, but I suppose this story adds utility which will be in turn used in later stories, it is intentionally left in the main code and not _test.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	_, err := clientset.CoreV1().ServiceAccounts(namespace).Create(context.Background(), &serviceAccount, metav1.CreateOptions{})
+	if err != nil {
+		if !apierr.IsAlreadyExists(err) {
+			return fmt.Errorf("Failed to create service account %q in namespace %q: %v", serviceAccountName, namespace, err)
+		}
+		return fmt.Errorf("ServiceAccount %q already exists in namespace %q", serviceAccountName, namespace)
+	}
+	log.Printf("ServiceAccount %q created in namespace %q", serviceAccountName, namespace)
+	return nil
```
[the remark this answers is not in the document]
samyak-jn: Aha, I'm very new to go, and go to know about this, thanks! :tada:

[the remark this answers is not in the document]
samyak-jn: Just btw do you use/know about any linters that suggest similar changes?

backend-shared/hack/service_account.go
samyak-jn: Makes sense, whenever we export it in the future, we can always make it capital.

backend-shared/hack/service_account.go
samyak-jn: I might be having basic knowledge in this area, so please correct me. I was using `fmt.Errorf` because of its ability to add errors with formatted strings, while when we use an extra method to format those strings `errors.New(Sprintf(""))`  hence less readability, can we not check for the above utility you mentioned? 
Because, both return the same type right(error)?

backend-shared/hack/service_account.go
samyak-jn: TIL, thanks! I don't think we might need it, I'll update it!

samyak-jn:
Thank you @drpaneas for an extensive review of my go skills and helping me out with correct formatting syntax (still learning :P), I got some good pointers and learning from here :D

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+			Kind:       "v1",
+			APIVersion: "ServiceAccount",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        serviceAccountName,
+			Namespace:   namespace,
+			ClusterName: "",
+		},
+	}
+	_, err := clientset.CoreV1().ServiceAccounts(namespace).Create(context.Background(), &serviceAccount, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
jgwest: I'm not sure I see the advantage of creating a new function for this, the fluent API of the clientset is pretty self-explanatory and self-contained as is.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+				Subjects: []rbacv1.Subject{{
+					Kind:      rbacv1.ServiceAccountKind,
+					Name:      ArgoCDManagerServiceAccount,
+					Namespace: ns,
+				}},
+			}
+			_, err = clientset.RbacV1().RoleBindings(namespace).Create(context.Background(), &roleBinding, metav1.CreateOptions{})
+			if err != nil {
+				if !apierr.IsAlreadyExists(err) {
+					_, err := clientset.RbacV1().RoleBindings(namespace).Update(context.Background(), &roleBinding, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
jgwest: As above, fluent API is from clientset is probably fine.

.github/workflows/static.yml
```diff
@@ -42,15 +48,18 @@ jobs:
         run: |
           cd $GITHUB_WORKSPACE/backend
           make gosec
-      - name: "Ensure 'cluster-agent' passes 'go-sec' - run 'make gosec' to identify issues"
+        - name: "Ensure 'backend' passes 'go-lint' - run 'make lint' to identify issues"
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.40.1
+          cd $GITHUB_WORKSPACE/backend
+          make lint - name: "Ensure 'cluster-agent' passes 'go-sec' - run 'make gosec' to identify issues"
```
jgwest, line 55: Looks like the `-name:` field from the following line got mixed up into the `make lint` line.

backend-shared/hack/service_account.go
jgwest: I recommend sticking with `fmt.Errorf` for now, and then if/when we want to do testing of specific errors, to refactor the code at that point.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+
+	serviceAccount := corev1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "v1",
+			APIVersion: "ServiceAccount",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        serviceAccountName,
+			Namespace:   namespace,
+			ClusterName: "",
```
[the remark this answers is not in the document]
jgwest: +1, ClusterName should be removed (according to doc comment, it's not even used by api server)

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+	}
+	_, err := clientset.CoreV1().ServiceAccounts(namespace).Create(context.Background(), &serviceAccount, metav1.CreateOptions{})
+	if err != nil {
+		if !apierr.IsAlreadyExists(err) {
+			return fmt.Errorf("Failed to create service account %q in namespace %q: %v", serviceAccountName, namespace, err)
```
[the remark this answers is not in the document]
jgwest: +1, and VS Code will show you all instances of this in the Problems view.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func InstallServiceAccount(clientset kubernetes.Interface, ns string, namespaces []string) (string, error) {
+	// ArgoCDManagerServiceAccount = "argocd-manager", is the name of the service account for managing a cluster
+
+	err := CreateServiceAccount(clientset, "argocd-manager", ns)
+	if err != nil {
+		return "", err
+	}
+	if len(namespaces) != 0 {
```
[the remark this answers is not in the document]
jgwest: The `if len(namespaces) != 0 {` can be removed, as the `range` on the next line will skip processing the namespace list if the list is empty.

(Also, while cognitive complexity of code is definitely a thing, I'm skeptical of using a tool to check it, versus a human brain judging whether the code is as complex as it should be,  but am willing to explore the idea further.)

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+	return GetServiceAccountBearerToken(clientset, ns, ArgoCDManagerServiceAccount)
+}
+
+// GetServiceAccountBearerToken will attempt to get the provided service account until it
+// exists, iterate the secrets associated with it looking for one of type
+// kubernetes.io/service-account-token, and return it's token if found.
+func GetServiceAccountBearerToken(clientset kubernetes.Interface, ns string, sa string) (string, error) {
+	var serviceAccount *corev1.ServiceAccount
+	var secret *corev1.Secret
+	var err error
```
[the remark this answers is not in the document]
jgwest: I recommend preserving consistency with the rest of the codebase, where we are not grouping variables like this (and also as a matter of preference, I'm not a fan of this format :smile: )

backend-shared/hack/service_account.go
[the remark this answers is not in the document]
jgwest: Exactly, this will be consumed in the code from latest stories (you can search the code for TODOs that reference the story number, for example.)

backend-shared/hack/service_account.go
jgwest: I'm not sure I agree with this as a stylistic format, and thus don't think it should be a requirement as a change to this PR.

backend-shared/hack/service_account.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func generateClientFromClusterServiceAccount(hostname string, bearerToken string) client.Client {
+	config := &rest.Config{
+		Host:        hostname,
+		BearerToken: bearerToken,
+	}
+	clientObj, err := client.New(config, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		log.Fatal(err)
```
jgwest, line 165: This should return the error, rather than logging and panic-ing.

jgwest (changes requested):
This is my first pass at a review, I still need to look at the functionality to make sure it is working as expected. Mainly just adding to some of the discussion around code style.

backend-shared/hack/service_account.go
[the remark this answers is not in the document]
drpaneas: That explains it, thx ;)

backend-shared/hack/service_account.go
drpaneas: Thanks for spotting this, I missed it during my review. Ouch!

backend-shared/hack/service_account.go
drpaneas: Yes, there's absolutely no need change the PR for such stuff, sorry if I didn't make it clear enough in my review. Honestly, I am also not sure I agree with some of these stylistic choices. I am just used to use https://github.com/bombsimon/wsl in my IDE, which is something I would be happy to get rid of :D

backend-shared/hack/service_account.go
samyak-jn: Thank you for that explanation, @drpaneas, noted :taco:

samyak-jn:
Hiya :hugs: 
Thank you for the code review here, just for actions items, the PR #54 should be merged before this, that will in turn fix the linting failure in the cluster agent mentioned in the CI!
Thanks :rocket:

jgwest (approved):
Looks great thanks @samyak-jn and @drpaneas for reviewing!

PR #52, opened 2022-01-10
Refactor devenv script

drpaneas:
In a nutshell:
* Make sure requires binaries are present to the local system (see `exit_if_binary_not_installed` function)
* Use `psql` as part of the postgres container (thus `psql.sh` is not needed anymore)
* refactor the 30-sec loop using the `wait-until` function so it aborts if there's failure
* create some variables for strings used in multiple places (e.g. container names)
* remove `-t` from the docker commands, otherwise CI fails.
* introduce testing after each command to make sure they work

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,5 +1,6 @@
 name: static checks
 on:
+  workflow_dispatch:
   push:
     branches:
       - "main"
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -1,66 +1,146 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+# Checks if a binary is present on the local system
+exit_if_binary_not_installed() {
+  for binary in "$@"; do
+    command -v "$binary" >/dev/null 2>&1 || {
+      echo >&2 "Script requires '$binary' command-line utility to be installed on your local machine. Aborting..."
+      exit 1
+    }
+  done
+}
 
-# Create docker network if one doesn't exist yet
-if [ "$(docker network ls --filter 'name=gitops-net' -q |wc -l)" == 0 ]; then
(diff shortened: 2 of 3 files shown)


samyak-jn (approved):

jgwest (changes requested):
Looks great, thanks @drpaneas! Always good to see someone with better Bash skills than my own (admittedly that doesn't take much :smile:).

One tweak requested: I'd like to keep `psql.sh` around if possible, for developers to run from the CLI in order to do local debugging of the database. I find it useful as a quick and easy way to debug the database (rather than using pgadmin), and having the shell script containing the database connection information already is helpful.

However, the old version of this script assumed that the user had psql installed; instead we can just run it within a Docker container, just like your PR is currently doing in `create-dev-env.sh`.

drpaneas:
Thanks @jgwest for the review :D
> One tweak requested: I'd like to keep `psql.sh` around if possible [...] instead we can just run it within a Docker container, just like your PR is currently doing in `create-dev-env.sh`.

I've added the `psql.sh` back. The behavior is the following:

* if postgres container is running:
```bash
./psql.sh -l
                                 List of databases
   Name    |  Owner   | Encoding |  Collate   |   Ctype    |   Access privileges
-----------+----------+----------+------------+------------+-----------------------
 postgres  | postgres | UTF8     | en_US.utf8 | en_US.utf8 |
 template0 | postgres | UTF8     | en_US.utf8 | en_US.utf8 | =c/postgres          +
           |          |          |            |            | postgres=CTc/postgres
 template1 | postgres | UTF8     | en_US.utf8 | en_US.utf8 | =c/postgres          +
           |          |          |            |            | postgres=CTc/postgres
(3 rows)
```

If postgress container is not running:

```bash
./psql.sh -l
Error: cannot interact with the PostgreSQL server: 'managed-gitops-postgres' container is not running.
Run './create-dev-env.sh script first
```

Let me know if this works for you ;)

jgwest (approved):
Looks great, thanks again @drpaneas!

PR #55, opened 2022-01-11
Enable weekly automatic dependency PRs

drpaneas:
Dependabot is nowadays bought by GitHub. This configuration does the following:

* Updates the GitHub actions
* Updates the `go.mod` for every Golang based sub-repo here.

The bot is going to wake up every Sunday at 08:00 (Berlin timezone), check if there are new versions, and submit PRs, tagged with "🔼:" commit-message and also adding the "🔼 dependencies". I have put myself as the assignee so I can review how this goes.

Reason is that refreshing the libraries and be update it a good thing, but I am mostly interested in the security vulnerabilities introduced by specific versions we use. Dependabot is reviewing those and creates PRs with their new patched versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,88 @@
+# https://github.com/dependabot
+# Configuration options for dependency updates
+# See: https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      timezone: Europe/Berlin
+      time: "08:00"
+    assignees:
+      - "drpaneas"
+    # Allow up to 10 open pull requests for GitHub dependencies
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: "🔼:"
+      prefix-development: "🔼"
+    labels:
+      - "🔼 dependencies"
+
+  # Maintain Go dependencies for backend repo
+  - package-ecosystem: "gomod"
+    directory: "/backend"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      timezone: Europe/Berlin
+      time: "08:00"
+      assignees:
+        - "drpaneas"
+      open-pull-requests-limit: 10
+      commit-message:
+        prefix: "🔼:"
+      labels:
+        - "🔼 dependencies"
+
+  # Maintain Go dependencies for backend-shared repo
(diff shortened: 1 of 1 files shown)


samyak-jn (commented):
Hi,
It will be very helpful, In the last story I worked on, some of the dependencies like controller-runtime, needs to be aligned with other k8s mods as well, from which one was not even with the latest upstream version release for the same, and hence I was facing conflict. We might need to test these dependent bot alerts before merging by running all the tests.
Also, there was a scenario that I wanted to ask a quick question on, Since the cluster-agent module was updated, the backend module needed to be updated as well, will dependent bot be responsible for adding version updates in one go (in all the submodules), or will there be different PRs for the same module version update in different sub-modules within the parent repository (because if that is the case, it might highly likely to fill due to gosec lint and fmt checks).

drpaneas:
This PR actually was inspired from your last problem with dependency issues. Each dependabot PR has CI tests running, which means in order to be merged, we will know what:

* Project builds fine
* Tests are passing

For any libraries we want to be _locked_ on a hardcoded version we can always put them in the `ignore` section of the configuration. But first we need to know _which_ these libraries are.

> Since the cluster-agent module was updated, the backend module needed to be updated as well, will the dependent bot be responsible for adding version updates in one go

If both sub-projects use the same dependency, I would expect to create two PRs, one for the cluster-agent and one for the backend module. This is a not a problem of the bot per-se, but it's an issue with having mono-repo structure.

Personally I am using _depandabot_ for many projects and I haven't noticed any issues (last famous words :P), given  running it in weekly basis, I don't expect this bot to create more than 1 or 2 PRs every month. I would be in favour of giving this a try, and if it creates problems or unnecessary noise, we can always disable it and throw it away :)

jgwest:
@drpaneas do you have an example of what the bot would currently recommend for our dependency versions? (e.g. is there a way to do an offline run of the update logic?)

drpaneas:
Good point @jgwest  -- not I don't think we can call this locally, but I can fork it in another repo and see what's going tell us there.

drpaneas:
here they are https://github.com/drpaneas/managed-gitops/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

As you can see, some PRs failed to pass the CI, which is a good indication to know which versions work for us and which do not.

e.g. 

```
Error: ../../../../go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/source/source.go:131:13: undefined: wait.PollImmediateUntilWithContext
```

drpaneas:
@jgwest what's your take on this?

drpaneas:
I will close this for now as stale, feel free to re-open if needed.

PR #57, opened 2022-01-12
Use GOBIN instead of ~/go/bin/

drpaneas:
When I tried `make start` it failed because the makefile expects `goreman` to be located at `~/go/bin`. I am not using `~/go/bin` for the go binaries but I do use `GOBIN` env variable to handle them. GOBIN is basically `GOPATH/bin` which is the place here the go binaries go after running `go install`. Would that work for you?

Also modified the readme at there is no more `make install` target and did some syntax cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 
 MAKEFILE_ROOT=$(shell pwd)
+GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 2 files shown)


jgwest (approved):
Looks great, thanks @drpaneas!

PR #58, opened 2022-01-20
Add ArgoCD Load Test part-2 (Solves: GITOPS-1508)

samyak-jn:
Hi, :hugs:! 

This is a draft PR for ArgoCD LoadTest [GITOPS-1508](https://issues.redhat.com/browse/GITOPS-1508), which adds a test for testing n applications with non-pod based resources. The repository that contains those manifests lies here for [now](https://github.com/samyak-jn/gitops-service-sample-k8s-resources).

### DIFF
--- a/utilities/load-test/loadtest/loadtest_test.go
+++ b/utilities/load-test/loadtest/loadtest_test.go
@@ -28,7 +28,7 @@ var (
 func TestNGuestbook(t *testing.T) {
 	// TEST 1: Deply 100 guestbook in different namespace
 	t.Log("Testing for 100 Guestbook applications\n")
-
+	numberOfAppsToCreate := 100
 	t.Log("\n\nPods Memory Resource before creation:\n\n")
 	memoryInit := GetPodInfo(kubeconfig, masterUrl, namespace, podName)
 	PodInfoParse(memoryInit)
@@ -36,7 +36,7 @@ func TestNGuestbook(t *testing.T) {
 	// Get Pod Restart Count
 	PodC := PodRestartCount(namespace)
 
-	for i := 1; i < 101; i++ {
+	for i := 1; i < numberOfAppsToCreate+1; i++ {
 		guestbookApp := &appv1.Application{
 			TypeMeta: v1.TypeMeta{
 				Kind:       "Application",
@@ -109,7 +109,7 @@ func TestNGuestbook(t *testing.T) {
 	PodInfoParse(delta)
 
 	// To delete a application from the given (ex: argocd) namespace
-	for i := 1; i < 101; i++ {
+	for i := 1; i < numberOfAppsToCreate+1; i++ {
 		err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "guestbook-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
 
 		if err_del != nil {
@@ -299,7 +299,7 @@ func TestAllApplication(t *testing.T) {
 	//	TEST 3: Deply Prometheus (heavy application) + 20 Guestbook Application (small application)
 
 	t.Log("Testing for Prometheus + 20 Guestbook Application\n")
-
+	numberOfGuestbookAppsToCreate := 20
 	t.Log("\n\nPods Memory Resource before creation:\n\n")
 	// Intial Memory Information
 	memoryInit := GetPodInfo(kubeconfig, masterUrl, namespace, podName)
@@ -342,7 +342,7 @@ func TestAllApplication(t *testing.T) {
 		return
 	}
 
-	for i := 1; i < 21; i++ {
+	for i := 1; i < numberOfGuestbookAppsToCreate+1; i++ {
 		guestbookApp := &appv1.Application{
 			TypeMeta: v1.TypeMeta{
 				Kind:       "Application",
@@ -417,7 +417,7 @@ func TestAllApplication(t *testing.T) {
 	// To delete a application from the given (ex: argocd) namespace
 	err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "prometheus-app", v1.DeleteOptions{})
 	// Deleting 10 guestbook applications
-	for i := 1; i < 21; i++ {
+	for i := 1; i < numberOfGuestbookAppsToCreate+1; i++ {
 		err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "guestbook-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
 
 		if err_del != nil {
@@ -433,9 +433,11 @@ func TestAllApplication(t *testing.T) {
 }
 
 func TestNResource(t *testing.T) {
-	cleanup := func() {
+	numberOfAppsToCreate := 100
+
+	cleanup := func(numberOfAppsToCreate int) {
 		// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
-		for i := 1; i < 101; i++ {
+		for i := 1; i < numberOfAppsToCreate+1; i++ {
 			err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "test-app-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
 			if err_del != nil {
 				t.Errorf("error, %v", err_del)
@@ -443,6 +445,7 @@ func TestNResource(t *testing.T) {
 			_ = GetE2EFixtureK8sClient().KubeClientset.CoreV1().Namespaces().Delete(context.TODO(), "test-namespace-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
 		}
 	}
+	defer cleanup(numberOfAppsToCreate)
 
 	// TEST: Deply 100 sample application in different namespace
 	t.Log("Testing for 100 applications that uses lightweight non-Pod-based resources\n")
@@ -454,7 +457,7 @@ func TestNResource(t *testing.T) {
 	// Get Pod Restart Count
 	PodC := PodRestartCount(namespace)
 
-	for i := 1; i < 101; i++ {
+	for i := 1; i < numberOfAppsToCreate+1; i++ {
 		applicationTest := &appv1.Application{
 			TypeMeta: v1.TypeMeta{
 				Kind:       "Application",
@@ -524,8 +527,6 @@ func TestNResource(t *testing.T) {
 	t.Log("\n\nDifference in the Pod Memory Usage (in Ki)\n\n")
 	delta := PodMemoryDiff(memoryInit, memoryPost)
 	PodInfoParse(delta)
-
-	defer cleanup()
 
 	t.Log("100 applications are created, deployed and deleted successfully!")
 }
--- a/utilities/load-test/loadtest/loadtest_test.go
+++ b/utilities/load-test/loadtest/loadtest_test.go
@@ -433,7 +433,18 @@ func TestAllApplication(t *testing.T) {
 }
 
 func TestNResource(t *testing.T) {
-	// TEST 1: Deply 100 guestbook in different namespace
+	cleanup := func() {
+		// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
+		for i := 1; i < 101; i++ {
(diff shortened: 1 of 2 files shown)


utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +413,95 @@ func TestAllApplication(t *testing.T) {
...
+		applicationTest := &appv1.Application{
+			TypeMeta: v1.TypeMeta{
+				Kind:       "Application",
+				APIVersion: "argoproj.io/v1alpha1",
+			},
+			ObjectMeta: v1.ObjectMeta{
+				Name:      "test-app-" + fmt.Sprintf("%d", i),
+				Namespace: "argocd",
+			},
+			Spec: appv1.ApplicationSpec{
```
jgwest, line 438: In order to ensure that Argo CD starts synchronizing the `Application` resource as soon as it is created, you can use `.spec.syncPolicy.automated`, like [this](https://github.com/argoproj/argo-cd/blob/9ac935d59fd90f6f568c04241047d5d45bb7bb9b/docs/operator-manual/application.yaml#L108) (in Go, rather than YAML, of course).

Once you do that, Argo CD will probably complain about the namespace being missing; you can use [this](https://github.com/argoproj/argo-cd/blob/9ac935d59fd90f6f568c04241047d5d45bb7bb9b/docs/operator-manual/application.yaml#L115) sync option in the ApplicationSpec, in order to have Argo CD create that namespace for us.

(But we should make sure to clean up those namespaces when the test ends)

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +413,95 @@ func TestAllApplication(t *testing.T) {
...
+	// To delete a application from the given (ex: argocd) namespace
+	for i := 1; i < 101; i++ {
+		err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "test-app-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
+
+		if err_del != nil {
+			t.Errorf("error, %v", err_del)
+			return
+		}
+	}
+	t.Log("100 applications are created, deployed and deleted successfully!")
```
jgwest, line 506: As mentioned above, we should clean up the namespaces as well (if Argo CD doesn't delete them for us, which I *think* it won't).

jgwest (changes requested):
Looking good, a couple of items that popped out at me.

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: Yes, I realised it later, I have code for that locally. I'll push it once refined. Thanks! :D

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: Good catch, thanks!
I'll take this in account :D

samyak-jn (commented):
Added options for automated syncOptions, and functionality to create a namespace.
Once, those test-namespaces are created and used for the given purpose, the namespaces will be destroyed from the cluster.

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,102 @@ func TestAllApplication(t *testing.T) {
...
+			},
+			ObjectMeta: v1.ObjectMeta{
+				Name:      "test-app-" + fmt.Sprintf("%d", i),
+				Namespace: "argocd",
+			},
+			Spec: appv1.ApplicationSpec{
+				Source: appv1.ApplicationSource{
+					Path:           "sample-app",
+					TargetRevision: "HEAD",
+					RepoURL:        "https://github.com/samyak-jn/gitops-service-sample-k8s-resources.git",
```
jgwest, line 460: Can you remove the namespace field from the metadata here:
https://github.com/samyak-jn/gitops-service-sample-k8s-resources/blob/52dad3c7a8526031561d4e9820ebc58ba04f7c6d/sample-app/roles.yaml#L5

It appears to be causing the Argo CD application not to be healthy.

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,102 @@ func TestAllApplication(t *testing.T) {
...
+
+		// list all Argo CD applications in the namespace
+		appList, err := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).List(context.TODO(), v1.ListOptions{})
+		if err != nil {
+			t.Errorf("error, %v", err)
+		}
+		// for each application, check if (app variable).Status.Health.Status
+		for _, app := range appList.Items {
+			// if status is empty, return false
+			if app.Status.Health.Status == "" {
```
jgwest, line 491: We should wait for all of the applications in the list to have Health Status of 'Healthy' and Sync Status of 'Synced'

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,102 @@ func TestAllApplication(t *testing.T) {
...
+
+	t.Log("\n\nPods Memory Resource after Creation:\n\n")
+	memoryPost := GetPodInfo(kubeconfig, masterUrl, namespace, podName)
+	PodInfoParse(memoryPost)
+
+	t.Log("\n\nDifference in the Pod Memory Usage (in Ki)\n\n")
+	delta := PodMemoryDiff(memoryInit, memoryPost)
+	PodInfoParse(delta)
+
+	// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
```
jgwest, line 517: At the moment, the test will not cleanup the resources that it created if we `return` before this for loop.

We can ensure that the test  cleans up after itself,  by moving the cleanup code into a function defined within the test, and calling it using `defer`.
```
func TestNResource(t *testing.T) {

	cleanup := func() {
		// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
		for i := 1; i < 101; i++ {
			err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "test-app-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
			if err_del != nil {
				t.Errorf("error, %v", err_del)
			}
			err_del_ns := GetE2EFixtureK8sClient().KubeClientset.CoreV1().Namespaces().Delete(context.TODO(), "test-namespace-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
			if err_del_ns != nil {
				t.Errorf("error, %v", err_del_ns)
			}
		}
	}

	defer cleanup()
// ...
```

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,102 @@ func TestAllApplication(t *testing.T) {
...
+	for i := 1; i < 101; i++ {
+		err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "test-app-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
+		if err_del != nil {
+			t.Errorf("error, %v", err_del)
+			return
+		}
+		err_del_ns := GetE2EFixtureK8sClient().KubeClientset.CoreV1().Namespaces().Delete(context.TODO(), "test-namespace-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
+		if err_del_ns != nil {
+			t.Errorf("error, %v", err_del_ns)
```
jgwest, line 526: We can skip reporting the error if the namespace doesn't exist (this is likely because argo cd has deleted it).

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,102 @@ func TestAllApplication(t *testing.T) {
...
+	t.Log("\n\nDifference in the Pod Memory Usage (in Ki)\n\n")
+	delta := PodMemoryDiff(memoryInit, memoryPost)
+	PodInfoParse(delta)
+
+	// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
+	for i := 1; i < 101; i++ {
+		err_del := GetE2EFixtureK8sClient().AppClientset.ArgoprojV1alpha1().Applications(namespace).Delete(context.TODO(), "test-app-"+fmt.Sprintf("%d", i), v1.DeleteOptions{})
+		if err_del != nil {
+			t.Errorf("error, %v", err_del)
+			return
```
jgwest, line 522: Can remove this 'return', so it keeps trying to remove the remaining items in the list, rather than stopping on the first error.

jgwest (changes requested):

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: Oh Yes, TIL, a good catch indeed. So, follow up question just for learning is, even if above code from the test execute return statement, the cleanup function defined within the test executes?

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: yes, I had encountered this situation, i forgot to remove it, thanks!

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: yep, added a constraint for this

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,101 @@ func TestAllApplication(t *testing.T) {
...
+
+	t.Log("\n\nPods Memory Resource after Creation:\n\n")
+	memoryPost := GetPodInfo(kubeconfig, masterUrl, namespace, podName)
+	PodInfoParse(memoryPost)
+
+	t.Log("\n\nDifference in the Pod Memory Usage (in Ki)\n\n")
+	delta := PodMemoryDiff(memoryInit, memoryPost)
+	PodInfoParse(delta)
+
+	defer cleanup()
```
jgwest, line 528: The defer needs to move to the start of the method, otherwise the cleanup will never be called.

utilities/load-test/loadtest/loadtest_test.go
```diff
@@ -413,3 +431,101 @@ func TestAllApplication(t *testing.T) {
 	}
 	t.Log("Prometheus, along with 20 guestbook applications are created, deployed and deleted successfully!")
 }
+
+func TestNResource(t *testing.T) {
+	cleanup := func() {
+		// To delete a application from the given (ex: argocd) namespace and delete the test-namespace created
+		for i := 1; i < 101; i++ {
```
jgwest, line 438: New requirement: can you add a constant at the top of the method, like:
```
numberOfAppsToCreate := 100
```
And then use this value in the various for loops? That we can easily control the number of applications to create as part of this test.

jgwest (changes requested):
Thanks @samyak-jn, two additional requests:

1) Can you create a new directory in https://github.com/samyak-jn/gitops-service-sample-k8s-resources.git , under, for example, `/sample-app-50`, that contains ~50 resources to deploy to the cluster? So for example:
- 13 config-maps
- 13 deployments
- 13 roles
- 13 secrets

This allows us to simulate a larger project (you don't need to update the test for this, just create the repo for our future use).

2) It looks like there's a bug in the code that gets the memory consumption of the pods; sometimes it shows 0.

This is because: in `GetPodInfo`, we are using `elements.Containers[0]` which is incorrect; some Pods contain multiple containers, so we should ensure that the memory usage we report is a sum of the memory usage of all the containers within the pod.

For example, this contains two containers, `argocd-application-controller` and `POD`.
```
apiVersion: v1
items:
- apiVersion: metrics.k8s.io/v1beta1
  containers:
  - name: argocd-application-controller
    usage:
      cpu: 223m
      memory: 1461428Ki 
  - name: POD
    usage:
      cpu: "0"
      memory: "0"
  kind: PodMetrics
```

utilities/load-test/loadtest/loadtest_test.go
samyak-jn: yes, it will clean the code as well by creating a better understanding thanks! :")

samyak-jn (commented):
Thank you for the review @jgwest, made relevant changes and fixed bug as reported.

Just a quick glimpse how the output for pod resource will look like:

```
Pod Name                                Container Name(s)                  Memory Usage (in Mi)
argocd-repo-server-5f9b5d9584-wcs2g     POD, argocd-repo-server               45
argocd-application-controller-0         argocd-application-controller, POD    59
argocd-dex-server-5dc4df4967-pfbzt      dex, POD                              13
argocd-redis-5b6967fdfc-kbvks           redis                                  2
argocd-server-86dcc9f88f-b64d6          argocd-server                         27
```
Since, we had various container resource within the pods, therefore the right thing was to mention all those in the list so added that functionality as well.

On testing on openshift cluster, I got to knew there were "POD" container resource just not for the argocd-application-controller but for other pods as well. But now the code is adjusted accordingly, thanks! :")

jgwest (approved):
Thanks @samyak-jn!

PR #61, opened 2022-01-31
Add test framework code for testing database tests (Solves: GITOPS-1634)

samyak-jn:
Hi, :hugs:!

This is in regards with issue GITOPS-1634, this PR mainly involves two main methods (NewEphemeralCreateTestFramework, and, NewEphemeralCleanTestFramework) which are responsible for creating and cleaning a ephemeral, disposable Postgresql container which is being merely used for testing purpose.

The tests written against above method might look like this:

```
func TestEphemeralCode(t *testing.T) {
	ctx := context.Background()
	// Create ephemeral DB
	ephemeralDB, err := NewEphemeralCreateTestFramework()
	assert.NoError(t, err)

	dbq := ephemeralDB.db
	defer dbq.CloseDatabase()

	// test code as normal, for example:
	var applicationStates []ApplicationState
	err = dbq.UnsafeListAllApplicationStates(ctx, &applicationStates)
	assert.NoError(t, err)

	errClean := NewEphemeralCleanTestFramework(ephemeralDB.dbContainerID, ephemeralDB.dockerNetwork)
	assert.NoError(t, errClean)
}

```

### DIFF
--- a/backend-shared/config/db/ephemeralPostgres.go
+++ b/backend-shared/config/db/ephemeralPostgres.go
@@ -182,3 +182,27 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 
 	return res, nil
 }
+
+// NewEphemeralCleanTestFramework will clean a docker container,
+// and network. This is merely defined for testing purpose
+func NewEphemeralCleanTestFramework(dockerContainerID string, tempNetworkName string) error {
+
+	dockerCmd := "docker rm -f %s"
+	fmt.Println("\nRunning: ", fmt.Sprintf(dockerCmd, dockerContainerID))
+
+	// To get the output of the command
+	_, err := exec.Command("docker", "rm", "-f", dockerContainerID).Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	dockerNetworkcmd := "docker network rm %s"
+	fmt.Println("\nRunning: ", fmt.Sprintf(dockerNetworkcmd, tempNetworkName))
+
+	// To get the output of the command
+	_, err = exec.Command("docker", "network", "rm", tempNetworkName).Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return nil
+}
--- a/backend-shared/config/db/ephemeralPostgres.go
+++ b/backend-shared/config/db/ephemeralPostgres.go
@@ -32,7 +32,7 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 	// To print which command is running
 	fmt.Println("\nRunning: ", s)
 
-	// To get the output of the command
+	// #nosec G204
 	dockerNetwork := exec.Command("docker", "network", "create", tempNetworkName)
 	dockerNetworkerr := dockerNetwork.Run()
 	if dockerNetworkerr != nil {
@@ -66,6 +66,7 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 	var dockerContainerID []byte
 	var errDockerRun error
 	errWait := wait.Poll(5*time.Second, 2*time.Minute, func() (bool, error) {
+		// #nosec G204
 		dockerContainerID, errDockerRun = exec.Command("docker", "run", "--name", dockerName,
 			"-v", string(tempDatabaseDir)+":/var/lib/postgresql/data:Z",
 			"-e", "POSTGRES_PASSWORD=gitops",
@@ -84,6 +85,7 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 			return false, errDockerRun
 		}
 		// check for container status
+		// #nosec G204
 		status, _ := exec.Command("docker", "container", "inspect", "-f", "{{.State.Status}}", string(dockerContainerID)).Output()
 		if string(status) == "running" {
 			return true, nil
@@ -102,6 +104,7 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 	fmt.Println("\nRunning: ", s)
 	// To get the output of the command
 	errWait = wait.Poll(5*time.Second, 2*time.Minute, func() (bool, error) {
+		// #nosec G204
 		psqlcmd := exec.Command("psql", "-h", "localhost", "-d", tempDBName, "-U", "postgres", "-p", "6432", "-c", "select 1")
 		psqlcmd.Env = os.Environ()
 		psqlcmd.Env = append(psqlcmd.Env, "PGPASSWORD=gitops")
--- a/backend-shared/config/db/ephemeralPostgres.go
+++ b/backend-shared/config/db/ephemeralPostgres.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"log"
-	"os"
 	"os/exec"
 	"strings"
 	"time"
@@ -24,13 +23,9 @@ type EphemeralDB struct {
 // merely defined for testing purpose
 func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 	dockerName := "managed-gitops-postgres-test"
-	dockerNetworkcmd := "docker network create %s"
 	uuid := uuid.New().String()
 	tempDBName := "db-" + uuid
 	tempNetworkName := "gitops-net-" + uuid
-	s := fmt.Sprintf(dockerNetworkcmd, tempNetworkName)
-	// To print which command is running
-	fmt.Println("\nRunning: ", s)
 
 	// #nosec G204
 	dockerNetwork := exec.Command("docker", "network", "create", tempNetworkName)
@@ -39,29 +34,11 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 		log.Fatal(dockerNetworkerr)
 	}
 
-	tempDatabaseDircmd := "mktemp -d -t postgres-XXXXXXXXXX"
-	s = fmt.Sprintf(tempDatabaseDircmd)
-	fmt.Println("\nRunning: ", s)
-
-	// To actually run the command (runs in background)
+	// creates a temp directory where postgres functionality are performed
 	tempDatabaseDir, err_run := exec.Command("mktemp", "-d", "-t", "postgres-XXXXXXXXXX").Output()
 	if err_run != nil {
 		log.Fatal(err_run)
 	}
-
-	// running a docker container
-	dockerContainerIDcmd := `docker run --name ` + dockerName + ` \
-	-v ` + string(tempDatabaseDir) + `:/var/lib/postgresql/data:Z \
-	-e POSTGRES_PASSWORD=gitops \
-	-e POSTGRES_DB=` + tempDBName + ` \
-	-p 6432:5432 \
-	--network ` + tempNetworkName + ` \
-	-d \
-	postgres:13 \
-	-c log_statement='all' \
-	-c log_min_duration_statement=0`
-
-	fmt.Println("\nRunning:", dockerContainerIDcmd)
 
 	var dockerContainerID []byte
 	var errDockerRun error
@@ -90,24 +67,16 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 		if string(status) == "running" {
 			return true, nil
 		}
-		// Todo: Once verified removed these print statements in order to maintain code coverage
-		fmt.Println("Docker Container ID: " + string(dockerContainerID))
 		return true, nil
 	})
 	if errWait != nil {
 		log.Fatal("error in executing docker run command: ", errWait)
 	}
 
-	dbcmd := "PGPASSWORD=gitops psql -h localhost -d %s -U postgres -p 6432 -c 'select 1'"
-	s = fmt.Sprintf(dbcmd, tempDBName)
-
-	fmt.Println("\nRunning: ", s)
-	// To get the output of the command
+	// connects to the database inside the docker container
 	errWait = wait.Poll(5*time.Second, 2*time.Minute, func() (bool, error) {
 		// #nosec G204
-		psqlcmd := exec.Command("psql", "-h", "localhost", "-d", tempDBName, "-U", "postgres", "-p", "6432", "-c", "select 1")
-		psqlcmd.Env = os.Environ()
-		psqlcmd.Env = append(psqlcmd.Env, "PGPASSWORD=gitops")
+		psqlcmd := exec.Command("docker", "exec", "--user", "postgres", "-e", "PGPASSWORD=gitops", "-i", dockerName, "psql", "-h", "localhost", "-d", tempDBName, "-U", "postgres", "-p", "5432", "-c", "select 1")
 		var outb, errb bytes.Buffer
 		psqlcmd.Stdout = &outb
 		psqlcmd.Stderr = &errb
@@ -121,22 +90,16 @@ func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
 		if psqlErr != nil {
 			return false, psqlErr
 		}
-		fmt.Printf("%s database is ready to use\n", tempDBName)
 		return true, nil
 	})
 	if errWait != nil {
 		log.Fatal("error in executing docker run command: ", errWait)
 	}
 
-	// creating a new database
+	// creating a new database inside the postgres container
 	newDBName := "postgres"
-	dbcmd = "PGPASSWORD=gitops psql -h localhost -d %s -U postgres -p 6432"
-	s = fmt.Sprintf(dbcmd, newDBName)
-	fmt.Println("\nRunning: ", s)
+	psqlcmd := exec.Command("docker", "exec", "--user", "postgres", "-e", "PGPASSWORD=gitops", "-i", dockerName, "psql", "-h", "localhost", "-d", newDBName, "-U", "postgres", "-p", "5432")
 
-	psqlcmd := exec.Command("psql", "-h", "localhost", "-d", newDBName, "-U", "postgres", "-p", "6432")
-	psqlcmd.Env = os.Environ()
-	psqlcmd.Env = append(psqlcmd.Env, "PGPASSWORD=gitops")
 	var errConnection bytes.Buffer
 	psqlcmd.Stderr = &errConnection
 	psqlErr := psqlcmd.Run()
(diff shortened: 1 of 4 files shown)


drpaneas (commented):
Before I review this PR, I would need some help to understand first what is the problem we are trying to solve here. At first sight, mixing application code with testing infrastructure code, sounds like a door to many problems to me, but I might be mistaken. The rest of this comment is not me preaching, but speaking my thoughts out loud:

* is it all about cleaning up the postgres docker container leftovers? If so, just add `docker --rm`  and the container will be ephemeral, so there is no need to keep track of it. Once it's finished testing, it will be automatically deleted. That's it :)

* is it about spawning a new ephemeral db container every time there is a Go function that wants to interact with the database? If so, what's the benefit? For sure this nor scalable or efficient way of doing it (imagine a situation where the tests interact 100 times with the db).

* is it about moving away from `bash` files (e.g. deleting `create-dev-env.sh`, `delete-dev-env.sh`, `psql.sh`, `stop-dev-env.sh` in the long run) and having a more _robust_ testing pipeline? If so, we can use `docker-compose` to bring up and breakdown the necessary _devenv_ we need.

Looking at the code, I am even more confused. The title says "add a test framework for testing database tests", but all this PR is doing is just replicating the bash commands by calling the very same bash commands themselves using the `os` package. That's easy. But also ugly.

For me it makes things less flexible (e.g. replacing create if something goes wrong, in order to debug, the developer would have to *recompile* the whole thing, instead of just modifying a bash file. IMHO if you want to write bash, write bash. If you want to write Go, write Go. But writing `bash` using `Go`, it's still `bash`. So, what's the benefit?

e.g. If you need to interact with Docker, then talk directly with Docker, "programmatically", so in that way you can make the project cross-platform as well. e.g. or you can use https://github.com/testcontainers/testcontainers-go as it was proposed in the JIRA story.

Also, I've noticed that at some places you have hardcoded paths like `"../../../db-schema.sql")`. If you absolutely need to do it, I would suggest to move the file into a `testdata` directory, so it's clear that this is used for testing, otherwise someone can easily move it around and break the test. Or an even better alternative to avoid reading the disk multiple times, would be to use the [Go embed](https://blog.jetbrains.com/go/2021/06/09/how-to-use-go-embed-in-go-1-16/) functionality introduced with Go 1.16.

FYI: what would bring a really big benefit in terms of database and flexibility in the future, is the ability to able to seamlessly switch from one database to another without having to re-write any anything (not even the tests). So replacing the current DB calls with https://github.com/madflojo/hord is something useful we need to consider if in the future we would like to go away from postgres.

Again, this is not me preaching here. I really enjoy writing long comments, hopefully you are also interested in reading them, hahahaha 😆 

I am just trying to understand what this PR is improving from the current `main` branch.

jgwest:
(I too like long walls of text :stuck_out_tongue: )

@drpaneas , I presume you've seen [the JIRA story](https://issues.redhat.com/browse/GITOPS-1634) that this PR corresponds to? It goes into a bit of detail on the goals of this. 

> Is it about spawning a new ephemeral db container every time there is a Go function that wants to interact with the database? If so, what's the benefit? For sure this not scalable nor efficient way of doing it (imagine a situation where the tests interact 100 times with the db).

This is the goal, but the idea is to spawn a new database for _each unit test that needs a database_. Will this be as performant as we need it to be? Only one way to find out, and that's to try to create a JIRA story for it, implement it, and and measure it :smile: . 

To remove any confusion, this PR doesn't contain product code, it is only for writing unit tests.

It's true that we could mock the database object, but IMHO this PR's approach is (or has the potential to be) preferable to mocking the database object, which I explored earlier in the year and found to be a flawed solution. Since I was already familiar with the idea of test-containers from the Java ecosystem, I thought it would be beneficial to explore here.

> All this PR is doing is just replicating the bash commands by calling the very same bash commands themselves using the os package. That's easy. But also ugly.
>
> For me it makes things less flexible (e.g. if something goes wrong, in order to debug or change a infrastructure related setting, the developer would have to recompile the application code instead of just modifying a bash file. IMHO if you want to write bash, write bash. If you want to write Go, write Go. But writing bash using Go, it's still bash. So, what's the benefit?
>
> e.g. If you need to interact with Docker, then talk directly with Docker, "programmatically", so in that way you can make the project cross-platform as well. e.g. or you can use https://github.com/testcontainers/testcontainers-go as it was proposed in the JIRA story.

FWIW I didn't propose *testcontainers-go* in the original story, and wasn't aware that it existed. I was only familiar with the Java 'testcontainers' projects (upon which the core idea for the JIRA is based). If I had known testcontainers-go, I probably would have suggested using it; it does look like a good fit for this. 

BUT, since I didn't know that when I wrote the story, and when I wrote the JIRA, Samyak implemented it how I personally would have implemented it. So I don't this is a fair criticism. 

Re: using Go, vs Bash, using Go was my suggestion for how to interact with Docker: the docker CLI is a stable API, and writing CLI commands to interact with it is straightforward. Samyak previously asked whether he should use bash, and I told him that Go was fine. We can probably reduce the amount of boilerplate in the code, but the approach would be the same.

I'm not convinced there is any huge benefit to be gained from writing it in bash vs Go. The commands we are issuing are very simple, and unlikely to change. Yes, you need to recompile Go to make an update, but this is trivial. You say Go is ugly, I say Bash is ugly :stuck_out_tongue:. But I say this as someone who speaks fluent Go, but not speak fluent Bash (as you can probably tell from some of the shell scripts and Makefiles I have written in this repo)


> Also, I've noticed that at some places you have hardcoded paths like "../../../db-schema.sql"). If you absolutely need to do it, I would suggest to move the file into a testdata directory, so it's clear that this is used for testing, otherwise someone can easily move it around and break the test. Or an even better alternative to avoid reading the disk multiple times, would be to use the Go embed functionality introduced with Go 1.16.

This PR is using the standard `db-schema.sql` at the root of the repo, which describes the database schema we using. Since this is part of a test case that will run as part of our automated CI, the test case will fail if anyone moves the `db-schema.sql`, thus blocking the merge, so it's really not possible for the test to get out of sync. 

I don't think we should move the schema to `testdata`, as then it is easy for the database schema to get out of sync with the test schema. And I'm not sure embedding it is possible, since we need to use psql on it.


> FYI: what would bring a really big benefit in terms of database and flexibility in the future, is the ability to seamlessly switch from one database to another without having to re-write any anything (not even the tests). So replacing the current DB calls with https://github.com/madflojo/hord is something useful we need to consider if in the future we would like to go away from postgres.

I don't expect we will be moving to another database any time soon, and if we did it would likely be another RDBMS (I like transactions, foreign keys, typed data, etc). If we did want database-neutrality, the maintainers of go-pg have [a replacement for go-pg](https://github.com/uptrace/bun) which proports to do this (but bun was not mature enough when this project was starting, and I'm not sure if it currently has enough traction). But that is obviously outside the scope of this PR.

drpaneas:
> This is the goal, but the idea is to spawn a new database for _each unit test that needs a database_. Will this be as performant as we need it to be? Only one way to find out, and that's to try to create a JIRA story for it, implement it, and and measure it 😄 .

That's perfectly fine then. +1 for experimentation :clap:

> BUT, since I didn't know that when I wrote the story, and when I wrote the JIRA, Samyak implemented it how I personally would have implemented it. So I don't this is a fair criticism.

Yes, you are correct. I also didn't know you didn't know 🔁

> You say Go is ugly, I say Bash is ugly 😛

Sorry I didn't put it clearly. What I meant is that Go is beautiful, bash is ugly, and by putting bash in Go as a wrapper, you make Go look ugly as well 😛

as always, thanks for the clarifications 💯

drpaneas (commented):
Hi @samyak-jn 👋

I see you had some great fun playing around with the `os/exec` pkg :D :tada:
Great work, looking forward using this! I've never tried to use container per unit test, and sounds quite exciting. Most likely this will take some iterations until it's robust, but we will that later down the road.

Looking at the code, from a quick glance, I think there's room for improvement for some repetitive parts. e.g. you can use a function for running shell commands, instead of repeating the same code again and again. For example, here's just a quick sample to explain better what I mean by that:

```go
const (
	succeed = "\u2713"
	failed  = "\u2717"
)

func runInBash(s string) (err error) {
	cmd := strings.Fields(s)
	execute := exec.Command(cmd[0], cmd[1:]...)
	if _, err = execute.Output(); err != nil {
		fmt.Printf("%s\t%s - %s\n", failed, execute, err)
	} else {
		fmt.Printf("%s\t%s\n", succeed, execute)
	}
	return err
}

func main() {
	if err := runInBash("docker ps"); err != nil {
		fmt.Printf("\t%s\n\n", "Error: Could not list docker processes")
	}

	if err := runInBash("ls -l -r -t -a"); err != nil {
		fmt.Println("Error: Could not list the current directory")
	}
}
```

This will print:

```shell
✗	docker ps - exec: "docker": executable file not found in $PATH
	Error: Could not list docker processes

✓	/bin/ls -l -r -t -a
````

I've also made some other comments in the code as well, but they are mostly questions to make sure you took some decisions intentionally or by accident. e.g. converting the `ConnectToDatabase` from unexported API to exported or calling `fatal` outside of `testing`, etc.

**Comment by samyak-jn** [comment]
Hehe, I started my morning with this long list of threads, but I do love to read long texts, it gives me a better way of exploring things out, and it often happens that I got to learn something new from them. So thank you @drpaneas and @jgwest for the above discussions.

Coming back to the PR,

> This is the goal, but the idea is to spawn a new database for each unit test that needs a database. Will this be as performant as we need it to be? Only one way to find out, and that's to try to create a JIRA story for it, implement it, and measure it. :smile:

Exactly, +1. The point here is just for testing purposes, as I have mentioned in my commit message as well as in my PR comment. (But yes, I should have verbally described that it) I explored using mock DB if I get that right from your suggestion, but I didn't get to a proper workaround to it mostly it was a problem with my configs that in isolation the test was working fine, but failed when ran along with other tests. I agree with your point that I should have researched more, instead of right diving into the next best solution I had. :astonished: 

> For me it makes things less flexible (e.g. if something goes wrong, in order to debug or change a infrastructure related setting, the developer would have to recompile the application code instead of just modifying a bash file. IMHO if you want to write bash, write bash. If you want to write Go, write Go. But writing bash using Go, it's still bash. So, what's the benefit?

As we discussed in our call as well, I agree with the fact that instead of writing the whole of this, I could have simply written two bash files one which results in creating docker and connecting to the DB and the other to clean that up and running this with os/exec package would be a clean way for this only approach. But I had dedicated almost my significant time, and the PR was almost completed.

> e.g. If you need to interact with Docker, then talk directly with Docker, "programmatically", so in that way you can make the project cross-platform as well. e.g. or you can use https://github.com/testcontainers/testcontainers-go as it was proposed in the JIRA story.

I took this as my learning point of not researching more here, I even didn't know test containers for golang even existed. I looked how java frameworks and how were they doing it. But as said, I'm adding this to my personal TODOs to explore and learn about it by opening a refactoring PR for this in the future. Thank you for bringing it into my presence. (see as I said long comments proves to be a learning thing for me :wink: )

> This PR is using the standard db-schema.sql at the root of the repo, which describes the database schema we using. 

+1 to this, Just to be in sync with the other tests and looking at the immobility of the schema I hardcoded it, but I'm happy to look for an alternative if any obstacle comes in the future.

Once again, thank you for the above discussion! :taco: :tada:

File: backend-shared/config/db/ephemeralPostgres.go [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+// network, connect to database and populate the schema. This is
+// merely defined for testing purpose
+func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
+	dockerName := "managed-gitops-postgres-test"
+	dockerNetworkcmd := "docker network create %s"
+	uuid := uuid.New().String()
+	tempDBName := "db-" + uuid
+	tempNetworkName := "gitops-net-" + uuid
+	s := fmt.Sprintf(dockerNetworkcmd, tempNetworkName)
+	// To print which command is running
```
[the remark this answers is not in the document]
samyak-jn: Yes, noted, that's why the PR is in WIP, I need to refactor the code once confirmed that the functionality to this is what we expect from the story.

backend-shared/config/db/postgres-integration.go
```diff
@@ -8,14 +8,18 @@ import (
 )
 
 // function to connect to Postgres
-func connectToDatabase(verbose bool) (*pg.DB, error) {
+func ConnectToDatabase(verbose bool, dbName string, args ...int) (*pg.DB, error) {
```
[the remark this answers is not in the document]
samyak-jn: Oh yes this is a mistake, I was using it as for testing purpose in another directory and forgot to roll it back, I'll change this, makes sense totally!

backend-shared/config/db/ephemeralPostgres.go
```diff
@@ -0,0 +1,210 @@
...
+// NewEphemeralCreateTestFramework will create a docker container,
+// network, connect to database and populate the schema. This is
+// merely defined for testing purpose
+func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
+	dockerName := "managed-gitops-postgres-test"
+	dockerNetworkcmd := "docker network create %s"
+	uuid := uuid.New().String()
+	tempDBName := "db-" + uuid
+	tempNetworkName := "gitops-net-" + uuid
+	s := fmt.Sprintf(dockerNetworkcmd, tempNetworkName)
```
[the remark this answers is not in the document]
samyak-jn: Again, from the lower comment I made this is by mistake committed, I have a commit in my local that fixes this, we have a function generate_uuid() already for this. I'll push a commit for this.
I was testing in my personal repository here samyak-jn/go-utilities, and that's where for ease I copied it directly.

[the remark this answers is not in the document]
samyak-jn: I'll research about it, but when I tested for these locally the network was getting created and deleting on the name generated by this seamlessly. 

```
➜ samjain@fedora  ~  docker network create gitops-net-dff7485d-25d1-4b0f-aa90-ba5af698dd7d
6683c7c745290d55633fe574ebfa09c97ebd7c15da74dc55c0b4f365f5b76dcb
➜ samjain@fedora  ~  docker network  ls                                                   
NETWORK ID     NAME                                              DRIVER    SCOPE
32dac68dd470   bridge                                            bridge    local
cd6a6008f06c   gitops-net                                        bridge    local
6683c7c74529   gitops-net-dff7485d-25d1-4b0f-aa90-ba5af698dd7d   bridge    local
2888161aca9d   host                                              host      local
25876db9a143   k3d-my-cluster                                    bridge    local
fcb0b2367961   none                                              null      local
➜ samjain@fedora  ~  docker network rm gitops-net-dff7485d-25d1-4b0f-aa90-ba5af698dd7d    
gitops-net-dff7485d-25d1-4b0f-aa90-ba5af698dd7d

```

backend-shared/config/db/ephemeralPostgres.go
```diff
@@ -0,0 +1,210 @@
...
+	dockerNetwork string
+}
+
+// NewEphemeralCreateTestFramework will create a docker container,
+// network, connect to database and populate the schema. This is
+// merely defined for testing purpose
+func NewEphemeralCreateTestFramework() (EphemeralDB, error) {
+	dockerName := "managed-gitops-postgres-test"
+	dockerNetworkcmd := "docker network create %s"
+	uuid := uuid.New().String()
```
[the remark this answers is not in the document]
samyak-jn: Again, from the lower comment I made this is by mistake committed, I have a commit in my local that fixes this, we have a function generate_uuid() already for this. I'll push a commit for this.
I was testing in my personal repository here samyak-jn/go-utilities, and that's where for ease I copied it directly.

backend-shared/config/db/ephemeralPostgres.go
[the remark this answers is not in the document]
samyak-jn: For the extra variable that I created (instead of s, that I removed now that is just used for printing purpose and will be removed) but for others, each tends to hold a meaning, let's say about above example, the same value in `tempNetworkName` is needed in another function as well to clean what's created and hence is separately created, but I'll skim once again to remove unwanted variable use.

backend-shared/config/db/ephemeralPostgres.go
```diff
@@ -0,0 +1,197 @@
...
+
+	fmt.Printf("* WARNING: Unsafe PostgreSQLDB object was created. You should never see this outside of test suites, or personal development.\n")
+	res := EphemeralDB{db: dbq, dbContainerID: strings.TrimSpace(string(dockerContainerID)), dockerNetwork: tempNetworkName}
+
+	return res, nil
+}
+
+// NewEphemeralCleanTestFramework will clean a docker container,
+// and network. This is merely defined for testing purpose
+func NewEphemeralCleanTestFramework(dockerContainerID string, tempNetworkName string) error {
```
jgwest, line 179: I think this function signature can change to...
```
func (ephemeralDB *EphemeralDB) Dispose() {
}
```
... and the parameters that are currently passed, can be retrieved from `EphemeralDB` struct.

backend-shared/config/db/ephemeralPostgres.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		// Todo: Once verified removed these print statements in order to maintain code coverage
+		fmt.Println("Docker Container ID: " + string(dockerContainerID))
+		return true, nil
+	})
+	if errWait != nil {
+		return EphemeralDB{}, fmt.Errorf("error in executing docker run command: %s", errWait)
+	}
+
+	dbcmd := "PGPASSWORD=gitops psql -h localhost -d %s -U postgres -p 6432 -c 'select 1'"
```
jgwest, line 95: For all cases where we are using psql: Rather than relying on `psql` to be installed on the user's machine, we can instead run psql in a container, similar to what we do with psql.sh: https://github.com/redhat-appstudio/managed-gitops/blob/254e180f93914f9260bc5cfea161bc84909d853c/psql.sh#L7

So for below, that might look something like this:
```go
psqlcmd := exec.Command("docker", "exec", "--rm", "--user", "postgres",  "-e" , "PGPASSWORD=gitops", "-it" "$POSTGRES_CONTAINER", "psql", /* ... */)
```

The `--rm` flag will remove the container as soon as it exists, so there is no need to cleanup after it.

backend-shared/config/db/postgres-integration.go
```diff
@@ -8,14 +8,18 @@ import (
 )
 
 // function to connect to Postgres
-func connectToDatabase(verbose bool) (*pg.DB, error) {
+func connectToDatabase(verbose bool, dbName string, args ...int) (*pg.DB, error) {
```
jgwest, line 11: I think we can change this function signature to
```
func connectToDatabase(verbose bool, dbName string, port int) (*pg.DB, error) {
```
The issue is that `...int` implies that we also support multiple ports (>=2), which we do not.

Another option would be to do this with 2 functions:
```go
func connectToDatabase(verbose bool, dbName string) (*pg.DB, error) {
    return connectToDatabase(vebose, dbName, DEFAULT_PORT)
}
```
and
```go
func connectToDatabaseWithPort(verbose bool, dbName string, port int) (*pg.DB, error) {
```

backend-shared/config/db/queries.go
```diff
@@ -199,8 +199,12 @@ type PostgreSQLDatabaseQueries struct {
 	allowUnsafe bool
 }
 
-func NewProductionPostgresDBQueries(verbose bool) (DatabaseQueries, error) {
-	db, err := connectToDatabase(verbose)
+func NewProductionPostgresDBQueries(verbose bool, args ...int) (DatabaseQueries, error) {
```
jgwest, line 202: Recommend creating two functions, so that the calling code can decide whether to specify a port, or not:
```golang

func NewProductionPostgresDBQueries(verbose bool) (DatabaseQueries, error) {
    return NewProductionPostgresDBQueriesWithport(verbose, DEFAULT_PORT)
}


func NewProductionPostgresDBQueriesWithPort(verbose bool, port int) (DatabaseQueries, error) {
    // As before
}
```

backend-shared/config/db/queries.go
```diff
@@ -215,8 +219,13 @@ func NewProductionPostgresDBQueries(verbose bool) (DatabaseQueries, error) {
 
 }
 
-func NewUnsafePostgresDBQueries(verbose bool, allowTestUuids bool) (AllDatabaseQueries, error) {
-	db, err := connectToDatabase(verbose)
+func NewUnsafePostgresDBQueries(verbose bool, allowTestUuids bool, args ...int) (AllDatabaseQueries, error) {
```
jgwest, line 222: Same as above.

jgwest (changes requested):
Looks great, some suggestions around function name changes, method signatures, and running psql within a container.

backend-shared/config/db/ephemeralPostgres.go
samyak-jn: Oh yes, silly me, this was the reason I changed the return signature, I'll update it :)

backend-shared/config/db/postgres-integration.go
samyak-jn: Totally makes sense, and yes we can do that as well. I'll fix this up quickly!

backend-shared/config/db/ephemeralPostgres.go
samyak-jn: I now configured psql command to run from the container side, just a quick note here, correct me if I'm wrong please but docker exec doesn't have rm flag right?

jgwest (approved):
LGTM, thanks @samyak-jn!

PR #63, opened 2022-01-31
Test dockerfile

drpaneas:
We used to have a bunch of broken Dockerfiles.
This PR makes sure the Dockerfile we have always builds.

See an example of a run, from my fork: https://github.com/drpaneas/managed-gitops/runs/5005481612?check_suite_focus=true

PS: this is the default template for Dockerfile building recommended by GitHub Actions.


samyak-jn (approved):
LGTM, :taco:

PR #65, opened 2022-02-08
Minor Refactoring in branch m2-demo-on-cluster's README

scoheb:


samyak-jn (approved):
Good catch!

samyak-jn:
Hi @scoheb, :wave: 

Can you add the commit from PR #66 here only, if that's possible? :smile: 
Also, let's keep this PR open until skim through all the steps one from the readme to add/refractor more if needed.

scoheb:
updated!

jgwest:
Good catch, thanks @scoheb  and @samyak-jn!

PR #67, opened 2022-02-10
Refactor database access to use unchecked behaviour by default

jgwest:
This PR:
- Refactors the `Unchecked*` functions, removing the `Unchecked` prefix. Thus calling the 'unchecked' method should now be the default way of accessing the database.
- Refactors the previously unprefixed functions, adding the `Checked` prefix. 
- Updates the comment at the top of queries.go to reflect the philosophy change around these queries.

## Why?

At the moment, we have a concept of checked/unchecked database methods in `queries.go`:
- *Checked*: Before a database query is performed, we check if the user has permission to issue this request.
- *Unchecked*: We do not perform the above check.

While this idea of 'check the database before it is used' works in principle, in practice I expect it to have a heavy performance impact, especially relative to the benefits it provides. We can probably better handle these same checks at a higher level, and with less impact.

This idea of checked/unchecked queries also increase the cognitive load of creating new database tables, and writing new code that interacts with the database.

My hope is that, in the future, we can re-examine our use of these, and incrementally integrate them into our codebase.

But for now, we should prefer unchecked functions, when writing new code, and I have marked the 'checked' functions as experimental only.

### DIFF
--- a/backend-shared/config/db/apicrtodatabasemapping.go
+++ b/backend-shared/config/db/apicrtodatabasemapping.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 )
 
-func (dbq *PostgreSQLDatabaseQueries) UncheckedDeleteAPICRToDatabaseMapping(ctx context.Context, obj *APICRToDatabaseMapping) (int, error) {
+func (dbq *PostgreSQLDatabaseQueries) DeleteAPICRToDatabaseMapping(ctx context.Context, obj *APICRToDatabaseMapping) (int, error) {
 	if err := validateQueryParamsEntity(obj, dbq); err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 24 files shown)


samyak-jn (commented):
Hi, just a little follow up on that, so, for now, are we agreeing to open a new database function with `unchecked` prefix for testing purposes et al, once verified we move it later? 

Also, was there any specific reason for leaving out these here: 
1. [uncheckedCreateOperation](https://github.com/jgwest/managed-gitops/blob/cc3d3252383348ea0c8bc56b9b33584ebb512ae6/backend/eventloop/application_event_runner.go#L1147)
2. [uncheckedWaitForOperationToComplete](https://github.com/jgwest/managed-gitops/blob/cc3d3252383348ea0c8bc56b9b33584ebb512ae6/backend/eventloop/application_event_runner.go#L1192)

jgwest:
@samyak-jn yah, from now on, when you create a new database function, it would be 'unchecked', with no prefix. For example, `GetIceCreamByID` to do an 'unchecked' get of the Ice Cream resource by it's primary key. 

RE: leaving those methods out, whoops, missed those!

samyak-jn (approved):
LGTM, @jgwest! Makes sense, and I'm rushing out to get an ice cream! :wink: :ice_cream:

drpaneas (approved):
/lgtm from my side as well.
I've tested it locally and doesn't break any functionality.

PR #71, opened 2022-02-25
Update database schema for manual sync

jgwest:
This PR:
- Adds `desired_state` to SyncOperation
- Add logic to handle deletion of a GitOpsDeploymentSyncRun CR; deletion of this API CR should cancel any sync operations that are running in Argo CD (but this sync cancellation logic will be in a separate PR).

This is a small chunk of a larger piece of work that I am still working on as part of manual sync. Because manual sync was depriotized in favour of automated sync, for this sprint, I am pushing this small piece as a PR so it doesn't get lost/out-of-sync with the rest of the codebase.

### DIFF
--- a/backend-shared/config/db/syncoperation.go
+++ b/backend-shared/config/db/syncoperation.go
@@ -5,6 +5,11 @@ import (
 	"fmt"
 )
 
+const (
+	SyncOperation_DesiredState_Running    = "Running"
+	SyncOperation_DesiredState_Terminated = "Terminated"
+)
+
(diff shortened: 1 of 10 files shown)


samyak-jn (approved):
LGTM, should we ship tests for these in another PR or so? :pie:

jgwest:
@samyak-jn yup, I'll be opening more stories for this shortly.

samyak-jn:
Thanks @jgwest!

PR #72, opened 2022-02-28
Manual sync

jgwest:
This PR:
- Adds Argo CD GRPC API client (`CredentialService`), and support for creating and caching sessions within that service
- Support for login/sync/terminate APIs
- Generated mocks of a few Argo CD API Client interfaces
- Unit tests for the above

This PR contains some commits from https://github.com/redhat-appstudio/managed-gitops/pull/71/files, which will need to go in first before this PR does.

Part of [GITOPSRVCE-82](https://issues.redhat.com/browse/GITOPSRVCE-82)


drpaneas (approved):
The code looks fine and the mock for unit-tests are great!

some concerns:

- [ ] how are going to maintain this code so it's compatible with the the future ArgoCD versions?
- [ ] is this client going to be used only for cluster-agent, right?

samyak-jn (approved):
Awesome 😎

jgwest:
Thanks @drpaneas!

> how are going to maintain this code so it's compatible with the the future ArgoCD versions?

This is definitely a valid concern. Concerns for compatibility are partially mitigated by:
- The GRPC API that the CLI is using does not change often. For example, you should be able to use the Argo CD v2.0 CLI with Argo CD v2.3.
- Once we have E2E tests, it should be easy for us to test our code against arbitrary versions of Argo CD. That way we can ensure that when we upgrade our Argo CD instance (for example v2.2.x to v2.3.x) that it doesn't break our GRPC API interface code.

Not perfect, but unfortunately Argo CD doesn't package this API into a way that is easily consumed (and other folks in the community have similarly complained about this). It would be cool to do something upstream to improve this later, but I think we don't have the bandwidth for it right now.

> is this client going to be used only for cluster-agent, right? if not maybe introduce a new pkg instead of utils.

Correct, cluster-agent only. Cluster-agent is responsible for interfacing with

PR #73, opened 2022-03-03
New 'appstudio-controller' component: Create GitOpsDeployments for corresponding AppStudio Applications

jgwest:
### This PR:
- Adds a new 'appstudio-controller' component to GitOps Service (plus changes to Makefile, Dockerfile, README.mds, and K8s deployment manifests)
- This component add initial support for interfacing with other AppStudio components, beginning with the HAS service. This will be the only component within the GitOps service that will vendor in other AppStudio components.
- New controller behaviour: when a new AppStudio `Application` resource is created, we create a corresponding `GitOpsDeployment` resource.
- Component bootstrapped with latest version of operator-sdk
- *Note*: No unit tests on this PR, since another team wants this PR merged ASAP (and because we need reviewers). Unit tests will be handled on a separate PR.

### How it works

We create/modify/delete a `GitOpsDeployment`, based on a corresponding parent `Application` resource:
```yaml
apiVersion: managed-gitops.redhat.com/v1alpha1
kind: GitOpsDeployment
metadata:
  name: (app name of Application)-deployment
  # if name is too long, we use a hash of the name instead
  namespace: (same namespace as Application)
spec:
  source:
    repoURL: (value from Application's Spec.GitOpsRepository.URL)
    path: (value from Application's Spec.GitOpsRepository.Context)
```

Fixes [GITOPSRVCE-97](https://issues.redhat.com/browse/GITOPSRVCE-97)


samyak-jn (commented):
Thanks, these changes LGTM @jgwest! :taco: 
Two quick things:
1. Other sub-repo (backend, backend-shared et al) uses go 1.16 so as a note we might need to bump them as well so to avoid any other conflicts 
2. Do we need to add this component for the existing GitHub actions we have for maintaining sanity checks?

appstudio-controller/Makefile
```diff
@@ -0,0 +1,223 @@
...
+# This recipe invokes 'opm' in 'semver' bundle add mode. For more information on add modes, see:
+# https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
+.PHONY: catalog-build
+catalog-build: opm ## Build a catalog image.
+	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+
+# Push the catalog image.
+.PHONY: catalog-push
+catalog-push: ## Push a catalog image.
+	$(MAKE) docker-push IMG=$(CATALOG_IMG)
```
drpaneas, line 223: comment those out as well to keep it consistent with the rest of the makefile components.

drpaneas:
> 2. Do we need to add this component for the existing GitHub actions we have for maintaining sanity checks?

most likely

samyak-jn (approved):

jgwest:
Agree with all review comments, and updated, thanks @drpaneas and @samyak-jn!

> Other sub-repo (backend, backend-shared et al) uses go 1.16 so as a note we might need to bump them as well so to avoid any other conflicts

Go 1.17 should be backwards compatible, but agreed we should move everything up together. Will do this as a separate PR (unless someone else beats me to it).

drpaneas:
Thx for the changes, looks perfect now :D

PR #79, opened 2022-03-11
Retrieve GitOps repository from devfile metadata of Application CR

jgwest:
This PR:
- Switches from using the GitOps repository URL from the AppStudio Application `spec` field (which is actually for _input_, as per API spec), to using the devfile metadata from the `status` field.


appstudio-controller/controllers/appstudio.redhat.com/application_controller.go
```diff
@@ -205,25 +213,71 @@ func validateApplication(asApplication applicationv1alpha1.Application) error {
...
+	}
+
+	var getErr error
+
+	// These strings are not defined as constants in the application service repo
+
+	metadata := curDevfile.GetMetadata()
+
+	// GitOps Repository is a required field
+	gitopsURL := metadata.Attributes.GetString("gitOpsRepository.url", &getErr)
```
sbose78, line 243: This is hard work ;)

sbose78 (approved):
Looks good to me. ( Assuming tests are being added separately 👍 )

PR #81, opened 2022-03-15
Refactor event loops into multiple packages, update names, add comments from Google Doc

jgwest:
This PR:
- Moves the application event loop into its own package (under event loops package)
- Splits the main application event loop file into multiple related files (one for handling GitOpsDeployments, one for handling GitOpsDeploymentSyncRuns)
- Adds comments from the Google Doc, to explain what each event loop is doing
- Miscellanous refactoring

There is more work that can be done, but this is a good start.


drpaneas (approved):
Beautiful work @jgwest , especially the _types_ file is very clean and digestible. I very much like this approach :D

I didn't test it locally, I assume it works. I just read through the code changes.
/lgtm

samyak-jn (approved):
Awesome work here tested locally for the test that we are shipping! :")

jgwest:
Thanks @drpaneas and @samyak-jn!

PR #84, opened 2022-03-16
Add comment description to task retry loop

jgwest:
This PR:
- Adds description to the task retry loop, and struct comment
- Fixes a magic number that should be using the existing constant value


samyak-jn (approved):
Thanks for writing this up, LGTM, @jgwest! :rocket:

drpaneas (approved):

PR #87, opened 2022-03-22
Add unit tests for sharedresourceloop

jparsai:
# Description
This PR is to add Unit test cases for sharedresourceloop.

## Type of changes 
1. Added unit test cases for following functions

- GetOrCreateClusterUserByNamespaceUID
- GetOrCreateSharedResources
- . GetGitopsEngineInstanceById

2. Updated signature of functions to return flags to identify whether resources were newly created or fetched from existing entry.
3. Used a cleanup function delete resources created by tests.

#### Changes done after review 
1. Upgraded Ginkgo to V2.0 in `backend` (Can be done for other components in separate PR.) for using `DeferCleanup` to clean up resources after each test.
2. Changed Ginkgo import statements in `backend`.
3. Replaced `RunSpecsWithDefaultAndCustomReporters` with `RunSpecs` in `backend/controllers/managed-gitops/suite_test.go`, as it's been deprecated in Ginkgo V2.

## Jira ticket 
https://issues.redhat.com/browse/GITOPSRVCE-109

### DIFF
--- a/backend-shared/config/db/util/utils.go
+++ b/backend-shared/config/db/util/utils.go
@@ -33,8 +33,9 @@ func GetGitOpsEngineSingleInstanceNamespace() string {
 	return DefaultGitOpsEngineSingleInstanceNamespace
 }
 
+// The bool return value is 'true' if ManagedEnvironment is created; 'false' if it already exists in DB or in case of failure.
 func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceNamespace v1.Namespace,
-	dbq db.DatabaseQueries, log logr.Logger) (*db.ManagedEnvironment, error) {
+	dbq db.DatabaseQueries, log logr.Logger) (*db.ManagedEnvironment, bool, error) {
 
 	workspaceNamespaceUID := string(workspaceNamespace.UID)
 
@@ -57,12 +58,12 @@ func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceN
 
 		if err == nil {
 			// Success: mapping exists, and so does the environment it points to
-			return &managedEnvironment, nil
+			return &managedEnvironment, false, nil
 		}
 
 		if !db.IsResultNotFoundError(err) {
 			// Failure: A generic error occurred.
-			return nil, fmt.Errorf("unable to retrieve resource mapping: %v", err)
+			return nil, false, fmt.Errorf("unable to retrieve resource mapping: %v", err)
 		}
 
 		// At this point, we found the mapping, but didn't find the ManagedEnvironment that the
@@ -70,11 +71,11 @@ func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceN
 
 		// Since the managed environment doesn't exist, delete the mapping; it will be recreated below.
 		if _, err := dbq.DeleteKubernetesResourceToDBResourceMapping(ctx, dbResourceMapping); err != nil {
-			return nil, fmt.Errorf("unable to delete K8s resource to DB mapping: %v", dbResourceMapping)
+			return nil, false, fmt.Errorf("unable to delete K8s resource to DB mapping: %v", dbResourceMapping)
 		}
 
 	} else if !db.IsResultNotFoundError(err) {
-		return nil, fmt.Errorf("unable to retrieve resource mapping: %v", err)
+		return nil, false, fmt.Errorf("unable to retrieve resource mapping: %v", err)
 	}
 
 	// At this point in the function, both the managed environment and mapping necessarily don't exist
@@ -90,7 +91,7 @@ func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceN
 		Serviceaccount_ns:           "serviceaccount_ns",
 	}
 	if err := dbq.CreateClusterCredentials(ctx, &clusterCreds); err != nil {
-		return nil, fmt.Errorf("unable to create cluster creds for managed env: %v", err)
+		return nil, false, fmt.Errorf("unable to create cluster creds for managed env: %v", err)
 	}
 
 	managedEnvironment := db.ManagedEnvironment{
@@ -98,7 +99,7 @@ func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceN
 		Clustercredentials_id: clusterCreds.Clustercredentials_cred_id,
 	}
 	if err := dbq.CreateManagedEnvironment(ctx, &managedEnvironment); err != nil {
-		return nil, fmt.Errorf("unable to create managed env: %v", err)
+		return nil, false, fmt.Errorf("unable to create managed env: %v", err)
 	}
 
 	dbResourceMapping = &db.KubernetesToDBResourceMapping{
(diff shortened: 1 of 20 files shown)


drpaneas (commented):
Since we have decided to adapt the Ginkgo testing framework, could you try adapting your `sharedresourceloop_test.go` into it?

jparsai:
@drpaneas Moved tests to ginkgo, PTAL.

backend-shared/config/db/util/utils.go
```diff
@@ -34,7 +34,7 @@ func GetGitOpsEngineSingleInstanceNamespace() string {
 }
 
 func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceNamespace v1.Namespace,
-	dbq db.DatabaseQueries, log logr.Logger) (*db.ManagedEnvironment, error) {
+	dbq db.DatabaseQueries, log logr.Logger) (*db.ManagedEnvironment, error, bool) {
```
jgwest, line 37: Can you add a quick comment like this: "the bool return value is true if the managedenvironment was created; false otherwise."
(that way it's obvious what the bool return value is for, for the next person to look at this function)

backend-shared/config/db/util/utils.go
```diff
@@ -109,22 +109,22 @@ func GetOrCreateManagedEnvironmentByNamespaceUID(ctx context.Context, workspaceN
...
-	return &managedEnvironment, nil
+	return &managedEnvironment, nil, true
 }
 
 // GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID gets (or creates it if it doesn't exist) a GitOpsEngineInstance database entry that
 // corresponds to an GitOps engine instance running on the cluster.
 func GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID(ctx context.Context,
 	gitopsEngineNamespace v1.Namespace, kubesystemNamespaceUID string,
-	dbq db.DatabaseQueries, log logr.Logger) (*db.GitopsEngineInstance, *db.GitopsEngineCluster, error) {
+	dbq db.DatabaseQueries, log logr.Logger) (*db.GitopsEngineInstance, bool, *db.GitopsEngineCluster, error) {
```
jgwest, line 122: As above, we can add a quick comment to the function comment, indicating what the bool is for.

backend/eventloop/shared_resource_loop/sharedresourceloop.go
```diff
@@ -108,8 +108,8 @@ func (srEventLoop *SharedResourceEventLoop) GetGitopsEngineInstanceById(ctx cont
 // Ensure the user's workspace is configured, ensure a GitOpsEngineInstance exists that will target it, and ensure
 // a cluster access exists the give the user permission to target them from the engine.
 func (srEventLoop *SharedResourceEventLoop) GetOrCreateSharedResources(ctx context.Context,
-	workspaceClient client.Client, workspaceNamespace corev1.Namespace) (*db.ClusterUser,
-	*db.ManagedEnvironment, *db.GitopsEngineInstance, *db.ClusterAccess, error) {
+	workspaceClient client.Client, workspaceNamespace corev1.Namespace) (*db.ClusterUser, bool,
```
jgwest, line 111: As above, add a quick comment mentioning that the bools refer to whether the particular object has been created.

backend/eventloop/shared_resource_loop/sharedresourceloop.go
```diff
@@ -359,21 +397,21 @@ func internalProcessMessage_GetOrCreateSharedResources(ctx context.Context, work
...
 	}
 
 	kubeSystemNamespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "kube-system", Namespace: "kube-system"}}
 	if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(kubeSystemNamespace), kubeSystemNamespace); err != nil {
-		return nil, fmt.Errorf("unable to retrieve kube-system namespace in determineGitOpsEngineInstanceForNewApplication")
+		return nil, false, nil, fmt.Errorf("unable to retrieve kube-system namespace in determineGitOpsEngineInstanceForNewApplication")
 	}
 
-	gitopsEngineInstance, _, err := dbutil.GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID(ctx, *namespace, string(kubeSystemNamespace.UID), dbq, log)
+	gitopsEngineInstance, isNewInatance, gitopsEngineCluster, err := dbutil.GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID(ctx, *namespace, string(kubeSystemNamespace.UID), dbq, log)
```
jgwest, line 412: Typo 'isNewInatance' :smile:

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+			gitopsDepl := &managedgitopsv1alpha1.GitOpsDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-gitops-depl",
+					Namespace: workspace.Name,
+					UID:       uuid.NewUUID(),
+				},
+			}
+
+			informer := sharedutil.ListEventReceiver{}
```
jgwest, line 53: It looks like the ProxyClient/ListEventReceiver functionality isn't being used here (eg you are not reading from `Events` inside that struct for any test purposes), so you can probably just use the k8sClientOuter directly.

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+	Context("Shared Resource Event Loop test", func() {
+
+		It("Should create or fetch a user by Namespace id.", func() {
+			// Create a fake K8s client
+			ctx := context.Background()
+
+			scheme,
+				argocdNamespace,
+				kubesystemNamespace,
+				workspace, _ := eventlooptypes.GenericTestSetup()
```
jgwest, line 43: Should expect on the err (the '_', here); same as with other places where GenericTestSetup is called.

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+				isNewClusterAccess,
+				gitopsEngineClusterOld,
+				err := sharedResourceEventLoop.GetOrCreateSharedResources(ctx, k8sClient, *workspace)
+
+			// Check if result is not Nil and error is Nil
+			Expect(err).To(BeNil())
+			Expect(clusterUserOld).NotTo(BeNil())
+			Expect(managedEnvOld).NotTo(BeNil())
+			Expect(gitopsEngineInstanceOld).NotTo(BeNil())
+			Expect(clusterAccessOld).NotTo(BeNil())
```
jgwest, line 150: Do we need to check 'gitopsEngineClusterOld', too? Noticed it was missing

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+			k8sClient := &sharedutil.ProxyClient{
+				InnerClient: k8sClientOuter,
+				Informer:    &informer,
+			}
+
+			sharedResourceEventLoop := &SharedResourceEventLoop{inputChannel: make(chan sharedResourceLoopMessage)}
+
+			go internalSharedResourceEventLoop(sharedResourceEventLoop.inputChannel)
+
+			// At first assuming there are no existing users, hence creating new.
```
drpaneas, line 69: Can you explain a bit what is this user that you expect to exist and why it's needed?

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+			usrOld,
+				err,
+				isNewUser := sharedResourceEventLoop.GetOrCreateClusterUserByNamespaceUID(ctx, k8sClient, *workspace)
+
+			// Check error in Nil and user is not Nil
```
drpaneas, line 74: redundant comment, the Ginkgo code is self-explanatory. You can rephrase the comment to write something describing the behavior you are testing, e.g. "Check if user has been created without problems".

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+			// Check new user is created
+			Expect(isNewUser).To(BeTrue())
+
+			// User is created in previous call, then same user should be returned instead of creating new.
+			usrNew,
+				err,
+				isNewUser := sharedResourceEventLoop.GetOrCreateClusterUserByNamespaceUID(ctx, k8sClient, *workspace)
+
+			// Check error in Nil and user is not Nil
```
drpaneas, line 86: Same here, plus minor typo _is Nil_.

jgwest (changes requested):
Looks great, some minor tweaks then good to go!

Using Ginkgo's `AfterEach` to do the cleanup for each test was great thinking (and of course this is what AfterEach is designed for), I will remember to suggest this to folks for future stories :+1:

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
jgwest: This is the cluster user database entity, which corresponds to a user of the gitops service (e.g. someone who is invoking the GitOps Service API, who owns all the related GitOpsDeployment et al CRs, the corresponding DB entries, and thus is using GitOps Service/Argo CD to manage deploying to clusters).

drpaneas:
Two things from me as well:

1. The same way you use 'AfterEach', you can also you 'BeforeEach' in order to avoid copy pasting the code for the creation of a fake kubernetes client.
2. This is pretty stylistic, so quite minor change: in Go, the `err` is usually placed as the last argument.

backend/eventloop/shared_resource_loop/sharedresourceloop_test.go
jparsai: It is not used anywhere for test, but it it is created prior to GitOps Engine Instance, so it needs to be deleted while cleaning up resources, that's why it is returned back to test and then sent to `DeferCleanup` function.
https://github.com/redhat-appstudio/managed-gitops/blob/8c72ce9bc207c389e260a2808514083accbe42cf/backend-shared/config/db/util/utils.go#L125

jgwest (approved):
Looks great, thanks again @jparsai!

PR #88, opened 2022-03-22
Add cache implementation for the ApplicationInfo (Application + ApplicationState) (Solves: GitOpsrvce-110)

samyak-jn:
Hello folks, :hugs: 

**Why this PR, why do we need a cache for ApplicationState, if so why not for others?**

The purpose of this Pull Request is that we do not want to overwhelm the controller (or the database) with various operations frequently (most of the time these reconciliations are changes that might not be useful for us). To resolves the frequent, CRUD operations, the cache is being implemented in backend-shared which will be used by cluster-agent controller reconciliation.
Currently, the changes that the controller makes to the database are for updating ApplicationState (the health and sync status) and that's why particularly the cache implementation for the ApplicationState is specifically needed..

**What are the changes made in this PR? Feature, Refactoring? etc**
The application_state_cache is created under the config/db/util package, the proposal to change it into config/db/cache is being discussed atm. The test for the same file is shipped along with it. The main features are creating, getting, updating, and deleting the application state from the cache.

There are some refactoring as well done with config/db package, the testSetup, testTeardown, and isEmpty functions are now exported to use with other tests as well (needed for appklication_state_cache_test.go).
This refactoring is also made to resolve an issue that the team is facing with the testing stories.

~~Note: Currently, this PR is in WIP, the integration of cache with cluster-agent is to be taken care of.~~


samyak-jn (commented):
Adding some changes from the last comment as per the discussion in the meeting,
The application_state_cache is now changed to application_info_cache, the changes in the nomenclature are done to achieve wider info in the cache and are just not about ApplicationState.

Moving forward with the story, it was found out that the reconciler method was looking for Applications using `DB.GetApplicationByID(..)` hence, we needed a reach cache functionality for ApplicationTable as well.
So the workaround to it is, the following:

The cache is now ApplicationInfoCache (s/ApplicationStateCache/ApplicationInfoCache/g)
```
type ApplicationInfoCache struct {
	channel chan applicationInfoCacheRequest
}

type applicationInfoCacheRequest struct {
	ctx                          context.Context
	msgType                      ApplicationInfoCacheMessageType
	createOrUpdateAppObject      db.Application
	createOrUpdateAppStateObject db.ApplicationState
	primaryKey                   string
	responseChannel              chan applicationInfoCacheResponse
}

type applicationInfoCacheResponse struct {
	application      db.Application
	applicationState db.ApplicationState
	err              error

	// valueFromCache is true if the value that was returned came from the cache, false otherwise.
	valueFromCache bool

	// Delete only: rowsAffectedForDelete contains the number of rows that were affected by the deletion oepration
	rowsAffectedForDelete int
}
```
For the same relevant methods are created and tests are shipped.

drpaneas:
> Note: Currently, this PR is in WIP, the integration of cache with cluster-agent is to be taken care of.

Please signal this either with putting WIP in the title or converting it into a draft, so it's more emphatic.

samyak-jn:
> Please signal this either by putting WIP in the title or converting it into a draft, so it's more emphatic.

The PR is up for review, there are gosec warnings (one from other PRs) that's the only reason CI is failing. I'll address those once we have an initial review done for this.

jgwest (changes requested):
Looks great, thanks @samyak-jn!

I'm completed an initial review below. 

Unfortunately, looking at the story, I've found an issue with the logic I originally proposed :disappointed:. 

With my original implementation/proposal, it seems like old entries are never removed from the applicationstate cache, and there is no obvious way to do so without other consequences. I think I have a solution, but I'm still poking at it to make sure it doesn't have any logic holes. 

In the mean time, feel free to respond to this first set of review comments, and I'll likely have some additional logic suggestions shortly! Stay tuned.

samyak-jn (commented):
Add a commit to fix for gosec warning (G601: Implicit memory aliasing in for loop.) This warning is coming from the recent test merges, have addressed the fix in the same PR.

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,358 @@
...
+		err: err,
+	}
+
+}
+
+func processUpdateAppStateMessage(dbQueries db.DatabaseQueries, req applicationInfoCacheRequest, cache map[string]db.ApplicationState) {
+
+	err := dbQueries.UpdateApplicationState(req.ctx, &req.createOrUpdateAppStateObject)
+
+	if err != nil {
```
jgwest, line 244: Think this should be `if err == nil`

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,358 @@
...
+			continue
+		}
+
+	}
+}
+
+func processCreateAppStateMessage(dbQueries db.DatabaseQueries, req applicationInfoCacheRequest, cache map[string]db.ApplicationState) {
+	err := dbQueries.CreateApplicationState(req.ctx, &req.createOrUpdateAppStateObject)
+
+	if err != nil {
```
jgwest, line 228: Think this should be `if err == nil`

jgwest (changes requested):

backend-shared/config/db/util/application_info_cache.go
samyak-jn: Silly me, yes! :)

jgwest:
Changes looks good, thanks @samyak-jn 

As mentioned on [my previous comment](https://github.com/redhat-appstudio/managed-gitops/pull/88#pullrequestreview-920199726), I've found some logic issues with the original story logic I proposed on GITOPSRVCE-110. 

So the general problem in the proposed logic is: once an entry is added to the cache, how is it removed?

With the current code / story logic:
- ApplicationState will only be removed from the cache when delete is called.
- Application will never be removed.

Both cases will lead to a stale cache, where, over time, the values in cache will no longer reflect the contents of the database (causing unexpected behaviour). For example: the 'backend' component may delete an Application from the database, but or cluster-agent cache will never be updated.

The situation to avoid is:
1) (Application database entry exists)
2) 'application_controller.go' does a GetApplication on the cache, which adds the Application to the cache.
3) The Application is removed from the database, by some code running in *backend* component.
4) 'application_controller.go' does another GetApplication on the cache: BUT, the Application is still in the cache, so the cache will just return the cached value (even though the Application doesn't actually exist in the database, because it was removed in step 3).

The problem is there is no way to detect when the Application has been deleted from the database, so that we can invalidate our cache. 


## Solution:

My proposed solution is to update the cache code in as follows:
- expire the cache entries after 1 minute: this prevents them from getting too stale (getting too out of date), and prevents old cache entries from never being cleaned up (and thus slowly occupying more and more memory)
- invalidate the cache when an error occurs: this lets us catch the case where backend has deleted an Application. (creating/updating an applicationstate table will automatically fail if the application has been deleted, due to the foreign key relationship on application)


### 1) Add a new field to the cache, which indicates when the cache entry should expire (e.g. 1 minute after creation)

For example, change:
```go
cacheAppState := map[string]db.ApplicationState{}
```
to:
```go
cacheAppState := map[string]applicationStateCacheEntry{}

with a new struct now containing both the state, and the expire time.
type applicationStateCacheEntry struct {
	appState db.ApplicationState
	cacheExpireTime time.Time // after this time, the entry should be removed from the cache.
	
}
```

Likewise need to update the cacheApp map as well, in the same way.

### 2) Every minute, the 'applicationInfoCacheLoop' goroutine should expire old cache entries

How:
- Create a Go timer which sends a message to 'applicationInfoCacheLoop', telling it to expire old cache entries. (See `startNewStatusUpdateTimer` for an example of a timer that sends an internal message to a goroutine, telling it to perform an action)
- When the timer expires, the aICLoop should iterate through the cache and remove any entries from `cacheAppState` and `cacheApp` that are >1 minute old.
- Once the old entries have been removed, start a new timer.

### 3) If a get/create/update operation on the database fails for an ApplicationState, then remove the corresponding _Application_ entry from the cache.

If a get/create/update operation on the database fails for an ApplicationState (in processGet/processCreate/processUpdate), then remove the corresponding Application entry from the cache.

For example, here is how this might look for `processUpdateAppStateMessage`:

```go
func processUpdateAppStateMessage(dbQueries db.DatabaseQueries, req applicationInfoCacheRequest, stateCache map[string]applicationStateCacheEntry, appCache map[string]applicationCacheEntry) {

	err := dbQueries.UpdateApplicationState(req.ctx, &req.createOrUpdateAppStateObject)

	if err == nil {
		// Update the cache on success
		var appState db.ApplicationState = req.createOrUpdateAppStateObject
		cache[req.primaryKey] = appState
	} else {
		// An error occurred, so remove the cache entry from both the application state, and the application,
		// so that it can be re-acquired.
		// A common error returned by dbQueries above, will be that the applicate state cannot be updated, because the application no 
		// longer exists in the database. 
		// This is normal, and occurs when a GitOpsDeployment is deleted (and then the corresponding Application/Application state are deleted as well).
		// In this case, we need to remove the Application/ApplicationState from the DB, as they have likely been deleted (and thus should no longer be cached.)
		delete(stateCache, req.createOrUpdateAppStateObject. Applicationstate_application_id)
		delete(appCache, req.createOrUpdateAppStateObject. Applicationstate_application_id)			
		
	}

	req.responseChannel <- applicationInfoCacheResponse{
		err: err,
	}

}
```

jgwest:
Expanding on 2) above.

# Every minute, the 'applicationInfoCacheLoop' goroutine should expire old cache entries

The items in the code to add/update would be:

#### Add a new message to the list of messages that the loop function receives, for example:
```
ApplicationStateCacheMessage_ExpireStateCacheEntries
```

#### Next, create a new function that starts the timer in a new goroutine, and when the timer expired, send that message to the loop:
```golang
func (asc *ApplicationInfoCache) startTimer() {
	go func() {
		// Up to 1 second of jitter
		// #nosec
		jitter := time.Duration(int64(time.Millisecond) * int64(rand.Float64()*1000))

		// Wait 60 seconds (plus a little) for the timer to complete
		statusUpdateTimer := time.NewTimer(time.Second*60 + jitter)
		<-statusUpdateTimer.C

		// Send the message, indicating its time to expire the old cache entries
		asc.channel <- applicationInfoCacheRequest{
			msgType: ApplicationStateCacheMessage_ExpireStateCacheEntries,
			// (...)
		}
		
	}()
}
```

In `applicationInfoCacheLoop`:
```
- Add a call to startTimer() to the first line of 'applicationInfoCacheLoop'
- Create a new `processExpireStateCacheEntriesMessage` function (see below), and add it to the existing if block in `applicationInfoCacheLoop`: it should be called whenever a ExpireStateCacheEntries message is received, similar to how the other messages are called. `processExpireStateCacheEntriesMessage(cacheAppState, cacheApp)`
```

Finally, in `processExpireStateCacheEntriesMessage`:
```golang
func processExpireStateCacheEntriesMessage(cacheAppState, cacheApp, ...) {
	// TODO: iterate through the cacheAppState map and remove any entries that are older than 60 seconds, using the `cacheExpireTime` value
	// TODO: iterate through the cacheApp map and remove any entries that are older than 60 seconds, using the `cacheExpireTime` value
	// ...
	// Once the above is complete, start the timer again:
	startTimer()
}
```

backend-shared/config/db/util/application_info_cache_types.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
+)
+
+type ApplicationInfoCacheMessageType int
+
+type applicationStateCacheEntry struct {
+	appState        db.ApplicationState
+	cacheExpireTime time.Time // after this time, the entry should be removed from the cache.
```
jgwest, line 14: Looks like you are not currently setting `cacheExpireTime` in your PR (correct me if I am wrong). When the entry is added to the cache, `cacheExpireTime` should be set to `time.Now().Add(1 * time.Minute)`.

This allows the logic that you added elsewhere to work, because we can know when the cache was added using this value.

backend-shared/config/db/util/application_info_cache_types.go
```diff
@@ -0,0 +1,46 @@
...
+
+}
+
+type applicationCacheEntry struct {
+	app             db.Application
+	cacheExpireTime time.Time // after this time, the entry should be removed from the cache.
```
jgwest, line 20: Need to set this, as above.

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,411 @@
...
+		Applicationstate_application_id: req.primaryKey,
+	}
+
+	var err error
+	var valueFromCache bool
+
+	if db.IsEmpty(req.primaryKey) {
+		log.Error("PrimaryKey should not be nil: " + req.primaryKey + " not found")
+	} else {
+		delete(cacheAppState, req.createOrUpdateAppStateObject.Applicationstate_application_id)
```
jgwest, line 291: These two lines in the else branch of this if statement will cause the cache to be cleared after each GetAppStateMessage (which defeats the purpose of a cache).

Presuming I understand/recall the logic correctly, I think these deletes should move below.

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,411 @@
...
+		delete(cacheApp, req.createOrUpdateAppStateObject.Applicationstate_application_id)
+	}
+
+	res, exists := cacheAppState[appState.Applicationstate_application_id]
+	if !exists {
+		// If it's not in the cache, then get it from the database
+		valueFromCache = false
+		err = dbQueries.GetApplicationStateById(req.ctx, &appState)
+		if err != nil {
+			appState = db.ApplicationState{}
```
jgwest, line 301: Deletes mentioned above should move into this if branch.

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,411 @@
...
+		Application_id: req.primaryKey,
+	}
+
+	var err error
+	var valueFromCache bool
+
+	if db.IsEmpty(req.primaryKey) {
+		log.Error("PrimaryKey should not be nil: " + req.primaryKey + " not found")
+	} else {
+		delete(cacheApp, req.createOrUpdateAppStateObject.Applicationstate_application_id)
```
jgwest, line 339: Just like with `processGetAppStateMessage`, I think these deletes can be moved below.

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,411 @@
...
+		delete(cacheAppState, req.createOrUpdateAppStateObject.Applicationstate_application_id)
+	}
+
+	if res, exists := cacheApp[app.Application_id]; !exists {
+		// If it's not in the cache, then get it from the database
+		valueFromCache = false
+		err = dbQueries.GetApplicationById(req.ctx, &app)
+		if err != nil {
+			app = db.Application{}
```
jgwest, line 348: Deletes move inside this if block

jgwest (changes requested):

backend-shared/config/db/util/application_info_cache.go
```diff
@@ -0,0 +1,411 @@
...
+		valueFromCache: valueFromCache,
+		err:            err,
+	}
+
+}
+
+func processExpireCacheEntriesMessage(cacheApp map[string]applicationCacheEntry, cacheAppState map[string]applicationStateCacheEntry, inputChan chan applicationInfoCacheRequest) {
+
+	for key, elements := range cacheApp {
+		if time.Duration(elements.cacheExpireTime.Minute()) > time.Minute {
```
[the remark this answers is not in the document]
samyak-jn: Awesome, it will same one extra check, and it works!
image

samyak-jn (commented):
On the quick look where CI is [failing](https://github.com/redhat-appstudio/managed-gitops/runs/6290271860?check_suite_focus=true#step:5:365), looks like the backend test problem, will take a look why it is happening otherwise immediate functionalities are added.

jgwest:
@samyak-jn: if you rebase to latest it should be fixed. FYI GitHub is showing 'This branch cannot be rebased due to conflicts', so I can't merge until this is resolved.

jgwest:
I'm merging now to unblock the other PRs, but feel free to shout out anything I broke @samyak-jn!

Most of these changes were stuff I found while investigating race conditions when running `make test`. I was really confused as to why different database tests were failing at different times, but it turns out I needed to add `-p=1` to the `make test` target in the Makefile. (See my Slack complaint about this :upside_down_face: )

Changes in my commit:
- **utils.go**: Miscellaneous tweaks, like changing the deletion order, fixing the assert.equal order, failing fast at the end of the function, and not 'return'ing until the end of the function. This was stuff was I discovered while debugging.
- **application_info_cache_test.go**: Shutdown the cache loop at the end of every test (not sure if this is actually needed, but can't hurt), plus a couple logic tweaks (let me know if I accidentally broke anything here!).
- **db-schema.sql**: added additional comments on deletion order
- **application_info_cache.go**: Added support for shutting down the cache loop (only potentially useful for unit tests, to avoid race conditions), switched to using "sigs.k8s.io/controller-runtime/pkg/log" as the logger, updated the cache logic to always replace the cache on create/update, changed the primary key check to return an error (ultimately I rethought whether it should throw an error or not, from my original code).
    -  Likewise, let me know if I accidentally broke any logic here :sweat_smile: 
- **Makefiles**: remove per-package test parallelism, because it breaks our database tests :P

jgwest (approved):
Looks great, thanks @samyak-jn ! This story turned out to be a lot more complex than originally expected, thanks to cache behaviour :eyes: :smile:  

(Ensuring that cache behaviour is correct is one of the ongoing, long-running problems throughout my entire career :thinking: )

samyak-jn:
Good to know that, thanks for assigning this story to me @jgwest! Sorry took a little time, but got to learn many things on the go. Taking more to learn on goroutines and more good code refactoring as to-do items! :heart:

PR #91, opened 2022-03-28
Add Log Statements for better debugging.

tirthyme:


jgwest:
@tirthyme - You may have already fixed this, but just in case, to save you some time, the error we are seeing in the logs is due to this:

The line below (as an example) is incorrect:
- The Info function doesn't work like fmt.Sprintf, or like fmt.Println, rather it works differently from both those.
- The Info function doesn't support `%v` or `%s` (etc) style values, they need to be passed as pairs of parameters, after the first parameter.
- The Info function requires values that you want to print to be in pairs e.g. `, "name of param", param`.

Bad:
```go
log.V(util.LogLevel_Debug).Info("Deleted KubernetesResourceToDBResourceMapping: %v", dbResourceMapping)
```

Good:
```go
log.V(util.LogLevel_Debug).Info("Deleted KubernetesResourceToDBResourceMapping", "dbResourceMapping", dbResourceMapping)
// or
log.V(util.LogLevel_Debug).Info("Deleted KubernetesResourceToDBResourceMapping", "dbResourceMapping", dbResourceMapping, "user's user name:", userName)
```
After the first parameter of Info, there should be an even number of parameters. Or said another way, there should be 0 or more pairs of parameters.

The pair is:
- name describing the parameter
- value of parameter.
- Example:
    -  name: `"dbResourceMapping"`
    -  value: `dbResourceMapping`

jgwest:
@tirthyme It looks like your commits have been mixed up with some of the other commits that have gone in (for example, due to a `git merge`). 

I have pulled out the contents of your commits and rebased them onto HEAD of main, is there anything missing from here:
https://github.com/redhat-appstudio/managed-gitops/compare/main...jgwest:logging?expand=1

If not, I'll push that branch to your PR branch, which should fix the issue.

tirthyme:
Includes everything, Jonathan. Thanks!

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -457,6 +468,8 @@ func (a applicationEventLoopRunner_Action) cleanOldGitOpsDeploymentEntry(ctx con
 	} else if rowsDeleted == 0 {
 		// Log the error, but continue
 		log.V(sharedutil.LogLevel_Warn).Error(nil, "unexpected number of rows deleted for application", "rowsDeleted", rowsDeleted, "appId", deplToAppMapping.Application_id)
+	} else {
```
jgwest, line 471: We've already got `log.Info("deleting database Application, id: " + deplToAppMapping.Application_id)` a few lines above in this file, so probably don't need to also include this line, unless you think it has a value add.

jgwest (changes requested):
Looks great @tirthyme , just some minor tweaks requested. Let me know when that's done, and I'll test it one more time and should be good to go.

One global comment:
- I didn't add this comment to the individual lines, but I think you can remove `log.V(util.LogLevel_Debug).` from each of the new log lines you've added. We want to log at the info level, rather than at the debug level, for these items (unless you think they are too verbose, for example, that we are printing some lines too often.)

jgwest (approved):
LGTM, thanks again @tirthyme!

PR #96, opened 2022-04-11
Add Makefile ArgoCD upstream deployment

drpaneas:
#### Description:
There is no related JIRA ticket for this.
I think it's useful for some quick verification locally and for demo purposes.
Think of it as a bonus/optional functionality provided by the Makefile.


jgwest (approved):
LGTM, thanks @drpaneas!

PR #99, opened 2022-04-19
Sanitize inputs before performing update/insert operations into DB.

jparsai:
# Description
This PR will add a functionality to validate length of user input data before inserting/updating data into DB tables.

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-62

## Type of changes

- Created and used `ValidateFieldLength` function before database insert/update operations are performed 
- Added unit test case for new functions, also updated existing test cases to avoid decrease in test coverage due to new function calls added at the places where database insert/update operations are performed.
- Updated `application_event_runner_deployments.go` and `application_event_runner_syncruns.go`, logic to set condition in `GirOpsDeployment.status` is added, if an error occurred while inserting/updating a database entry.
- Changed a variable name in `config/db/operations_test.go` from `Default` to `DefaultValue` to avoid this error `Default redeclared during import 'github.com/onsi/gomega'`


backend-shared/config/db/apicrtodatabasemapping_test.go
```diff
@@ -57,4 +58,8 @@ func TestApiCRToDBMappingFunctions(t *testing.T) {
 	err = dbq.GetDatabaseMappingForAPICR(ctx, &fetchRow)
 	assert.True(t, IsResultNotFoundError(err))
 
+	// Set the invalid value
+	item.APIResourceName = strings.Repeat("abc", 100)
+	err = dbq.CreateAPICRToDatabaseMapping(ctx, &item)
+	assert.Error(t, err)
```
jgwest, line 64: For this `assert.Error()` (and others like it in your PR), I  recommend checking that the err string contains "value exceeds maximum size", to verify that your `validateFieldLength` logic is working as expected.

Something like:
```go
if err != nil {
 assert.True(t, strings.Contains(err.Error(), "value exceeds maximum size")
}
```

backend-shared/config/db/application_test.go
```diff
@@ -95,6 +96,11 @@ func TestCreateApplications(t *testing.T) {
 		return
 	}
 	assert.Equal(t, applicationput, applicationget)
+
+	// Set the invalid value
+	applicationput.Name = strings.Repeat("abc", 100)
+	err = dbq.CreateApplication(ctx, &applicationput)
+	assert.Error(t, err)
```
jgwest, line 103: Recommend checking the error string in all these cases, as above.

backend-shared/config/db/utils.go
```diff
@@ -168,3 +169,43 @@ func generateUuid() string {
 func isEmpty(str string) bool {
 	return len(strings.TrimSpace(str)) == 0
 }
+
+// A generic function to validate length of string values in input provided by users.
+// The max length of string is checked using constant variables defined for each type and field in db_field_constants.go
+func ValidateFieldLength(obj interface{}) error {
```
jgwest, line 175: Since this function isn't used by other Go packages, it doesn't need to be exported: it can be renamed to `validateFieldLength`.

backend-shared/config/db/utils_test.go
```diff
@@ -84,3 +88,52 @@ func TestIsEmptyValues(t *testing.T) {
 		})
 	}
 }
+
+var _ = Describe("Utils Test", func() {
+	Context("Utils test", func() {
```
jgwest, line 93: Can change this to `Context("Ensure ValidateFieldLength reports database rows that are too long", func() {`, just to give more context to the test.

backend/eventloop/application_event_loop/application_event_runner_test.go
```diff
@@ -160,6 +161,59 @@ var _ = Describe("ApplicationEventLoop Test", func() {
...
+					Name:      strings.Repeat("abc", 100),
+					Namespace: workspace.Name,
+					UID:       uuid.NewUUID(),
+				},
+			}
+
+			informer := sharedutil.ListEventReceiver{}
+
+			k8sClientOuter := fake.NewClientBuilder().WithScheme(scheme).WithObjects(gitopsDepl, workspace, argocdNamespace, kubesystemNamespace).Build()
+			k8sClient := &sharedutil.ProxyClient{
```
jgwest, line 181: I think you aren't using the ProxyClient/ListenEventReceiver functionality in this test? So this could be removed.

backend/eventloop/application_event_loop/application_event_runner_test.go
```diff
@@ -241,7 +295,92 @@ var _ = Describe("ApplicationEventLoop Test", func() {
...
+				Spec: managedgitopsv1alpha1.GitOpsDeploymentSyncRunSpec{
+					GitopsDeploymentName: gitopsDepl.Name,
+					RevisionID:           "HEAD",
+				},
+			}
+
+			informer := sharedutil.ListEventReceiver{}
+
+			k8sClientOuter := fake.NewClientBuilder().WithScheme(scheme).WithObjects(gitopsDepl, gitopsDeplSyncRun, workspace, argocdNamespace, kubesystemNamespace).Build()
+			k8sClient := &sharedutil.ProxyClient{
```
jgwest, line 328: Likewise, think ProxyClient/ListEventReceiver can be removed.

backend-shared/config/db/db_field_constants.go
```diff
@@ -90,3 +90,90 @@ func TruncateVarchar(s string, maxLength int) string {
...
+	"SyncOperationApplicationID":                                        SyncOperationApplicationIDLength,
+	"SyncOperationOperationID":                                          SyncOperationOperationIDLength,
+	"SyncOperationDeploymentName":                                       SyncOperationDeploymentNameLength,
+	"SyncOperationDeploymentNameField":                                  SyncOperationDeploymentNameLength,
+	"SyncOperationRevision":                                             SyncOperationRevisionLength,
+	"SyncOperationDesiredState":                                         SyncOperationDesiredStateLength,
+}
+
+// Get value of constants based on constant variable name given as String.
+func GetConstantValue(variable string) int {
```
jgwest, line 177: Since this function isn't used by any other packages (AFAICT), it doesn't need to be exported, and can be renamed to `getConstantValue`.

backend-shared/config/db/db_field_constants.go
```diff
@@ -90,3 +90,90 @@ func TruncateVarchar(s string, maxLength int) string {
...
 	return s
 }
+
+// A Map object to be used by utils.ValidateFieldLength function.( Map[<Constant Variable Name as String>] <Constant Variable> )
+// At run time we need to check if number of characters in user input are within the max limit defined by constant variables above.
+// Because constant variable names (Ex: ClusterUserUserNameLength) are not same as field names (Ex: ClusterUser.User_Name) of object,
+// we need to format field name to get constant variable name we need (i.e <ObjectType><FieldName><Length>),
+// but issue here that after formating we will get constant variable name as a String, and golang does not support eval() functionality similar to Node, Python etc,
+// we need to create a Map[<Constant Variable Name as String>] <Constant Variable> object.
+var dbFieldMap = map[string]int{
```
jgwest, line 100: This is a good solution, and thanks for the detailed comment here!

I think we need a CI check to make sure that developers don't forget to add new database fields to this list. I can easily see someone adding a field to the `db-schema.sql` file, and to the constants above, but forgetting to update this map.
- Why? If we don't do this, then when we add a new field to the database, we may forget to update the map, and the length value for the field will be returned as 0 by GetConstantValue (thus breaking the functionality that uses it). 

So, can you add some quick code to to `hack/db-schema-sync-check/main.go` in backend-shared, to verify that constants defined at the top of the `db_field_constants.go` are also included in this dbFieldMap?

You should just be able to do some simple string parsing, similar to what was done of part of the `db-schema-sync-check/main.go`. Then, to execute the check, run `run-db-schema-sync-check.sh` in `hack/`.

jgwest (changes requested):

jgwest (approved):
Looks great, thanks @jparsai!

PR #103, opened 2022-04-22
Write unit test cases for application controller in cluster-agent #116

Rizwana777:
Description:
- Unit test cases for application controller in cluster agent using ginkgo.

JIRA Story : https://issues.redhat.com/browse/GITOPSRVCE-116

### DIFF
--- a/cluster-agent/controllers/argoproj.io/application_controller_test.go
+++ b/cluster-agent/controllers/argoproj.io/application_controller_test.go
@@ -1,0 +1,539 @@
+package argoprojio_test
+
+import (
+	"context"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-appstudio/managed-gitops/backend/eventloop/eventlooptypes"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
+	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
+	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend/apis/managed-gitops/v1alpha1"
+	argo "github.com/redhat-appstudio/managed-gitops/cluster-agent/controllers/argoproj.io"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	name      = "my-application"
+	namespace = "argocd"
+	dbID      = "databaseID"
+)
+
+var _ = Describe("Application Controller", func() {
+	Context("Application Controller Test", func() {
+		var ctx context.Context
+		var k8sClient *sharedutil.ProxyClient
+		var managedEnvironment *db.ManagedEnvironment
+		var gitopsEngineInstance *db.GitopsEngineInstance
+		var dbQueries db.AllDatabaseQueries
+		var guestbookApp *appv1.Application
+		var reconciler argo.ApplicationReconciler
+		var err error
+
+		BeforeEach(func() {
+			scheme, argocdNamespace, kubesystemNamespace, workspace, _ := eventlooptypes.GenericTestSetup()
+			err = appv1.AddToScheme(scheme)
+			Expect(err).To(BeNil())
+
+			gitopsDepl := &managedgitopsv1alpha1.GitOpsDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-gitops-depl",
+					Namespace: workspace.Name,
+					UID:       uuid.NewUUID(),
+				},
+			}
+
+			informer := sharedutil.ListEventReceiver{}
+
+			// Fake kube client.
+			k8sClientOuter := fake.NewClientBuilder().WithScheme(scheme).WithObjects(gitopsDepl, workspace, argocdNamespace, kubesystemNamespace).Build()
+			k8sClient = &sharedutil.ProxyClient{
+				InnerClient: k8sClientOuter,
+				Informer:    &informer,
+			}
+
+			dbQueries, err = db.NewUnsafePostgresDBQueries(true, true)
+			Expect(err).To(BeNil())
+
+			err = testSetup()
+			Expect(err).To(BeNil())
+
+			_, managedEnvironment, _, gitopsEngineInstance, _, err = createSampleData(dbQueries)
+			Expect(err).To(BeNil())
+
+			guestbookApp = &appv1.Application{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Application",
+					APIVersion: "argoproj.io/v1alpha1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+					Labels: map[string]string{
+						dbID: "test-my-application",
+					},
+				},
+				Spec: appv1.ApplicationSpec{
+					Source: appv1.ApplicationSource{
+						Path:           "guestbook",
+						TargetRevision: "HEAD",
+						RepoURL:        "https://github.com/argoproj/argocd-example-apps.git",
+					},
+					Destination: appv1.ApplicationDestination{
+						Namespace: "guestbook",
+						Server:    "https://kubernetes.default.svc",
+					},
+					Project: "default",
+				},
+			}
+
+			reconciler = argo.ApplicationReconciler{
+				Client:        k8sClient,
+				Scheme:        scheme,
+				DB:            dbQueries,
+				TaskRetryLoop: sharedutil.NewTaskRetryLoop("application-reconciler"),
+			}
+		})
+
+		It("Create New Application CR in namespace and database and ApplicationState does not exist so it will create ApplicationState", func() {
+			ctx = context.Background()
+
+			// Add databaseID label to applicationID.
+			databaseID := guestbookApp.Labels[dbID]
+			applicationDB := &db.Application{
+				Application_id:          databaseID,
+				Name:                    name,
+				Spec_field:              "{}",
+				Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+				Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+			}
+
+			// Create a new ArgoCD Application.
+			err = reconciler.Create(ctx, guestbookApp)
+			Expect(err).To(BeNil())
+
+			// Create Application in Database.
+			err = reconciler.DB.CreateApplication(ctx, applicationDB)
+			Expect(err).To(BeNil())
+
+			// Call reconcile function.
+			result, err := reconciler.Reconcile(ctx, newRequest(namespace, name))
+			Expect(err).To(BeNil())
+			Expect(result).ToNot(BeNil())
+
+		})
+
+		It("ApplicationState already exists, so just update it", func() {
+			ctx = context.Background()
+
+			// Add databaseID label to applicationID.
+			databaseID := guestbookApp.Labels[dbID]
+			applicationDB := &db.Application{
+				Application_id:          databaseID,
+				Name:                    name,
+				Spec_field:              "{}",
+				Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+				Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+			}
+
+			applicationState := &db.ApplicationState{
+				Applicationstate_application_id: applicationDB.Application_id,
+				Health:                          "Healthy",
+				Sync_Status:                     "Synced",
+			}
+
+			// Create a new ArgoCD Application.
+			err = reconciler.Create(ctx, guestbookApp)
+			Expect(err).To(BeNil())
+
+			// Create Application in Database.
+			err = reconciler.DB.CreateApplication(ctx, applicationDB)
+			Expect(err).To(BeNil())
+
+			// Create ApplicationState in Database.
+			err = reconciler.DB.CreateApplicationState(ctx, applicationState)
+			Expect(err).To(BeNil())
+
+			// Call reconcile function.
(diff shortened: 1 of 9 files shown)


jparsai:
@Rizwana777 it would be good if you provide Jira ticket, details of changes done and purpose of this PR in PR description box.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,539 @@
...
+			informer := sharedutil.ListEventReceiver{}
+
+			// Fake kube client.
+			k8sClientOuter := fake.NewClientBuilder().WithScheme(scheme).WithObjects(gitopsDepl, workspace, argocdNamespace, kubesystemNamespace).Build()
+			k8sClient = &sharedutil.ProxyClient{
+				InnerClient: k8sClientOuter,
+				Informer:    &informer,
+			}
+
+			dbQueries, err = db.NewUnsafePostgresDBQueries(true, true)
```
jparsai, line 62: You need to close the DB connection at the end of test execution. like this `defer dbq.CloseDatabase()`

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,539 @@
...
+
+			gitopsDepl := &managedgitopsv1alpha1.GitOpsDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-gitops-depl",
+					Namespace: workspace.Name,
+					UID:       uuid.NewUUID(),
+				},
+			}
+
+			informer := sharedutil.ListEventReceiver{}
```
jparsai, line 53: I believe you don't need `ListEventReceiver` because you are not reading from Events.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,539 @@
...
+	clusterAccess := db.ClusterAccess{
+		Clusteraccess_user_id:                   testClusterUser.Clusteruser_id,
+		Clusteraccess_managed_environment_id:    managedEnvironment.Managedenvironment_id,
+		Clusteraccess_gitops_engine_instance_id: gitopsEngineInstance.Gitopsengineinstance_id,
+	}
+
+	return clusterCredentials, managedEnvironment, gitopsEngineCluster, gitopsEngineInstance, clusterAccess
+}
+
+func testSetup() error {
```
jparsai, line 342: I hope you would have confirmed through ./psql.sh or pgadmin, that all tables are being cleaned up.

cluster-agent/controllers/argoproj.io/application_controller_test.go
Rizwana777: Yes @jparsai I will update it.

cluster-agent/controllers/argoproj.io/application_controller_test.go
Rizwana777: Yes, It is not required, I will remove it @jparsai

Rizwana777:
> description

cluster-agent/controllers/argoproj.io/application_controller_test.go
Rizwana777: Yes I did @jparsai

Rizwana777:
@jparsai Thank you for the review, updated the changes, PTAL

jparsai:
LGTM

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+						Path:           "guestbook",
+						TargetRevision: "HEAD",
+						RepoURL:        "https://github.com/argoproj/argocd-example-apps.git",
+					},
+					Destination: appv1.ApplicationDestination{
+						Namespace: "guestbook",
+						Server:    "https://kubernetes.default.svc",
+					},
+					Project: "default",
+				},
```
jgwest, line 93: You may need to include the `status.health` and `status.sync` field of the `Application` resource, so that the controller know what sync/health to include in the ApplicationState row.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "my-gitops-depl",
+					Namespace: workspace.Name,
+					UID:       uuid.NewUUID(),
+				},
+			}
+
+			// Fake kube client.
+			k8sClientOuter := fake.NewClientBuilder().WithScheme(scheme).WithObjects(gitopsDepl, workspace, argocdNamespace, kubesystemNamespace).Build()
+			k8sClient = &sharedutil.ProxyClient{
```
jgwest, line 57: Looks like you can remove the `sharedutil.ProxyClient` here, because the test code doesn't use it's functionality (AFAICT).

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+
+			applicationState := &db.ApplicationState{
+				Applicationstate_application_id: applicationDB.Application_id,
+				Health:                          "Healthy",
+				Sync_Status:                     "Synced",
+			}
+
+			// Create a new ArgoCD Application.
+			err = reconciler.Create(ctx, guestbookApp)
+			Expect(err).To(BeNil())
```
jgwest, line 159: You should ensure that the `Application` resource created here has a `.status.health` and `.status.sync` values that are different from the `Healthy` and `Synced` that you have set in the `ApplicationState` row above, so that you can detect when the ApplicationState has been updated by `Reconcile`.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+
+			// Create ApplicationState in Database.
+			err = reconciler.DB.CreateApplicationState(ctx, applicationState)
+			Expect(err).To(BeNil())
+
+			// Call reconcile function.
+			result, err := reconciler.Reconcile(ctx, newRequest(namespace, name))
+			Expect(err).To(BeNil())
+			Expect(result).ToNot(BeNil())
+
```
jgwest, line 173: After `Reconcile` is called, you should verify that the `ApplicationState` row in the database is now equal to the `.status.health` and `.status.syncStatus` set in your `Application` resource.

Here we want to ensure the the Reconcile call will always update the `ApplicationState` row to be consistent with the Argo CD Application sync status and health, replacing the existing values in `ApplicationState` for these fields.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+			databaseID := guestbookApp.Labels[dbID]
+			applicationDB := &db.Application{
+				Application_id:          databaseID,
+				Name:                    name,
+				Spec_field:              "{}",
+				Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+				Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+			}
+
+			// Create Application in Database.
```
jgwest, line 257: It seems like we are just creating the Application in the database, and then immediately deleting it? We can probably just not create it to begin with.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+
+		})
+
+		It("Delete Application CR from namespace", func() {
+			// Close database connection.
+			defer dbQueries.CloseDatabase()
+
+			ctx = context.Background()
+
+			// Create a new ArgoCD Application.
```
jgwest, line 226: You can just call `Delete` in order to delete the `Application` resource; you don't need to create it first.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,556 @@
...
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
+	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
+	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend/apis/managed-gitops/v1alpha1"
+	argo "github.com/redhat-appstudio/managed-gitops/cluster-agent/controllers/argoproj.io"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	name      = "my-application"
```
jgwest, line 24: Can these const values be moved inside of `Context()`, below?

jgwest (changes requested):
Thanks Rizwana!

One thing I noticed is that I think a couple of the story criteria are still outstanding (correct me if I am wrong, e.g. if these are already covered by existing tests!) There should be tests that do the following:

#### Delete the Argo CD `Application` resource, and call Reconcile. Assert that no error is returned:
- Create an `Application` database row
- Create an Argo CD `Application` resource in the fake client, and make sure the `.metadata.labels` field has a label of `databaseID: (primary key of Application row)`
- Call Reconcile and make sure that an ApplicationState is created
- Delete the Argo CD Application resource from the fake k8s client
- Call Reconcile again, and make sure that no error is returned.


#### Call Reconcile on an Application that doesn't have a corresponding database entry; it should return an error
- Create an Argo CD Application in the fake client, and make sure the `.metadata.labels` field has a label of `databaseID: (a random uuid; this should be a value that doesn't exist in the database)`
- Make sure there is no Application database row, or ApplicationState row that has the primary key of `databaseID`.
- Call Reconcile.
- No error should be returned (I've updated the story criteria to reflect this new behaviour).

These correspond to the 3rd and 4th story criteria.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,692 @@
...
+				Spec_field:              "{}",
+				Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+				Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+			}
+
+			applicationState := &db.ApplicationState{
+				Applicationstate_application_id: applicationDB.Application_id,
+			}
+
+			// Create a new ArgoCD Application.
```
jgwest, line 129: Ginkgo actually has a neat way to include comments like these in the test output itself.

You can convert these explanatory comments in all your tests to: 
`By("...")`
for example, in this case, this comment would become
`By("Create a new Argo CD Application")`

Then this comment would be output during the test, which helps with test debugging.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,692 @@
...
+			err = reconciler.DB.GetApplicationStateById(ctx, applicationState)
+			Expect(err).To(BeNil())
+
+			// Verify that Health and Status of ArgoCD Application is equal to the Health and Status of Application in database.
+			Expect(applicationState.Health).To(Equal(string(guestbookApp.Status.Health.Status)))
+			Expect(applicationState.Sync_Status).To(Equal(string(guestbookApp.Status.Sync.Status)))
+
+		})
+
+		It("Application CR missing corresponding namespace entry", func() {
```
jgwest, line 202: Can change this to something like "calls Reconcile on an Argo CD Application resource that doesn't exist" (if my read of the test purpose is correct!)

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,692 @@
...
+			databaseID := guestbookApp.Labels[dbID]
+			applicationDB := &db.Application{
+				Application_id:          databaseID,
+				Name:                    name,
+				Spec_field:              "{}",
+				Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+				Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+			}
+
+			// Verify Application is present in database.
```
jgwest, line 282: Think this should be 'Verify Application is _not_ present in database."

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,692 @@
...
+			Expect(err).To(BeNil())
+
+			// Call reconcile function.
+			result, err := reconciler.Reconcile(ctx, newRequest(namespace, name))
+			Expect(err).To(BeNil())
+			Expect(result).ToNot(BeNil())
+
+		})
+
+		It("Application CR missing corresponding database entry", func() {
```
jgwest, line 230: Can likewise change this to something like "has an Argo CD Application that points to a database entry that doesn't exist"

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -0,0 +1,692 @@
...
+
+			// Create Application in Database.
+			err = reconciler.DB.CreateApplication(ctx, applicationDB)
+			Expect(err).To(BeNil())
+
+			// Call reconcile function.
+			result, err := reconciler.Reconcile(ctx, newRequest(namespace, name))
+			Expect(err).To(BeNil())
+			Expect(result).ToNot(BeNil())
+
```
jgwest, line 400: Did you want to check here that the health and sync status had been sanitized? It looks like that was the plan, based on the test description.

jgwest (changes requested):
Looks great, thanks @Rizwana777 ! Some minor tweaks then good to go.

cluster-agent/controllers/argoproj.io/application_controller_test.go
Rizwana777: Yes @jgwest I am checking here health and status had been sanitized

jgwest (approved):
Looks great, thanks @Rizwana777! :tada:

jgwest:
@Rizwana777  - Pushed some minor tweaks after rebasing on Samyak's merged PR:
- Move testSetup() into SetupForTestingDBGinkgo, so that other tests can use it (plus some tweaks, similar to the ones I made on Samyak's)
- Export DebugOnly_Shutdown so that tests can use it to shutdown the application state controller
- Ensure the application cache is started in cluster-agent main.go
- Ensure the application cache is started in application_controller_test.go

This should let other people use Ginkgo-based testSetup from this PR. Feel free to let me know if I broke anything with this commit :sweat_smile: .

Rizwana777:
Thank you @jgwest 
Everything is working fine :)

PR #104, opened 2022-04-25
Add 'URL' to db constants check

drpaneas:
This PR add the `URL` to be part of the check, as upper-case.
For example:

* db-schema.sql: `repository_url VARCHAR (512) NOT NULL,`
* db_field_constants.go: `RepositoryCredentialRepositoryURLLength`

This fails without this patch.


drpaneas (commented):
Thanks @jparsai :)

PR #107, opened 2022-05-02
PoC E2E test framework: fixtures for E2E tests plus a simple test

jgwest:
This PR implements an initial, proof-of-concept E2E test framework, plus a simple, sample test for it.

E2E tests are different from unit tests, in that they test against an actual K8s cluster (rather than a mock k8s client). Likewise, E2E tests run in an environment that is 'closer-to-production' than unit tests or a local development environment. 

With E2E tests (unlike with unit tests), the GitOps service is a black box: we invoke the GitOps Service public API (`GitOpsDeployment` k8s resources), and watch the API for expected changes.

#### PR Description:
- Switch default Argo CD namespace for gitops service from `argocd` to `gitops-service-argocd` (to match how it is deployed in staging/production)
- Update Panos' `argocd.sh` script:
    - @drpaneas: Requesting feedback on all these changes to your script, let me know if you love/hate them :sweat_smile: 
    - Change the listen port from 8080 to 18080, so that it doesn't conflict with pg-admin container (for folks using local postgresql dev env)
    - Allow the target Argo CD namespace to be customized via ARGO_CD_NAMESPACE environment variable.
    - Allow the target Argo CD version to be customized via ARGO_CD_VERSION environment variable.
    - Change default Argo CD namespace from `argocd` to `gitops-service-argocd` (same reason as above)
- Add new scripts which start Argo CD, either via kubectl or OpenShift GitOps, for use by E2E test scripts. (For other uses, Panos' script is preferred, since it has better UX, and provides port forwarding)
- Update Makefile to support running E2E tests, either standalone or as part of a CI task (CI task coming soon)
- Implement Ginkgo-based test fixtures for E2E tests:  
    - Test fixtures which create/wait for K8s resources, GitOps deployment, and handle test cleanup.
    - @drpaneas: feedback welcome if you happen to notice I'm doing something wrong (un-idiomatic) here, for Ginkgo/Gomega
- Implement a new test using this fixtures

#### Information for reviewers

Reviewers: feedback welcome! This is my first shot at putting together some E2E test fixtures for the GitOps Service, plus a first test that uses that.

You should be able to run the E2E test by doing this:
1) Get a clusterbot cluster and log in to it
2) Run the following commands:
```
git clone https://github.com/jgwest/managed-gitops
cd managed-gitops
git checkout e2e-tests

# Install Argo CD into gitops-service-argocd, using OpenShift GitOps operator
make setup-e2e-openshift

# Start GitOps Service (in E2E configuration)
make start-e2e

# In a separate terminal window, to run the tests:
make test-e2e
```


backend-shared/config/db/db_field_constants.go
```diff
@@ -142,6 +142,7 @@ var DbFieldMap = map[string]int{
 	"DeploymentToApplicationMappingNamespaceLength":                           DeploymentToApplicationMappingNamespaceLength,
 	"DeploymentToApplicationMappingDeploymentNamespaceLength":                 DeploymentToApplicationMappingNamespaceLength,
 	"DeploymentToApplicationMappingWorkspaceUIDLength":                        DeploymentToApplicationMappingWorkspaceUIDLength,
+	"DeploymentToApplicationMappingNamespaceUIDLength":                        DeploymentToApplicationMappingWorkspaceUIDLength,
```
samyak-jn, line 145: Is this not duplicated with other changes in place? The fix for this is already in #108

tests-e2e/Makefile
```diff
@@ -0,0 +1,79 @@
...
+# This is a requirement for 'setup-envtest.sh' in the test target.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
+##@ General
+
+# The help target prints out all targets with their descriptions organized
+# beneath their categories. The categories are represented by '##@' and the
+# target descriptions by '##'. The awk commands is responsible for reading the
```
samyak-jn, line 21:
```suggestion
# target descriptions by '##'. The awk commands are responsible for reading the
```

samyak-jn (commented):
Hi @jgwest, this initial implementation is very neatly written, thank you for this! :rocket: 
The first thing, that I wanted to bring up as someone who is seeing it for the first time is, can we add your PR description as a draft README for this (and open a story to refine it afterward)?
Lastly, I suggested two very low cosmetic suggestions feel free to ignore then :laughing:

Makefile
```diff
@@ -117,14 +118,24 @@ test-appstudio-controller: ## Run test for appstudio-controller only
...
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ #
 
+install-argocd-openshift: ## Using OpenShift GitOps, install Argo CD to the gitops-service-argocd namespace
+	manifests/openshift-argo-deploy/deploy.sh
+
+install-argocd-k8s: ## (Non-OpenShift): Install Argo CD to the gitops-service-argocd namespace
+	ARGO_CD_VERSION=$(ARGO_CD_VERSION) manifests/k8s-argo-deploy/deploy.sh
+
+uninstall-argocd: ## Uninstall Argo CD from gitops-service-argocd namespace (from either OpenShift or K8s)
+	kubectl delete namespace gitops-service-argocd || true
```
wtam2018, line 136: Should use the `ARGO_CD_NAMESPACE` env var?

tests-e2e/fixture/k8s/fixture.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+
+	fmt.Println("Creating ", obj.GetName())
+
+	if err := client.Create(context.Background(), obj); err != nil {
+		fmt.Println("Error on creating ", obj.GetName(), err)
+		return err
+	}
+
```
drpaneas, line 30: I would add a validation, by grabbing the newly created object to make sure this is indeed created. We are doing similar functionality for the `.Delete` as well.

tests-e2e/fixture/fixture.go
```diff
@@ -0,0 +1,196 @@
...
+		return false, nil
+	}); err != nil {
+		return err
+	}
+
+	// Create the namespace again
+	_, err = kubeClientSet.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
+		Name: GitOpsServiceE2ENamespace,
+		Labels: map[string]string{
+			"argocd.argoproj.io/managed-by": "gitops-service-argocd",
```
drpaneas, line 67: can you maybe use the  already existing`const DefaultGitOpsEngineSingleInstanceNamespace = "gitops-service-argocd"` ?

tests-e2e/fixture/fixture.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		count := 0
+		// Exit the poll loop when there exists at least one rolebinding containing 'gitops-service-argocd' in it's name
+		// This helps us avoid a race condition where the namespace is created, but Argo CD has not yet
+		// set up proper roles for it.
+		for _, item := range roleBindings.Items {
+			if strings.Contains(item.Name, "gitops-service-argocd-") {
```
drpaneas, line 90: same here, try to use the _const_ `DefaultGitOpsEngineSingleInstanceNamespace`

tests-e2e/fixture/fixture.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		count := 0
+		// Exit the poll loop when there exists at least one rolebinding containing 'gitops-service-argocd' in it's name
+		// This helps us avoid a race condition where the namespace is created, but Argo CD has not yet
+		// set up proper roles for it.
+		for _, item := range roles.Items {
+			if strings.Contains(item.Name, "gitops-service-argocd-") {
```
drpaneas, line 117: Same here as well with the const.

drpaneas (changes requested):
Excellent work @jgwest :ok_hand: many thanks for taking the brave step of introducing the rabbit hole of e2e tests :star2: :smile_cat:

Some comments from my side:

1. Update the `README.md` of the root repo to mention about the new tests-e2e directory and its purpose and instructions on how to _run_ them.
2. I've made some suggestions to the bash script but looks fine to me. They are just improvements of the same functionality. I haven't tested it _yet_ but it's in my todo list.
3. And some general tips/suggestions on the e2e simple test:

* It would be great if you can write a comment for all of the fixtured (exported functions).  Just a simple one-liner description in the comment area to provide a brief understanding of what they are doing and how they work. Some have it, some don't.
* When it comes to logging, I am fan of consistency, so I would prefer to have identical logging mechanism for the fixtures. This will help debugging faster and easier and it will allow allow Prometheus to consume our tests and provide some metrics (e.g. which errors fail more often, etc). A basic logging skeleton could be to: 1) announce what the functions are doing in the beginning, and 2) inform us of the outcome (**failed** or **passed**) of the test. Yet again, this is very opinionated so I am not forcing you or anyone to follow this pattern.
* Continuing the same idea, it would be nice to create an`errFunctioname` error for each one of them, where they are returning it in case of failure. Inside of it, the the `errFunctioname` can wrap the **actual** error coming from the system itself but the overall error could be parsable (e.g. `k8s.ExistByName() Error: Object does not exists in ExistByName: foobar`.) 
* Also, every test has some kind of 3 sections: "setup", "run" and/or "clean". Personally I like to have both a visual and a err-log-friendly representation of these sections by using comments in the fixtures. For example:

```go
// --------------------- //
//.         SETUP.       //
// --------------------- //
errExistByNameSetup := errors.New("....")
k8sClient, err := fixture.GetKubeClient()
if err != nil {
  // print the errExistByNameSetup by wrapping the err as well
  return false
}

// --------------------- //
//.         TEST.        //
// --------------------- //
errExistByNameTest := errors.New("....")
err = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(k8sObject), k8sObject)
if err != nil {
   // same principle here as well
}
```

Once again, please take these suggestions with a grain of salt.  I do understand this is very much opinionated so I m not forcing anybody to follow this pattern of course. Take this as a suggestion from someone who has been QA-ing for many years and had to debug hundreds of failed tests.

FYI: last but not least we have also to think if we need to duplicate some these functions introducing deadlines, retries, etc ... which is a major pain. But that's another story ...

Congrats again for the PR, looking forward to start e2e testing :D

Makefile
drpaneas: I think so as well ^

tests-e2e/fixture/fixture.go
```diff
@@ -0,0 +1,196 @@
...
+
+}
+
+func GetKubeConfig() (*rest.Config, error) {
+
+	overrides := clientcmd.ConfigOverrides{}
+
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	clientConfig := clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &overrides, os.Stdin)
+
```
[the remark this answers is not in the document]
jgwest: I tried to add this one, but the logging was way too verbose: it output a 'getting the kubeconfig' message every second (because the existing fixtures are retrieving the client at the start of every attempt).

tests-e2e/fixture/fixture.go
```diff
@@ -0,0 +1,196 @@
...
+	restConfig, err := clientConfig.ClientConfig()
+
+	if restConfig != nil {
+		// Sanity check that we're not running on a known staging system
+		if strings.Contains(restConfig.Host, "x99m.p1.openshiftapps.com") {
+			return nil, fmt.Errorf("E2E tests should not be run on staging server")
+		}
+	}
+
```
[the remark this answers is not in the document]
jgwest: > define an errGetKubeConfig and return this in case of an error, wrapping inside of it the errors coming from the system itself.

Not sure I see what the request is here, can you provide an example of what you mean?

> In case there is no error, print a statement it saying that kubeconfig is configured properly against X target cluster.

Tried this but too verbose, see above for result.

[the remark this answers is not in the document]
drpaneas: > Not sure I see what the request is here, can you provide an example of what you mean?

```go
// errors per operation
var errClient = errors.New("couldn't get k8s client")
var errCreate = errors.New("couldn't create k8s resource")
var errDelete = errors.New("couldn't delete k8s object from the namespace")
var errYouMentionIt = errors.New("think of another error")

// Create creates the given K8s resource, returning an error on failure, or nil otherwise.
func Create(obj client.Object) error {
	// ~~~ setup ~~~ //
	kubeClient, err := fixture.GetKubeClient()
	if err != nil {
		return fmt.Errorf("%v (error: %w)", errClient, err)
	}

	// ~~~ test ~~~ //
	fmt.Println("Creating ", obj.GetName())

	if err := kubeClient.Create(context.Background(), obj); err != nil {
		return fmt.Errorf("%v (error: %w)", errCreate, err)
	}

	return nil
}
```

> 
> > In case there is no error, print a statement it saying that kubeconfig is configured properly against X target cluster.
> 
> Tried this but too verbose, see above for result.

Yes it's too verbose but for a food reason. As I said yesterday in the parking lot, in the end it's all about finding the right balance. If we have just a few tests written by us and maintained by us -- we are probably good to go without that level of verbosity.  But if we are going to be responsible to debug a few hundreds of tests written by a whole organization, then verbosity would be greatly appreciated. 

Another idea might be to consider introducing debug levels where you can have all the verbosity you need + run a few interesting kubectl commands (in case of failed testsuite) on the cluster itself (`oc get nodes`, `oc top`, etc), and finally export the output into multiple files and generate a compressed logs.tar.gz file for us. 

But anyway, we can discuss all those things later, not in this PR.

tests-e2e/fixture/k8s/fixture.go
```diff
@@ -0,0 +1,84 @@
...
+	. "github.com/onsi/gomega"
+
+	matcher "github.com/onsi/gomega/types"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// Create creates the given K8s resource, returning an error on failure, or nil otherwise.
+func Create(obj client.Object) error {
+
+	client, err := fixture.GetKubeClient()
```
drpaneas, line 19: the variable `client` collides with imported package name `client "sigs.k8s.io/controller-runtime/pkg/client"`

```suggestion
	kubeClient, err := fixture.GetKubeClient()
```

tests-e2e/fixture/fixture.go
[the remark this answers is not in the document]
jgwest: @drpaneas: ah, that is a lot clearer for me, thanks!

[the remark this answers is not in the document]
jgwest: @drpaneas: And +1 for different debug log levels for our components (for example, the ability to enable debug for a particular file/component only), +1 for provided a more consistent log format (our current log format is very anarchic :smile: ).

samyak-jn:
This is neatly done and awesome work @jgwest 
I'm book marking this PR for my learning purpose 🎉

PR #108, opened 2022-05-03
Fix: Reset getErr variable to be used in next step

Michkov:
getErr with ignored error was passed into next part where context load
failed on this ignored error.

#### Description:
- Fixing issue when branch is not defined, that error is ignored but in context check the error is still set.


mmorhun (approved):

jgwest (approved):
Thanks @Michkov!

PR #110, opened 2022-05-04
Fix 'odd number of arguments passed as key-value pairs for logging' in appstudio-controller

jgwest:
#### Description:
- Fix improper use of log API: log.Info doesn't support Sprintf-style arguments.


jgwest:
Merging this without reviewers, so I can quickly unblock an issue folks are hitting with our service on staging cluster, but feedback still welcome if there is anything controversial :smile: .

PR #111, opened 2022-05-05
Update memory/cpu req/limits for components in K8s resource manifests

jgwest:
#### Description:
- Increase the resource limits on the controllers, when running on appstudio staging cluster
 

jgwest:
Merging without reviewers, to fix ongoing issue with appstudio staging cluster.

PR #116, opened 2022-05-10
Fix unreported panic on login credentials test

jgwest:
#### Description:
- The `argocd_login_credentials_test.go` was reporting a mock error in the product code which wasn't being caught by the unit test itself.
- This PR updates the product code to detects panics (which are what the mock returns on unexpected value), which will invalidate the cache, and directly return an error
- Add the missing version client call to the mock


drpaneas (approved):
lgtm

samyak-jn (approved):
LGTM as well, thanks for the description for the fix @jgwest!

PR #117, opened 2022-05-11
Add unit test for GetOrCreateManagedEnvironmentByNamespaceUID function.

jparsai:


backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+			clusterCredentials := db.ClusterCredentials{
+				Clustercredentials_cred_id: managedEnvironment.Clustercredentials_id,
+			}
+			err = dbQueries.GetClusterCredentialsById(ctx, &clusterCredentials)
+			Expect(err).To(BeNil())
+
+			// ----------------------------------------------------------------------------
+			By("If called 2nd time, it should return existing ManagedEnvironment instead of creating new and flag should be False.")
+			// ----------------------------------------------------------------------------
+			managedEnvironmentSecond, isNewSecond, errSecond := GetOrCreateManagedEnvironmentByNamespaceUID(ctx, workspace, dbQueries, log)
```
samyak-jn, line 75: I think `isNew` as a prev variable name will suffice for this as well, do we have a use-case of using `isNewSecond` separately?

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
 	"github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
 	"github.com/stretchr/testify/assert"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	logger "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var _ = Describe("Test utility functions.", func() {
+	Context("It should execute and test GetOrCreateManagedEnvironmentByNamespaceUID function.", func() {
```
samyak-jn, line 22:
```suggestion
	Context("Testing for GetOrCreateManagedEnvironmentByNamespaceUID function.", func() {
```

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+			log := log.FromContext(context.Background())
+
+			managedEnvironment, isNew, err := GetOrCreateManagedEnvironmentByNamespaceUID(ctx, workspace, dbQueries, log)
+			Expect(err).NotTo(BeNil())
+			Expect(isNew).To(BeFalse())
+			Expect(managedEnvironment).To(BeNil())
+		})
+	})
+
+	Context("It should execute and test GetOrCreateDeploymentToApplicationMapping function.", func() {
```
samyak-jn, line 125:
```suggestion
	Context("Testing for GetOrCreateDeploymentToApplicationMapping function.", func() {
```

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+			// ---------------------------------------------------------------------------
+			// Create deploymentToApplicationMapping request data
+
+			deploymentToApplicationMapping = &db.DeploymentToApplicationMapping{
+				Deploymenttoapplicationmapping_uid_id: string(uuid.NewUUID()),
+				Application_id:                        application.Application_id,
+				DeploymentName:                        "my-depl-to-app-mapping",
+				DeploymentNamespace:                   "my-namespace",
+				NamespaceUID:                          "test-1",
+			}
```
samyak-jn, line 211: For the above function, can't we use CreateSampleData functionality for the required components, rest can be created as above?

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+		})
+
+		It("Should create new DeploymentToApplicationMapping and if called second time, it should return existing resource instead of creating new.", func() {
+
+			// ----------------------------------------------------------------------------
+			By("Create new DeploymentToApplicationMapping resource.")
+			// ----------------------------------------------------------------------------
+			isNewFirst, errFirst := GetOrCreateDeploymentToApplicationMapping(ctx, deploymentToApplicationMapping, dbQueries, log)
+			Expect(errFirst).To(BeNil())
+			Expect(isNewFirst).To(BeTrue())
```
samyak-jn, line 248:
```suggestion
			isNew, err := GetOrCreateDeploymentToApplicationMapping(ctx, deploymentToApplicationMapping, dbQueries, log)
			Expect(err).To(BeNil())
			Expect(isNew).To(BeTrue())
```

samyak-jn: Feel free to ignore these, as it is just added to remove confusion of too many variables!

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+			})
+			Expect(err).To(BeNil())
+
+			// ----------------------------------------------------------------------------
+			By("Try to create same DeploymentToApplicationMapping, it should return existing resource.")
+			// ----------------------------------------------------------------------------
+
+			isNewSecond, errSecond := GetOrCreateDeploymentToApplicationMapping(ctx, deploymentToApplicationMapping, dbQueries, log)
+			Expect(errSecond).To(BeNil())
+			Expect(isNewSecond).To(BeFalse())
```
samyak-jn, line 265: Similarly here (as mentioned in above comment)^^

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+		})
+
+		It("Should delete exiting DeploymentToApplicationMapping if a resource is passed having same name/namespace, so there is be only one resource per name and namespace.", func() {
```
samyak-jn, line 268:
```suggestion
		It("Should delete existing DeploymentToApplicationMapping if a resource is passed having same name/namespace, so there is be only one resource per name and namespace.", func() {
```

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+
+			// ----------------------------------------------------------------------------
+			By("Create first DeploymentToApplicationMapping resource.")
+			// ----------------------------------------------------------------------------
+
+			isNewFirst, errFirst := GetOrCreateDeploymentToApplicationMapping(ctx, deploymentToApplicationMapping, dbQueries, log)
+			Expect(errFirst).To(BeNil())
+			Expect(isNewFirst).To(BeTrue())
+
```
samyak-jn, line 277: Similarly here (as mentioned in the above comments)^^

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,321 @@
...
+				Application_id:                        application.Application_id,
+				DeploymentName:                        "my-depl-to-app-mapping",
+				DeploymentNamespace:                   "my-namespace",
+				NamespaceUID:                          "test-1",
+			}
+
+			isNewSecond, errSecond := GetOrCreateDeploymentToApplicationMapping(ctx, deploymentToApplicationMappingSecond, dbQueries, log)
+			Expect(errSecond).To(BeNil())
+			Expect(isNewSecond).To(BeTrue())
+
```
samyak-jn, line 293: Similarly here (as mentioned in the above comments)^^

samyak-jn (commented):
Hi @jparsai, LGTM, thank you for working on this! Just added some minor suggestions here.

backend-shared/config/db/util/utils_test.go
jparsai: Yes, it could be done, my thinking here was that it Is is not repeated anywhere, just used at one place. If needed I am Okay to move it into a function. WDYT?

backend-shared/config/db/util/utils_test.go
jparsai: Yes, it could be done, my thinking here was that it Is is not repeated anywhere, just used at one place. If needed I am Okay to move it into a function. WDYT?

samyak-jn: Hmm, so what I meant was using CreateSampleData for creating *ClusterCredentials, *ManagedEnvironment, *GitopsEngineCluster, and *GitopsEngineInstance from the existing function, the rest of the code can be kept as similar to what you have written? Is that what you were saying or did I miss something?

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+			// Create GitopsEngineInstance
+
+			gitopsEngineInstance = db.GitopsEngineInstance{
+				Namespace_name:   "my-namespace",
+				Namespace_uid:    "test-1",
+				EngineCluster_id: gitopsEngineCluster.Gitopsenginecluster_id,
+			}
+			err = dbQueries.CreateGitopsEngineInstance(ctx, &gitopsEngineInstance)
+			Expect(err).To(BeNil())
+
```
samyak-jn, line 279:
```suggestion
clusterCredentials, managedEnvironment, gitopsEngineCluster, gitopsEngineInstance, _, err := db.CreateSampleData(dbq)
Expect(err).To(BeNil())

```
Rest we can leave as it is

jparsai: Yes I tried that too, but I get error but I get error `primary key should be empty` from the function, so I didnt want to change anything in function so I am created required resources, it also allowed me to give resource details (namespace, name etc) as I need.
Could you please try it once, to confirm that I am not doing something wrong.

samyak-jn: Sorry just saw your reply on this, will give this a try to reciprocate why it is happening. Github should add system notifications when the tabs are open :(

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+			Expect(err).To(BeNil())
+			Expect(isNew).To(BeTrue())
+
+			// ----------------------------------------------------------------------------
+			By("Verify that other database entries are also created.")
+			// ----------------------------------------------------------------------------
+
+			// Check KubernetesToDBResourceMapping resource
+			kubernetesToDBResourceMapping := db.KubernetesToDBResourceMapping{
+				KubernetesResourceType: "Namespace",
```
jgwest, line 160: You can use `db.K8sToDBMapping_Namespace` for this

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+				KubernetesResourceUID:  string(workSpaceUid),
+				DBRelationType:         "ManagedEnvironment",
```
jgwest, line 162: You can use `db.K8sToDBMapping_ManagedEnvironment` for this

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+			Expect(isNew).To(BeTrue())
+
+			// ----------------------------------------------------------------------------
+			By("Check first DeploymentToApplicationMapping is deleted in DB.")
+			// ----------------------------------------------------------------------------
+
+			err = dbQueries.GetDeploymentToApplicationMappingByDeplId(ctx, &db.DeploymentToApplicationMapping{
+				Deploymenttoapplicationmapping_uid_id: deploymentToApplicationMapping.Deploymenttoapplicationmapping_uid_id,
+			})
+			Expect(err).NotTo(BeNil())
```
jgwest, line 383: I suggest using `IsResultNotFoundError` on the error, to make sure we are getting the correct error result here.

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+			By("To cleanup the resource created by test")
+			// ----------------------------------------------------------------------------
+
+			// Fetch KubernetesToDBResourceMapping to be cleaned in AfterEach.
+			kubernetesToDBResourceMapping, _, err = findKubernetesToDBResourceMappingInTable(ctx, dbQueries, gitopsEngineCluster.Gitopsenginecluster_id, db.K8sToDBMapping_GitopsEngineCluster)
+			Expect(err).To(BeNil())
+		})
+	})
+
+	Context("Testing for GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID function.", func() {
```
jgwest, line 668: It can be tough to tell whether the story criteria items are being met when it comes to writing tests, due to the nature of test writing.

In order to tell if a particular part of a test is being covered, you can verify this by running `make test`, which will produce a coverage.out file. Then, in VS Code, select `View` > `Command Palette`, then `Go: Apply Cover Profile`, and specify the absolute path of the coverage.out file.

I did that, and it appears that the following story criteria items are not being fully tested:
- To test: if a KubernetesToDBResourceMapping mapping exists, and the engine instance row doesn't exist: verify that the function deletes the old mapping, and verify it creates a new KubernetesToDBResourceMapping and GitOpsEngineInstance row
- To test: if a KubernetesToDBResourceMapping mapping exists, and the engine instance row exists: verify that the function does not create a new KubernetesToDBResourceMapping or a new GitOpsEngineInstanceRow... it; it should just retrieve existing values from database

Then, If you follow the above steps, then you open `GetOrCreateGitopsEngineInstanceByInstanceNamespaceUID` function, you will see that 'Scenario C' and 'Scenario D' code blocks are not covered. This corresponds to the two test items above. (Don't worry about 'Scenario B'.)

(https://issues.redhat.com/browse/GITOPSRVCE-135)

backend-shared/config/db/util/utils_test.go
```diff
@@ -1,12 +1,881 @@
...
+			})
+			Expect(err).To(BeNil())
+
+			// send second resource object to get deleted in AfterEach
+			deploymentToApplicationMapping = deploymentToApplicationMappingSecond
+		})
+	})
+
+	Context("Testing for GetOrCreateGitopsEngineClusterByKubeSystemNamespaceUID function.", func() {
+
```
jgwest, line 400: It can be tough to tell whether the story criteria items are being met when it comes to writing tests, due to the nature of test writing.

In order to tell if a particular part of a test is being covered, you can verify this by running `make test`, which will produce a coverage.out file. Then, in VS Code, select `View` > `Command Palette`, then `Go: Apply Cover Profile`, and specify the absolute path of the coverage.out file.

In this case, after doing that, it appears that the following story criteria items are not being fully tested:
- if a KubernetesToDBResourceMapping mapping exists, and the GitOpsEngineCluster row doesn't exist: The GetOrCreateGitopsEngineClusterByKubeSystemNamespaceUID function should delete the old mapping, and create a new mapping/cluster
- if a KubernetesToDBResourceMapping mapping exists, and the GitOpsEngineCluster row exists: The  GetOrCreateGitopsEngineClusterByKubeSystemNamespaceUID should not create, and should just retrieve existing values from the database

You you follow the above steps, then open `GetOrCreateGitopsEngineClusterByKubeSystemNamespaceUID` function, you will see that 'Scenario C' and 'Scenario D' code blocks are not covered. This corresponds to the two test items above. (Don't worry about 'Scenario B'.) You can verify using `Go: Apply Cover Profile`, as above.

(https://issues.redhat.com/browse/GITOPSRVCE-134)

jgwest (changes requested):

backend-shared/config/db/util/utils_test.go
jparsai: Actually these story criteria items are covering these tests. 
`To test: if a KubernetesToDBResourceMapping mapping exists, and the engine instance row doesn't exist: verify that the function deletes the old mapping, and verify it creates a new KubernetesToDBResourceMapping and GitOpsEngineInstance row`
**It("Should delete existing KubernetesToDBResourceMapping and create new KubernetesToDBResourceMapping and GitopsEngineInstance.", func(){}**

`To test: if a KubernetesToDBResourceMapping mapping exists, and the engine instance row exists: verify that the function does not create a new KubernetesToDBResourceMapping or a new GitOpsEngineInstanceRow... it; it should just retrieve existing values from database`
**It("Should return existing KubernetesToDBResourceMapping and GitopsEngineInstance, instead of creating new.", func(){}**

In test coverage report `Scenario C` is not covered, but what I understand, in current implementation this code block will not execute. 
This block will execute if instance exists, but there is no mapping for it. In that case condition written above will execute and control will not go upto block written for Section C

```
// <This condition will satisfy because mapping is not available, 
// hence it will not go in else block and instance will not be fetched from DB, 
// it means instance will be Nil>

if err := dbq.GetDBResourceMappingForKubernetesResource(ctx, dbResourceMapping); err != nil {
	dbResourceMapping = nil
} else {
	// This will search for GitOpsEngineCluster and GitOpsEngineInstance
}

// <Mapping and instance both are Nil so Scenario A will execute instead of Scenario C>

if dbResourceMapping == nil && gitopsEngineInstance == nil {
		// Scenario A) neither exists: create both
}  else if dbResourceMapping == nil && gitopsEngineInstance != nil {
		// Scenario C) this will happen if the instance exists, but there is no mapping for it
}
```

backend-shared/config/db/util/utils_test.go
jparsai: Yes, I missed here.

PR #118, opened 2022-05-13
Add: DB migration versioning.

tirthyme:
#### Description:
- Adds DB versioning to prod db.

[GITOPSRVCE-115 Investigate: database version migration](https://issues.redhat.com/browse/GITOPSRVCE-115)


.github/workflows/static.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - "main"
+      - "db-migration"
```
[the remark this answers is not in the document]
tirthyme: yes, for testing that is needed to enable workflows whenever I change something to static.yml and push it into my private repo. will remove it once the PR is approved.

.github/workflows/static.yml
```diff
@@ -130,6 +131,49 @@ jobs:
...
+          cd $GITHUB_WORKSPACE/migration
+          go run migrate_db.go drop_smtable
+      - name: "Create a pg_dump file for migration schema"
+        run: |
+          cd $GITHUB_WORKSPACE
+          pg_dump postgresql://postgres:gitops@localhost:5432/postgres > migration.sql && cat migration.sql
+      - name: "Execute diff to check for any discrepencies"
+        run: |
+          cd $GITHUB_WORKSPACE
+          diff migration.sql master.sql
```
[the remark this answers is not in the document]
tirthyme: When files are not equal, the diff will fail which will lead to the pipeline failing as well alongside with showing what is changed. It is intended behavior.

.github/workflows/static.yml
```diff
@@ -130,6 +131,49 @@ jobs:
...
+          cd $GITHUB_WORKSPACE/migration
+          go mod download
+      - name: "Start PostgreSQL"
+        run: |
+          cd $GITHUB_WORKSPACE
+          ./create-dev-env.sh
+      - name: "Create a pg_dump file for master schema"
+        run: |
+          cd $GITHUB_WORKSPACE
+          pg_dump postgresql://postgres:gitops@localhost:5432/postgres > master.sql && cat master.sql
```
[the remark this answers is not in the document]
tirthyme: noted.

migration/readme.md
```diff
@@ -0,0 +1,17 @@
...
+- https://github.com/golang-migrate/migrate/blob/master/FAQ.md#what-does-dirty-database-mean
+
+### How are migrations applied? 
+
+- Whenever you need to make changes to the current applied schema, in the migrations folder, run `migrate create -ext sql -seq {{file_name}}`
+- A new version of a migration will be created. Once the changes are written and verified.
+- Make sure to write a rollback script in `__.down.sql` file. Otherwise the rollback of migrations won't be possible.
+- To apply the latest version of the migration, simply execute `go run migrate_db.go`
+- For additional utilities, for eg: drop the entire db, simply pass drop as a runtime argument like `go run migrate_db.go drop`
+- **DO NOT** drop `schema_migrations` table as that will lead to migration failure.
```
samyak-jn, line 17: I would also add on a suggestion to play more with markdowns, it will be helpful to beautify and make your writeups readable easily by the devs or end-user. Also, avoid using terms like `you/I` in the README(s). I have made some quick first mind suggestions that came to my mind but feel free to explore more. 
Try to add about the nomenclature of the migrateDB scaling files, also can add a point that these are autogenerated.

samyak-jn (commented):
Hi @tirthyme, thanks for working on these, looks good. I have added some quick suggestions for refinement of the migrate repo.

migration/readme.md
tirthyme: Thanks for the suggestion, samyak. Will keep these in mind and hopefully can make better documents moving ahead!

migration/migrate_db.go
```diff
@@ -0,0 +1,93 @@
...
+	db := pg.Connect(opts)
+
+	if err := checkConn(db); err != nil {
+		return nil, fmt.Errorf("%v, unable to connect to database: Host:'%s' User:'%s' Pass:'%s' DB:'%s' ", err, opts.Addr, opts.User, opts.Password, opts.Database)
+	}
+
+	return db, nil
+}
+
+func main() {
```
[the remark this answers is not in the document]
tirthyme: yeah, so point of having package main is for it to run separately as it is basically meant to be used as a cli tool instead of trying to import it anywhere as it updates the schema of the db.

migration/migrate_db.go
```diff
@@ -0,0 +1,93 @@
+package main
```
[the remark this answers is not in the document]
drpaneas: to my understanding talking to @tirthyme, the intention here is to have this as a standalone binary/executable. This code is not meant to be exported and consumed by other packages. The use case is to do something like this from the Makefile:

```shell
cd hacks/migration; go run migrate_db.go
```

To do that, the package has to be `main`. It's a common pattern to create isolated files for your code:

```shell
/ (root)
 -> examples (dir)
      -> ex1 (dir)
            ex1.go (main pkg)
      -> ex2 (dir)
             ex2.go (main pkg)
      -> ex3 (dir)
             ex3.go (main pkg)
```

Ideally, we can introduce a `cmd` package for the CLI stuff. In that case, yes it could name the package `db` because it would have been imported by the `cmd` one. Right now we have just the make file where is the common denominator for calling all of these _hacks_.

migration/migrate_db.go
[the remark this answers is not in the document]
drpaneas: to my understanding talking to @tirthyme, the intention here is to have this as a standalone binary/executable. This code is not meant to be exported and consumed by other packages. The use case is to do something like this from the Makefile:

```shell
cd hacks/migration; go run migrate_db.go
```

To do that, the package has to be `main`. It's a common pattern to create isolated files for your code:

```shell
/ (root)
 -> examples (dir)
      -> ex1 (dir)
            ex1.go (main pkg)
      -> ex2 (dir)
             ex2.go (main pkg)
      -> ex3 (dir)
             ex3.go (main pkg)
```

Ideally, we can introduce a `cmd` package for the CLI stuff. In that case, yes it could name the package `db` because it would have been imported by the `cmd` one. Right now we have just the make file where is the common denominator for calling all of these _hacks_.

migration/.gitignore
```diff
@@ -0,0 +1,6 @@
+bin/
+dist/
+coverage.out
+vendor/
+cover.out
+.idea/
```
wtam2018, line 6:
```suggestion
.idea/
.vscode/
```

migration/migrate_db.go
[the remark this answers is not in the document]
wtam2018: These are quite generic db util functions which probably will get reused as some point. I think we should do either the hack pattern or make it like a CLI as @drpaneas described. Either way is not hard. :-)

migration/migrations/000001_init_db.down.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN;
+
+
+DROP TABLE IF EXISTS ApplicationState;
+DROP TABLE IF EXISTS DeploymentToApplicationMapping;
+DROP TABLE IF EXISTS KubernetesToDBResourceMapping ;
```
wtam2018, line 6:
```suggestion
DROP TABLE IF EXISTS KubernetesToDBResourceMapping;
```

migration/migrations/000001_init_db.down.sql
```diff
@@ -0,0 +1,21 @@
...
+DROP TABLE IF EXISTS APICRToDatabaseMapping ;
```
wtam2018, line 7:
```suggestion
DROP TABLE IF EXISTS APICRToDatabaseMapping;
```

migration/readme.md
```diff
@@ -0,0 +1,27 @@
+
+# Managed GitOps Cluster-Agent
+
+[![GoDoc](https://godoc.org/github.com/redhat-appstudio/managed-gitops/migration?status.svg)](https://pkg.go.dev/mod/github.com/redhat-appstudio/managed-gitops/migration)
+[![License](https://img.shields.io/:license-apache-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0.html)
+
+----
+
+The Managed GitOps migration component is responsible for making sure the database can be versioned UP and DOWN according to the need.
```
wtam2018, line 9:
```suggestion
The Managed GitOps Database Migration component is responsible for making sure that the database can be versioned UP and DOWN according to the need.
```

migration/readme.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Getting started with migrations for manged-gitops
+
```
wtam2018, line 12:
```suggestion
## Getting started with database migrations for manged-gitops

```

migration/readme.md
```diff
@@ -0,0 +1,27 @@
...
+### For the Installation Guide refer to the [migration guide](https://github.com/golang-migrate/migrate/blob/master/cmd/migrate/README.md).
+
+### For best practices refer to the [migrate pkg documentations](https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md).
+
+### Facing a dirty migration issue? Refer: [migrate FAQs](https://github.com/golang-migrate/migrate/blob/master/FAQ.md#what-does-dirty-database-mean)
+
+## How are migrations applied? 
+
+- Whenever you need to make changes to the currently applied schema, in the migrations folder, run `migrate create -ext sql -seq {{file_name}}`
```
wtam2018, line 21: I don't see where the `migrate` is made in Makefile.

wtam2018: It looks `migrate` binary is installed. Where does it come from?

db/migrations/000001_init_db.down.sql
```diff
@@ -0,0 +1,21 @@
...
+DROP TABLE IF EXISTS Operation;
+DROP TABLE IF EXISTS ClusterAccess;
+DROP TABLE IF EXISTS GitopsEngineInstance;
+DROP TABLE IF EXISTS GitopsEngineCluster;
+DROP TABLE IF EXISTS ManagedEnvironment;
+DROP TABLE IF EXISTS ClusterCredentials;
+DROP TABLE IF EXISTS ClusterUser;
+
+
+
```
drpaneas, line 19: Remove the unnecessary empty lines.

drpaneas (changes requested):
1. This is not a new component.
2. Use this via the Makefile.
3. Provide tutorial examples of migration up and down at https://github.com/redhat-appstudio/managed-gitops/tree/main/docs
4. See my other comments in your code.

drpaneas (changes requested):
Great work re-using the db libraries :D We still have some way to go before this is ready. Let me address my concerns:

1. As I said in the parking-lot, I don't see any strong evidence why this evaluates being a new component, and I suggested a better location would be to put it into `backend-shared/hacks/db_migration/main.go`. At that location, we already have _db-schema-sync-check_, so it looks like a natural place to put also _db-migration_ there as well. Grouping db related packages under the same roof (aka `backend-shared` module) is one of the benefits but there is also another one: you would **not** need to use the `replace` in the `go.mod` file, as they both will be part of the `backend-shared` dependencies. 

With the current approach, you are introducing a second `db` package, in the Root of the monorepo, so that would make 2 different db packages, namely:

* github.com/redhat-appstudio/managed-gitops/db
* github.com/redhat-appstudio/managed-gitops/backend-shared/config/db

As if this is not confusing enough, the first db package (`github.com/redhat-appstudio/managed-gitops/db`) imports the other "db" package (`github.com/redhat-appstudio/managed-gitops/backend-shared/config/db`). See the code, at:

```go
// backend/db/migrate_db.go
package main
// ... core removed for brevity
   db "github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
```

I can clearly myself getting confused with that structure... but since this is the 2nd time we are discussing the same topic, I don't want to force you into it, neither block you from merging this. So let's see what others think about this @jgwest @samyak-jn @wtam2018 any thoughts?

2. Also, as we discussed in the parking-lot, the documentation preferably goes into the _docs_ directory. If not, then at least put a link from docs to your `README.md`.

3. There are some old review comments from the previous review quest, but they still remain unaddressed. No worries though, I also find myself getting lost with too many comments as well. Looking at the "Files changed" tab might reveal them for you.

db/migrate_db.go
```diff
@@ -0,0 +1,52 @@
...
+			if err := m.Drop(); err != nil {
+				log.Error(err, fmt.Sprintf("%v", err))
+			}
+		} else {
+			log.Info("Invalid argument passed.")
+		}
+	} else {
+		// applies every migrations till the lastest migration
+		if err := m.Up(); err != nil {
+			log.Error(err, fmt.Sprintf("%v", err))
```
drpaneas, line 49: why not using `log.Errorf` instead?

samyak-jn (commented):
I indeed agree with Panos in regards to package name and position, db is used twice which brings confusion, simply either rename it to db-migrate or something similar to map a line difference!

For the position, we have three choices, I'll list priority wise with quick explanations
1. If we need to create a standalone go file out it which is only needed for migrations, we have utilities section present in the mono-repo, we have load-test sub dir, similarly one can be created for db-migrations (again keeping the nomenclature in mind)

2. Placing it here (github.com/redhat-appstudio/managed-gitops/backend-shared/config/db) as Panos suggested, but I found place 1 to be neater than this

3. If you want to keep the existing one, just change db to db-migrations 

Plus, I recommend squashing these commits into a meaningful bunch for easy navigation after addressing the review comments.

tirthyme:
Closing this pull request and raising a new one due to unsolvable dependency conflicts in the repo. For sake of this issue, moving the folder into utilities.

PR #119, opened 2022-05-13
Convert ArgoCD Sync tests to ginkgo #123

Rizwana777:
#### Description:
- This PR is about converting ArgoCD Sync tests to ginkgo in cluster-agent
- Test files which are converted are :
           cluster-agent/utils/argocd_login_credentials_test.go
           cluster-agent/utils/argocd_terminate_app_command_test.go
           cluster-agent/utils/argocd_login_command_test.go
           cluster-agent/utils/argocd_sync_command_test.go

https://issues.redhat.com/browse/GITOPSRVCE-123)

### DIFF
--- a/cluster-agent/utils/argocd_login_command_test.go
+++ b/cluster-agent/utils/argocd_login_command_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+// These tests are expected to run in parallel with other test cases, We can run the ginkgo tests parallelly by passing "ginkgo -p" command in CLI.
 var _ = Describe("ArgoCD Login Command", func() {
 	Context("ArgoCD Login Command Test", func() {
 		It("Ensure we are able to login to get a token”", func() {
--- a/cluster-agent/utils/argocd_login_command_test.go
+++ b/cluster-agent/utils/argocd_login_command_test.go
@@ -1,34 +1,31 @@
 package utils
 
 import (
-	"testing"
-
 	"github.com/argoproj/argo-cd/v2/pkg/apiclient/session"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/redhat-appstudio/managed-gitops/cluster-agent/utils/mocks"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
-func TestArgoCDLoginCommand(t *testing.T) {
+var _ = Describe("ArgoCD Login Command", func() {
+	Context("ArgoCD Login Command Test", func() {
+		It("Ensure we are able to login to get a token”", func() {
 
-	// Ensure we are able to login to get a token
+			mockSessionClient := &mocks.SessionServiceClient{}
+			mockAppClient := &mocks.Client{}
 
-	t.Parallel()
+			mockAppClient.On("NewSettingsClient").Return(mockCloser{}, nil, nil)
+			mockAppClient.On("NewSessionClient").Return(mockCloser{}, mockSessionClient, nil)
 
-	mockSessionClient := &mocks.SessionServiceClient{}
+			mockSessionClient.On("Create", mock.Anything, &session.SessionCreateRequest{
+				Username: "admin",
+				Password: "password",
+				Token:    "",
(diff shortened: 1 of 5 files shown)


cluster-agent/utils/argocd_login_command_test.go
```diff
@@ -1,34 +1,31 @@
...
 	"github.com/redhat-appstudio/managed-gitops/cluster-agent/utils/mocks"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
-func TestArgoCDLoginCommand(t *testing.T) {
-
-	// Ensure we are able to login to get a token
-
-	t.Parallel()
```
jparsai, line 16: I guess you discussed it in meetings and IIRC there was some approach suggested to execute ginkgo tests in parallel, is that right?

Rizwana777: Yes @jparsai I discussed this and we need to pass "ginkgo -p"  command to run ginkgo tests parallel

jparsai: Then it would be better if you mention it somewhere, maybe a comment at the starting of functions to inform others that these tests are expected to be executed parallelly.

Rizwana777: Yes @jparsai  added comment

Rizwana777:
@jparsai  Thank you for the review, updated changes , PTAL

cluster-agent/utils/argocd_login_command_test.go
jparsai: Is CI (PR Build) is also running it in parallel?

cluster-agent/utils/argocd_login_credentials_test.go
```diff
@@ -2,118 +2,108 @@ package utils
...
+				mockClient: mockAppClient,
+			}
+
+			cs := NewCredentialService(&clientGenerator, true)
+			Expect(err).To(BeNil())
+
+			creds, argoClient, err := cs.GetArgoCDLoginCredentials(context.Background(), "openshift-gitops", "12", true, k8sClient)
+			Expect(err).To(BeNil())
+			Expect(argoCDCredentials{} == creds).To(BeFalse())
+			Expect(argoClient).ToNot(BeNil())
```
jparsai, line 90: You are supposed to check for empty object here as well.

jparsai:
@Rizwana777 it LGTM, just need one minor change.

cluster-agent/utils/argocd_login_credentials_test.go
Rizwana777: For argoClient ? @jparsai

cluster-agent/utils/argocd_login_command_test.go
Rizwana777: I don't think so @jparsai

cluster-agent/utils/argocd_login_credentials_test.go
jparsai: Yes, in existing test I see empty check for argoClient as well.
https://github.com/redhat-appstudio/managed-gitops/blob/2fe8209b9711418db191393fc601eea2fbd6f2df/cluster-agent/utils/argocd_login_credentials_test.go#L102

Rizwana777: argoClient is "var argoClient argocdclient.Client"  here argocdclient.Client is an interface, so we cannot check for empty object @jparsai

jparsai: Okay, I see in case of failure `GetArgoCDLoginCredentials` returns `nil` for `argoClient`, so I guess no need to check for empty like existing test just check for `nil` is enough.
https://github.com/redhat-appstudio/managed-gitops/blob/2fe8209b9711418db191393fc601eea2fbd6f2df/cluster-agent/utils/argocd_login_credentials.go#L59

Rizwana777: Yup @jparsai  then there is no need to change this "Expect(argoClient).ToNot(BeNil())" right ?

jgwest (approved):
Looks great, thanks @Rizwana777 , and thanks for reviewing @jparsai !

PR #125, opened 2022-05-19
Adds postgres-server packager to dockerfile, req by openshift-ci

samyak-jn:
#### Description:
- This is in regard to the story for enabling openshift-ci on managed-gitops. Adding postgres-server for the Dockerfile which is required in order to access psql binary required by the ci!


samyak-jn:
Merging this @flacatus who is helping out with our openshift-ci enablement on our repository! The requirement for merging is urgent, will update it more on standup! All checks are passed for the same!

PR #131, opened 2022-05-26
Rename database fields and references from workspace to namespace where appropriate

jgwest:
#### Description:
- I previously used the term "workspace" as a catch-all term, to refer to the namespace where users were creating our API resources (`GitOpsDeployment` resource, etc).
- However, now that we are moving to KCP, which has the concept of workspaces, this use of 'workspace' is confusing. 
- I have thus updated the code from workspace -> namespace, or workspace -> API namespace as appropriate. 
- This PR only updates names/fields/database rows, there are no functional changes.


jparsai (approved):
LGTM

drpaneas (approved):
:D lgtm

samyak-jn (approved):
LGTM, thanks! :")

PR #135, opened 2022-05-30
Convert miscellaneous, remaining small unit tests to Ginkgo

tirthyme:
#### Description:
- Migrates multiple go tests to ginkgo.

[#126](https://issues.redhat.com/browse/GITOPSRVCE-126)


drpaneas (changes requested):
Hi Tirth,

1. Some tests (e.g. [db_field_constants_test.go](https://github.com/redhat-appstudio/managed-gitops/pull/135/files#diff-edc5e7fc4401f5af27394f8d85ef00f1f45064701e947a701b02c5cea91b1c37)) are written in a [Table Driven](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) fashion. The Ginkgo way to port those is documented at: https://onsi.github.io/ginkgo/#table-specs
2. There are some weird spacing issues at some parts of your code. @raghavi101 had the same once and she managed to resolve this, maybe she can help you on this.
3. Do not export functions that are not meant to be consumed by other pkg or people (e.g. the ones in the hack folder). You can use the same package for testing those (instead of package_test).

backend-shared/config/db/injection_test.go
```diff
@@ -1,281 +1,93 @@
...
-		Serviceaccount_ns:           "Serviceaccount_ns",
-	}
-
-	err = dbq.CreateClusterCredentials(ctx, &clusterCredentials)
-	assert.NoError(t, err)
-	{
-		managedEnvironment := ManagedEnvironment{
-			Managedenvironment_id: "test-managed-env-1",
-			Clustercredentials_id: clusterCredentials.Clustercredentials_cred_id,
-			Name:                  "test'env",
```
jgwest, line 132: Think a few of the old injection_tests were not ported over to the new tests. For example, here we are creating a managed environment with an name that might cause issues with the SQL logic (if we were doing it wrong).

backend-shared/config/db/db_field_constants.go
```diff
@@ -175,6 +175,6 @@ var DbFieldMap = map[string]int{
 }
 
 // Get value of constants based on constant variable name given as String.
-func getConstantValue(variable string) int {
+func GetConstantValue(variable string) int {
```
jgwest, line 178: As @drpaneas mentioned, this can be kept at `getConstantValue`, and the tests can be kept in the package they were defined in (e.g. `db`, rather than `db_test` package... unless there was a good reason to move them to a new package?)

jgwest (approved):
LGTM, thanks @tirthyme! I'm happy to see with this PR that we are finally to remove `SetupforTestingDB`, because with this PR all of our non-Ginkgo tests have been moved over :tada:

PR #141, opened 2022-06-03
Integrate migration at backend startup.

tirthyme:
#### Description:
- Calls DB Migration code when backend starts up.


jgwest:
@samyak-jn @tirthyme I think the prow job is failing because I made [the timeout value here](https://github.com/redhat-appstudio/managed-gitops/blob/1fb4d804a8c2d949c14fc64f3fbd73522a0b31cf/backend/Makefile#L52) too short, feel free to increase it in your PR (double it to 30, for example).

utilities/db-migration/migrate/migrate_db.go
```diff
@@ -12,41 +12,56 @@ import (
 	db "github.com/redhat-appstudio/managed-gitops/backend-shared/config/db"
 )
 
-func Main(op_type string) {
+func Main(op_type string, call_type string) bool {
```
jgwest, line 15: Recommend tweaking the method signature to:
```
func Migrate(opType string, migrationPath string) error {
}
```

So that's:
- Change the function name to Migrate (since it's no longer a main function)
- Inside `Migrate` return an error when an error occurs, rather than returning a bool.
- Change params to go parameter syntax
- Migrate should take the path to the migrations dir, rather than a constant indicating the type. It should be up to functions that are calling Migrate to figure out where the path is.

utilities/db-migration/main.go
```diff
@@ -3,13 +3,13 @@ package main
...
+	migrate "github.com/redhat-appstudio/managed-gitops/utilities/db-migration/migrate"
 )
 
 func main() {
 	op_type := ""
 	if len(os.Args) >= 2 {
 		op_type = os.Args[1]
 	}
-	migrate.Main(op_type)
+	migrate.Main(op_type, "migrate")
```
jgwest, line 14: Recommend tweaking this to
```golang
err := migrate.Migrate(opType, "migrate")
if err != nil {
    // output the error
    os.Exit(1) // terminate the process with an error code.
    return
}
```

utilities/db-migration/migrate/migrate_db.go
```diff
@@ -12,41 +12,56 @@ import (
...
 		log.Error(err, fmt.Sprintf("%v", err))
+		return false
 	}
 	if len(op_type) > 0 {
 		// op_type := os.Args[1]
 		if op_type == "drop_smtable" {
 			dbq, err := db.ConnectToDatabaseWithPort(true, "postgres", port)
 			if err != nil {
 				log.Error(err, fmt.Sprintf("%v", err))
+				return false
```
jgwest, line 40: Can replace all of these log+returns, with, for example,  `return fmt.Errorf("unable to connect to DB: %v", err)`

backend/main.go
```diff
@@ -66,6 +67,10 @@ func init() {
 }
 
 func main() {
+	if !migrate.Main("", "backend") {
+		setupLog.Error(fmt.Errorf("migration was not successful"), "Fatal Error: Unsuccessful Migration")
```
jgwest, line 71: Can change this to:
```golang
if err := migrate.Migrate("", "(path to migrations folder)"); err != nil {
    setupLog.Error(err, "SEVERE: Unsuccessful Migration")
    os.Exit(1)
}
```

jgwest (changes requested):
Looks good! I haven't had a change to fully test it yet, but have a few minor tweaks requested, see below for details.

Re: rebasing, don't worry about rebasing, I'll squash your commits when I merge. You may want to see if you can determine what it is in your workflow that is causing issues conflicts. I would personally recommend:
- Avoid using the GitHub UI button which claims to rebase your commits 
- About using the `git merge` command.

jgwest (approved):
LGTM, thanks @tirthyme! It's great to get the migration logic now fully supported in our code, so I don't have to do hacky manual updates to the schema of the staging server every time :sweat_smile:.

PR #143, opened 2022-06-06
Periodic Reconciliation of Workspaces/Namespaces.

jparsai:
# Description
This PR is to add a function into Cluster Agent which iterates through each Workspace/Namespace present in DB after a fixed interval and ensures that the state of resources (Applications) in Cluster is in Sync with DB.

# Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-57

### DIFF
--- a/backend-shared/config/db/application_test.go
+++ b/backend-shared/config/db/application_test.go
@@ -82,4 +82,54 @@ var _ = Describe("Application Test", func() {
 		Expect(true).To(Equal(db.IsMaxLengthError(err)))
 
 	})
+
+	It("Should Get Application in batch.", func() {
+		err := db.SetupForTestingDBGinkgo()
+		Expect(err).To(BeNil())
+
+		ctx := context.Background()
+		dbq, err := db.NewUnsafePostgresDBQueries(true, true)
+		Expect(err).To(BeNil())
+		defer dbq.CloseDatabase()
+
+		_, managedEnvironment, _, gitopsEngineInstance, _, err := db.CreateSampleData(dbq)
+		Expect(err).To(BeNil())
+
+		// Create multiple application entries.
+		applicationput := db.Application{
+			Application_id:          "test-my-application-1",
+			Name:                    "my-application",
+			Spec_field:              "{}",
+			Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
+			Managed_environment_id:  managedEnvironment.Managedenvironment_id,
+		}
+
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		applicationput.Application_id = "test-my-application-2"
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		applicationput.Application_id = "test-my-application-3"
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		applicationput.Application_id = "test-my-application-4"
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		applicationput.Application_id = "test-my-application-5"
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		var listOfApplicationsFromDB []db.Application
+		err = dbq.GetApplicationBatch(ctx, &listOfApplicationsFromDB, 2, 0)
+		Expect(err).To(BeNil())
+		Expect(len(listOfApplicationsFromDB)).To(Equal(2))
+
+		err = dbq.GetApplicationBatch(ctx, &listOfApplicationsFromDB, 3, 1)
+		Expect(err).To(BeNil())
+		Expect(len(listOfApplicationsFromDB)).To(Equal(3))
+	})
 })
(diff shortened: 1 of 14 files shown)


cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -163,6 +169,153 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+						}
+					} else {
+						log.Error(err, "Error occurred in Namespace Reconciler while fetching application from cluster: "+applicationRowFromDB.Application_id)
+						continue
+					}
+				}
+
+				// At this point we have the applications from ArgoCD and DB, now compare them to check if they are not in Sync.
+				if CompareApplications(applicationFromArgoCD, applicationFromDB, log) {
+					log.Info("Argo application is in Sync with DB, Application:" + applicationRowFromDB.Application_id)
```
jgwest, line 267: Can change this to log at debug level, because I suspect this will be very verbose with a large number of Applications.

cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -163,6 +169,153 @@ func (r *ApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+				} else {
+					log.Info("Argo application is not in Sync with DB, Updating ArgoCD App. Application:" + applicationRowFromDB.Application_id)
+				}
+
+				// At this point application from ArgoCD and DB are not in Sync, so need to update ArgoCD application according to DB entry
+
+				// Get Special user from DB because we need ClusterUser for creating Operation and we don't have one.
+				// Hence created a dummy Cluster User for internal purpose.
+				var specialClusterUser db.ClusterUser
+				if err := r.DB.GetOrCreateSpecialClusterUser(ctx, &specialClusterUser); err != nil {
```
jgwest, line 278: You can move this GetOrCreate call to outside the for loop: we don't need to retrieve it for every iteration. You move this to where `log.Info("Triggered Namespace Reconciler to keep Argo application in sync with DB.")`, is currently, I believe.

cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -229,3 +382,118 @@ func CompressResourceData(resources []appv1.ResourceStatus) ([]byte, error) {
...
+		// Fetch corresponding DB entry
+		dbOperation := db.Operation{
+			Operation_id: k8sOperation.Spec.OperationID,
+		}
+		if err := dbq.GetOperationById(ctx, &dbOperation); err != nil {
+			log.Error(err, "Unable to fetch DB Operation entry for "+string(k8sOperation.Spec.OperationID))
+			continue
+		}
+
+		if dbOperation.State != "Completed" {
```
jgwest, line 484: I think this should be `if dbOperation.State != db.OperationState_Completed || dbOperation.State != db.OperationState_Failed`, feel free to correct my logic if wrong.

cluster-agent/main.go
```diff
@@ -120,6 +121,25 @@ func main() {
 	}
 	//+kubebuilder:scaffold:builder
 
+	//==============================================
+	// Process to trigger Namespace Reconciler
+
+	// Get Special user created for internal use,
+	// because we need ClusterUser for creating Operation and we don't have one.
+	// Hence created a dummy Cluster User for internal purpose.
+	var speCialClusterUser db.ClusterUser
```
jgwest, line 130: Tweak: `specialClusterUser`

jgwest (changes requested):
Looks great, nice work @jparsai! I added some a number of review comments, but they are minor, and should be easy to tweak. Let me know if any of my comments need further details/explanations, happy to oblige.

backend-shared/config/db/application_test.go
```diff
@@ -82,4 +82,54 @@ var _ = Describe("Application Test", func() {
...
+		Expect(err).To(BeNil())
+
+		applicationput.Application_id = "test-my-application-5"
+		err = dbq.CreateApplication(ctx, &applicationput)
+		Expect(err).To(BeNil())
+
+		var listOfApplicationsFromDB []db.Application
+		err = dbq.GetApplicationBatch(ctx, &listOfApplicationsFromDB, 2, 0)
+		Expect(err).To(BeNil())
+		Expect(len(listOfApplicationsFromDB)).To(Equal(2))
```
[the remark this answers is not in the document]
jparsai: This is good to have, but it is not possible as of now because there are other tests present in code where DB entries are not deleted after test completion, and rows we are creating here are written on top of it, now for example if we try to read first two rows the we would get application data which is saved by other tests and application ID won't match so tests will fail.

backend-shared/config/db/application_test.go
```diff
@@ -82,4 +82,54 @@ var _ = Describe("Application Test", func() {
...
+
+		err = dbq.GetApplicationBatch(ctx, &listOfApplicationsFromDB, 5, 1)
+		Expect(err).To(BeNil())
+		Expect(len(listOfApplicationsFromDB)).To(Equal(5))
```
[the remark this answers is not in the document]
jparsai: It was because of the same issue I mentioned above, I am inserting 5 rows. Now offset from 2 to 6 should return 4 rows as offset 6 is not present, but there are old application rows so offset 6 is no longer empty and we are getting 5 rows instead of 4. Now I changed the offset and batch size.

jgwest (approved):
LGTM, thanks @jparsai !

PR #146, opened 2022-06-10
Fix sanity test failure in task retry loop, fix processOperation_Application logic, constantize databaseID

jgwest:
#### Description:
- Fix sanity test failure in task retry loop, due to adding values to one field of `waitingTaskContainer` but not the other (this was an oversight when the code was written)
- Re-examine the logic of `processOperation_Application`, as it was not retrying for some errors, when it should be.
- Converting `databaseID` into a constant string, and some minor refactor of log statements


jgwest:
@drpaneas @wtam2018 @jparsai Any review comments before I merge?

PR #151, opened 2022-06-14
Update ServiceAccount client to use client.Client, refactor ServiceAccount creation logic, tests

jgwest:
#### Description:
- Update ServiceAccount client to use client.Client, rather than a clientset
- Refactor ServiceAccount creation logic, switch to cluser-scoped clusterrole/rolebinding
- Tweak tests, switch to Ginkgo v2


samyak-jn (commented):
The interesting part to see, the unit tests from openshift-ci are passing, whereas from GitHub Actions are failing :exploding_head:

jgwest:
@samyak-jn Makes sense in this case, actually, because the tests requires a K8s cluster: which openshift CI has and github actions doesn't. I broke the Skip() call, whose job was to skip the test if there isn't a k8s cluster present.

jgwest:
No review comments after a week, and the changes are uncontroversial, so merging.

PR #152, opened 2022-06-15
Argocd standalone setup

raghavi101:
#### Description:
Create utility code that sets up a new, standalone Argo CD instance, plus E2E test to ensure it works with manual sync code

https://issues.redhat.com/browse/GITOPSRVCE-142


cluster-agent/utils/argocd_create.go
```diff
@@ -0,0 +1,453 @@
...
+	// token := secret.Data["token"]  not sure whether this will work
+	token := secretResource.Data["token"]
+	// no need to decode token, it is unmarshalled from base64
+
+	// decodedToken, err := base64.StdEncoding.DecodeString(string(token))
+	// if err != nil {
+	// 	fmt.Printf("Error decoding Base64 encoded data %v", err)
+	// }
+
+	type ClusterSecretTLSClientConfigJSON struct {
```
jgwest, line 396: TIL that you can define types within Go functions :smile:. This is useful information, but I think we can just move these type structs out of the function, for now.

cluster-agent/utils/argocd_create.go
```diff
@@ -0,0 +1,453 @@
...
+	//   server: $API_URL
+	//   config: |
+	//     {
+	//       "bearerToken": "$SECRET_TOKEN",
+	//       "tlsClientConfig": {
+	//         "insecure": true
+	//       }
+	//     }
+
+	secretResource, secretErr := kubeClientSet.CoreV1().Secrets(secret.Namespace).Get(context.Background(), secret.Name, metav1.GetOptions{})
```
jgwest, line 383: I believe we need to use a `wait.Poll(...)` here, similar to what the [gist](https://gist.github.com/jgwest/64cdc63a978324958ab8ac91aea74700) was doing on line 86 to 96. In the case of the gist, it was using a for loop, but in Go a `wait.Poll` would be the preferred idiom.

It would be, something like this, I believe:
```golang
wait.Poll(intervals, func() bool {
    // get the secret
    // check if the secret has a token field, which is non-empty
    // return true (indicating success!) if the secret exists and has the token, otherwise false
})
```
That way, the code will keep waiting for the secret to exist. (otherwise, it's a race condition)

jgwest (changes requested):
Looks great @raghavi101, I added some review comments that are mainly around changes to how we are calling the K8s API, and some minor opinionated tweaks :smile:, feel free to take a look.

I haven't had a chance to test the E2E tests, due to issues with installing the GitOps Operator (mentioned on internal slack), but will try that again on the next review round.

PR #153, opened 2022-06-15
Fix CI: Switch to using migration to initialize the database scheme when running on k8s, increase timeouts

jgwest:
#### Description:
- Fixes a OpenShift CI failure that is being seen on PRs.
- Switches to using the migration logic to initialize the database scheme when running on k8s, similar to what we are doing in all other cases (docker devenv, staging cluster)
- Increase timeouts to make sure we wait long enough for PostgreSQL to come up

### DIFF
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -10,10 +10,10 @@ exit_if_binary_not_installed() {
   done
 }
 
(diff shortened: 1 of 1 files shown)


samyak-jn (approved):
LGTM, thanks @jgwest!

PR #154, opened 2022-06-17
Enable prune, allowEmpty, and selfHeal on generated Argo CD Applications

chetan-rns:
#### Description:
- Enable prune, allowEmpty, and selfHeal on an Argo CD Application generated for GitOpsDeployment with Automated type.
- Enable the above fields in the backend before adding the application spec to the database.
- Check if there is any diff in the syncPolicy in the `operation_event_loop`
- Add unit test and e2e test to verify the above changes

https://issues.redhat.com/browse/GITOPSRVCE-164


backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
```diff
@@ -0,0 +1,97 @@
...
+		It("Input spec with automated enabled should set automated sync policy", func() {
+			input := fakeInputSpec(true, false)
+			application, err := createSpecField(input)
+			Expect(err).To(BeNil())
+			Expect(application).To(Equal(validApplicationWithAutomatedSync()))
+		})
+	})
+})
+
+func fakeInputSpec(automated, unsanitized bool) argoCDSpecInput {
```
drpaneas, line 33: The name of the function is to generic, as `fakeInputSpec` could mean anything within the context of `application_event_loop` package. Just a alternative suggestion:

```suggestion
func getFakeArgoCDSpecInput(automated, unsanitized bool) argoCDSpecInput {
```

backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if unsanitized {
+		input.sourcePath = "environments/&&prod\n"
+		input.sourceRepoURL = "https://github.com/```%test/test"
+	}
+	return input
+}
+
+func validApplicationWithAutomatedSync() string {
```
drpaneas, line 51: why not create a `const` instead ?

tests-e2e/core/argocd_application_test.go
```diff
@@ -0,0 +1,79 @@
...
+					}
+					Expect(err).To(BeNil())
+				}
+
+				syncPolicy := app.Spec.SyncPolicy
+				if syncPolicy != nil && syncPolicy.Automated != nil {
+					return syncPolicy.Automated.AllowEmpty && syncPolicy.Automated.Prune && syncPolicy.Automated.SelfHeal
+				}
+				return false
+			}, "5m", "1s").Should(BeTrue())
```
drpaneas, line 75: Just a thought, what about creating a `timeouts.go` and putting the all there instead of hardcoded values inside individual tests?

drpaneas (commented):
Great work @chetan-rns  💯 👍 
Some suggestions and questions from my side only.

backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
chetan-rns: Thanks. Updated the name

backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
chetan-rns: Yes. It makes sense to use const here. I've generalized it further to use structs in a function so we can unmarshall it when required. It avoids duplication and we no longer have to maintain raw YAML strings.

tests-e2e/core/argocd_application_test.go
chetan-rns: Currently, the timeout duration varies with test specs. Should I use a common timeout for all the specs?

drpaneas: Hm, let's discuss this in parking lot.

drpaneas (approved):
@chetan-rns great work!!!
/lgtm

jgwest (approved):
Nice work, thanks @chetan-rns!  Great first merged repository PR, welcome to the team and to the project! :tada: :tada: :tada:

PR #158, opened 2022-06-22
E2e test for GitOpsDeployment API

raghavi101:
#### Description:
This PR contains e2e test for GitOpsDeployment API

https://issues.redhat.com/browse/GITOPSRVCE-154

### DIFF
--- a/tests-e2e/core/gitopsDeployment_test.go
+++ b/tests-e2e/core/gitopsDeployment_test.go
@@ -5,10 +5,14 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	routev1 "github.com/openshift/api/route/v1"
 	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend/apis/managed-gitops/v1alpha1"
 	fixture "github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture"
 	gitopsDeplFixture "github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture/gitopsdeployment"
 	"github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture/k8s"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
 	typed "k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -16,6 +20,12 @@ import (
 const (
 	name    = "my-gitops-depl"
 	repoURL = "https://github.com/redhat-appstudio/gitops-repository-template"
+
+	// ArgoCDReconcileWaitTime is the length of time to watch for Argo CD/GitOps Service to deploy the resources
+	// of an Application (e.g. to reconcile the Application resource)
+	// We set this to be at least twice the default reconciliation time (3 minutes) to avoid a race condition in updating ConfigMaps
+	// in Argo CD. This is an imperfect solution.
+	ArgoCDReconcileWaitTime = "7m"
 )
 
 var _ = Describe("GitOpsDeployment E2E tests", func() {
@@ -25,7 +35,121 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 		Expect(err).To(BeNil())
 		ctx := context.Background()
 
-		It("Should ensure succesfull creation of GitOpsDeployment, by creating the GitOpsDeployment", func() {
+		AfterEach(func() {
+			// At end of a test, output the state of Argo CD Application, for post-mortem debuggign
+			err := fixture.ReportRemainingArgoCDApplications(k8sClient)
+			Expect(err).To(BeNil())
+		})
+
+		// Generate expected resources for "https://github.com/redhat-appstudio/gitops-repository-template"
+		getResourceStatusList_GitOpsRepositoryTemplateRepo := func(name string) []managedgitopsv1alpha1.ResourceStatus {
+			return []managedgitopsv1alpha1.ResourceStatus{
+				{
+					Group:     "apps",
+					Version:   "v1",
+					Kind:      "Deployment",
+					Namespace: fixture.GitOpsServiceE2ENamespace,
+					Name:      name,
+					Status:    managedgitopsv1alpha1.SyncStatusCodeSynced,
+					Health: &managedgitopsv1alpha1.HealthStatus{
+						Status: managedgitopsv1alpha1.HeathStatusCodeHealthy,
+					},
+				},
+				{
+					Group:     "route.openshift.io",
+					Version:   "v1",
+					Kind:      "Route",
+					Namespace: fixture.GitOpsServiceE2ENamespace,
+					Name:      name,
+					Status:    managedgitopsv1alpha1.SyncStatusCodeSynced,
+					Health: &managedgitopsv1alpha1.HealthStatus{
+						Status:  managedgitopsv1alpha1.HeathStatusCodeHealthy,
+						Message: "Route is healthy",
+					},
+				},
+				{
+					Group:     "",
+					Version:   "v1",
+					Kind:      "Service",
+					Namespace: fixture.GitOpsServiceE2ENamespace,
+					Name:      name,
+					Status:    managedgitopsv1alpha1.SyncStatusCodeSynced,
+					Health: &managedgitopsv1alpha1.HealthStatus{
+						Status: managedgitopsv1alpha1.HeathStatusCodeHealthy,
+					},
+				},
+			}
+		}
+
+		// Generate  expected resource from "https://github.com/managed-gitops-test-data/deployment-permutations-*" using name
+		getResourceStatusList_deploymentPermutations := func(name string) []managedgitopsv1alpha1.ResourceStatus {
+			return []managedgitopsv1alpha1.ResourceStatus{
+				{
+					Group:     "",
+					Version:   "v1",
+					Kind:      "ConfigMap",
+					Name:      name,
+					Namespace: fixture.GitOpsServiceE2ENamespace,
+					Status:    managedgitopsv1alpha1.SyncStatusCodeSynced,
+				},
+			}
+		}
+
+		// Generate expected identifier configmap for "https://github.com/managed-gitops-test-data/deployment-permutations-*"
+		createResourceStatusListFunction_deploymentPermutations := func() []managedgitopsv1alpha1.ResourceStatus {
+			return []managedgitopsv1alpha1.ResourceStatus{
+				{
+					Group:     "",
+					Version:   "v1",
+					Kind:      "ConfigMap",
+					Name:      "identifier",
+					Namespace: fixture.GitOpsServiceE2ENamespace,
+					Status:    managedgitopsv1alpha1.SyncStatusCodeSynced,
+				},
+			}
+		}
+
+		// Returns true if the all param resources no longer exist (have been deleted), and false otherwise.
+		expectAllResourcesToBeDeleted := func(expectedResourceStatusList []managedgitopsv1alpha1.ResourceStatus) {
+
+			Eventually(func() bool {
+
+				k8sclient, err := fixture.GetKubeClient()
+				Expect(err).To(BeNil())
+
+				for _, resourceValue := range expectedResourceStatusList {
+					ns := typed.NamespacedName{
+						Namespace: resourceValue.Namespace,
+						Name:      resourceValue.Name,
+					}
+
+					var obj client.Object
+
+					if resourceValue.Kind == "ConfigMap" {
+						obj = &corev1.ConfigMap{}
+					} else if resourceValue.Kind == "Deployment" {
+						obj = &appsv1.Deployment{}
+					} else if resourceValue.Kind == "Route" {
+						obj = &routev1.Route{}
+					} else if resourceValue.Kind == "Service" {
+						obj = &corev1.Service{}
+					} else {
+						GinkgoWriter.Println("unrecognize kind:", resourceValue.Kind)
+						return false
+					}
+					err := k8sclient.Get(context.Background(), ns, obj)
+					if !apierr.IsNotFound(err) {
+						return false
+					}
+				}
+
+				// If all the resources were found, return true
+				return true
+			}, "2m", "1s").Should(BeTrue())
+
+		}
+
+		It("Should ensure succesful creation of GitOpsDeployment, by creating the GitOpsDeployment", func() {
 
 			Expect(fixture.EnsureCleanSlate()).To(Succeed())
 			gitOpsDeploymentResource := buildGitOpsDeploymentResource(name,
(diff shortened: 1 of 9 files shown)


tests-e2e/core/gitopsDeploymentConditions_test.go
```diff
@@ -0,0 +1,223 @@
+package core
```
jparsai, line 1: I think file name can be changed to `gitOpsDeployment_test.go` to align with functionality it is testing.
`gitopsDeploymentConditions_test.go` would be conflicting with file for `gitOpsDeployment.status.conditions` tests.

jparsai (changes requested):

jparsai:
Is it ready for review?

raghavi101:
> 



> Is it ready for review?

Yes

tests-e2e/core/simple_test.go
```diff
@@ -84,3 +84,24 @@ func buildGitOpsDeploymentResource(name, repoURL, path, deploymentSpecType strin
 
 	return gitOpsDeploymentResource
 }
+
+func buildTargetRevisionGitOpsDeploymentResource(name, repoURL, path, target, deploymentSpecType string) managedgitopsv1alpha1.GitOpsDeployment {
```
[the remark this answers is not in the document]
jparsai: If you dont want to add an extra argument in existing function then you can reuse it in new function.
I still dont agree with idea to duplicate 9 out of 10 lines from existing function into new one, just to add a new line to set `TargetRevision`.  
```
func buildTargetRevisionGitOpsDeploymentResource(name, repoURL, path, target, deploymentSpecType string) managedgitopsv1alpha1.GitOpsDeployment {
	gitOpsDeploymentResource:= buildGitOpsDeploymentResource(name, repoURL, path, deploymentSpecType)
	gitOpsDeploymentResource.Spec.Source.TargetRevision = target
	return gitOpsDeploymentResource
}
```

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+		})
+
+		It("Checks whether a change in repo URL is reflected within the cluster", func() {
+
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+
+			By("creating a new GitOpsDeployment resource")
+			gitOpsDeploymentResource := buildGitOpsDeploymentResource("gitops-depl-test-status",
+				"https://github.com/redhat-appstudio/managed-gitops", "environments/overlays/dev",
```
jparsai, line 131: For now it is Okay, but in future after we create bunch of sample repositories (each with different resource/kind/pattern) for testing, it would be better to use a valid deployment here. Since this is a invalid deployment it is will anyway fail and wont create any valid resources, so we are not actually testing the real usecase here, like how a real user of GitOps-service would use it.

jparsai (approved):
LGTM

samyak-jn (commented):
Hi @raghavi101

Thanks for working on this, can you dig up the logs to check why this e2e is failing on the openshift-ci? Otherwise, this looks good to me as well.

raghavi101:
Hi @samyak-jn it looks like the test timed out waiting for app to be synced and healthy at : 
`tests-e2e/core/argocd_application_test.go:33`

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+			err = k8s.Delete(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+			err = k8s.Get(&gitOpsDeploymentResource)
+			Expect(err).ToNot(Succeed())
+		})
+
+		It("Should ensure synchronicity of create and update of GitOpsDeployment", func() {
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+			By("ensuring GitOpsDeployment should update successfully on changing value(s) within Spec")
```
jgwest, line 92: Can move this `By` into `It`, as well, I think.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			By("ensurng no updates are done in existing deployment, if CR is submitted again without any changes")
+
+			gitOpsDeploymentResource := buildGitOpsDeploymentResource(name,
+				repoURL, "environments/overlays/dev",
+				managedgitopsv1alpha1.GitOpsDeploymentSpecType_Automated)
+
+			err := k8s.Create(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+
+			gitOpsDeploymentResource.Spec.Source.RepoURL = repoURL
```
jgwest, line 67: Think we should add this here, after Create:
```
		Eventually(gitOpsDeploymentResource, "2m", "1s").Should(
				SatisfyAll(
					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
				),
			)
```
We need to wait for Argo CD to finish deploying the created Argo CD `Application` before we go on to update it.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			Expect(gitOpsDeploymentResource.Spec.Source.RepoURL).To(Equal(repoURL))
+			Expect(gitOpsDeploymentResource.Spec.Source.Path).To(Equal("environments/overlays/dev"))
+
+			Eventually(gitOpsDeploymentResource, "2m", "1s").Should(
+				SatisfyAll(
+					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
+					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
+				),
+			)
+
```
jgwest, line 83: After this `Eventually`, we can add a `Consistently` that should ensure that it continues to be `Synced` and `Healthy` for 10 seconds.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			Expect(err).To(Succeed())
+			err = k8s.Get(&gitOpsDeploymentResource)
+			Expect(err).ToNot(Succeed())
+
+		})
+
+		It("Should ensure synchronicity of create and update of GitOpsDeployment", func() {
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+
+			By("ensurng no updates are done in existing deployment, if CR is submitted again without any changes")
```
jgwest, line 58: I think this `By` can instead be included in the name of the test itself, 
`It("Should ensure synchronicity of create and update of GitOpsDeployment, by ensuring no updates (...)"`), since I believe this is describing what the entire `It()` block is doing, and not just the individual line.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			gitOpsDeploymentResource.Spec.Source.Path = "environments/overlays/dev"
+
+			err = k8s.Update(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+
+			err = k8sClient.Get(ctx, client.ObjectKeyFromObject(&gitOpsDeploymentResource), &gitOpsDeploymentResource)
+			Expect(gitOpsDeploymentResource.Spec.Source.RepoURL).To(Equal(repoURL))
+			Expect(gitOpsDeploymentResource.Spec.Source.Path).To(Equal("environments/overlays/dev"))
+
+			Eventually(gitOpsDeploymentResource, "2m", "1s").Should(
```
jgwest, line 77: Rather than use an `Eventually` here, we should use a `Consistently`, like this:
```
Consistently(gitOpsDeploymentResource, "20s", "1s").Should(
				SatisfyAll(
					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
				),
			)
```
(pseudocode)

This means, check every 1 second, for 20 seconds, to make sure that it is still synced and healthy. If it has any other status, even for a second, then the test will fail (which is desired).

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+		It("Should ensure synchronicity of create and update of GitOpsDeployment", func() {
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+			By("ensuring GitOpsDeployment should update successfully on changing value(s) within Spec")
+
+			gitOpsDeploymentResource := buildGitOpsDeploymentResource(name,
+				"https://github.com/redhat-appstudio/managed-gitops", "manifests/appstudio-controller-rbac",
+				managedgitopsv1alpha1.GitOpsDeploymentSpecType_Automated)
+
+			err := k8s.Create(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
```
jgwest, line 99: After create, we need to wait for Argo CD to deploy the resource (e.g. an `Eventually` call). 

One thing to note: just because a K8s resource has been created, it does not mean that the particular action that a resource describes has completed.

For example, in this case, just because we have created a `GitOpsDeployment`, e.g. just because the `Create` call has completed, it does not mean the the Git repository has been deployed to the cluster.

Calling `k8s.Create(...)` only _starts_ the process, it does not indicate that it is complete. (e.g. it is an asynchronous operation)

So, as mentioned, here we should wait for Synched/Healthy, and it also couldn't hurt to wait for (`Eventually`) at least one K8s resource from the Git repository was deployed.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+			gitOpsDeploymentResource.Spec.Source.RepoURL = repoURL
+			gitOpsDeploymentResource.Spec.Source.Path = "environments/overlays/dev"
+
+			err = k8s.Update(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+
+			err = k8sClient.Get(ctx, client.ObjectKeyFromObject(&gitOpsDeploymentResource), &gitOpsDeploymentResource)
```
jgwest, line 107: Likewise, after `Update`, and before we our `Synced`/`Healthy`, we should wait for (`Eventually`) a resource from the Git repository to be deployed. This resource should be one that is ONLY in the new repository URL, so that we can be sure that an update has taken place. See parent review comment for some further details on this.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+
+			By("ensuring GitOpsDeployment Fails to create and deploy application when an invalid field input is passed")
+			gitOpsDeploymentResource := buildGitOpsDeploymentResource("",
+				"abc", "path/path/path",
+				managedgitopsv1alpha1.GitOpsDeploymentSpecType_Automated)
+
+			err := k8s.Create(&gitOpsDeploymentResource)
+			Expect(err).ToNot(Succeed())
```
jgwest, line 519: Depending on what you are testing with this test case, I think the `Create` call should still succeed: If create fails, it means K8s was not able to create the object, and there is no point in testing anything else (because the object being tested will never exist, and thus GitOps Service won't receive any events for it, and thus no action will take place.).

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+			Eventually(gitOpsDeploymentResource, "2m", "1s").ShouldNot(
```
jgwest, line 521: I believe this should be an `Eventually` followed by a `Consistently`.

The Eventually should wait for it to have the target expected state (`unhealthy`/`unsynced`, I guess), then Consistently should ensure it stays in that state for 60 seconds (checking once per second).

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+				SatisfyAll(
+					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
+					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
+				),
+			)
+
+			err = k8s.Get(&gitOpsDeploymentResource)
+			Expect(err).ToNot(Succeed())
```
jgwest, line 529: Likewise I would expect this to succeed, if the `Create` succeeded.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			err = k8s.Delete(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+			err = k8s.Get(&gitOpsDeploymentResource)
+			Expect(err).ToNot(Succeed())
+
+		})
+
+		It("Checks whether a change in repo URL is reflected within the cluster", func() {
+
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
```
jgwest, line 127: Recommend refactoring this update test, see parent comment for details.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+				Expect(err).ToNot(Succeed())
+			}
+
+		})
+
+		It("Checks whether a change in source path is reflected within the cluster", func() {
+
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+
+			By("creating a new GitOpsDeployment resource")
```
jgwest, line 226: Recommend refactoring this update test, see parent comment for details.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+
+				err := k8sClient.Get(ctx, ns, &gitOpsDeploymentResource)
+				Expect(err).ToNot(Succeed())
+			}
+
+		})
+
+		It("Checks whether a change in target revision is reflected within the cluster", func() {
+
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
```
jgwest, line 322: Recommend refactoring this update test, see parent comment for details.

tests-e2e/core/gitopsDeployment_test.go
```diff
@@ -0,0 +1,534 @@
...
+			err := k8s.Create(&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+
+			Eventually(gitOpsDeploymentResource, "2m", "1s").Should(
+				SatisfyAll(
+					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
+					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
+				),
+			)
+			err = k8s.Delete(&gitOpsDeploymentResource)
```
jgwest, line 48: After `Delete`, you should verify that all the resources deployed from Git, by Argo CD, have been cleaned up from the namespace.

jgwest (changes requested):
Hey Raghavi, looks good! I've got some refactoring suggestions below for the 3 update tests: these are mostly consistent with the original story critera, but do provide some additional details not present in the original story.

For the following tests:
- 'Checks whether a change in repo URL is reflected within the cluster'
- 'Checks whether a change in source path is reflected within the cluster'
- 'Checks whether a change in target revision is reflected within the cluster'
it appears the currently implementation is not fully testing the update functionality, mainly because we are not verifying k8s resources after create, then after update.

An update test would generally follow this format:
1) Create a `GitOpsDeployment`
2) Wait for the resources described in Git repository of the GitOpsDeployment to be deployed.
3) Wait for `GitOpsDeployment` to be synced/healthy
4) Make a change to the source (repo url/branch/path) field of the `GitOpsDeployment`
5) Wait for a resource described ONLY in the new source to be deployed, and wait for any old resources to be deleted.
6) Wait for `GitOpsDeployment` to be synced/healthy

In order to help with this, I've created a new GitHub organization (https://github.com/orgs/managed-gitops-test-data/) and some new Git repositories (https://github.com/orgs/managed-gitops-test-data/repositories). (This was what Jayendra [mentioned in this comment](https://github.com/redhat-appstudio/managed-gitops/pull/158/files#r911682608), but I've gone ahead and done it now :smile: . )
	
These should be useful for the GitOpsDeployment tests where you are changing the repo url/path/branch.

The new version of the test, using these new repositories, would look like this:
1) Create a `GitOpsDeployment` resource targeting:
	- repoUrl: https://github.com/managed-gitops-test-data/deployment-permutations-a
	- branch: branchC
	- path: pathB
2) Wait for Argo CD to deploy the resources:
	- Wait for resource `ConfigMap` resource, with name `identifier` to exist. Wait for it to have data.identifier value of  "(deployment-permutations-a)-(branchC)-(pathB)"
	- Wait for resource `ConfigMap` resource, with name `deployment-permutations-a-branchC-pathB` to exist.
	- Wait for GitOpsDeployment to show sync status of healthy/synced.
3) Update the `GitOpsDeployment` resource, for example, now targeting:
	- repoUrl: https://github.com/managed-gitops-test-data/deployment-permutations-b
	- branch: branchA
	- path: pathC
4)  Wait for Argo CD to deploy the resources:	
	- Wait for resource `ConfigMap` resource, with name `identifier` to exist. Wait for it to have data.identifier value of  "(deployment-permutations-a)-(branchC)-(pathB)"
	- Wait for resource `ConfigMap` resource, with name `deployment-permutations-a-branchC-pathB` to exist.
	- Wait for GitOpsDeployment to show sync status of healthy/synced.

Also take a look at the documentation under the https://github.com/orgs/managed-gitops-test-data/ repositories for more details on the repository format.

I recommend keeping at least one of the existing tests you have for these 3, as they are doing an excellent job of verifying that the individual resources are being created. (But they are currently testing Create, rather than Update, scenario.)

jgwest (approved):
LGTM, thanks @raghavi101!

You were right that there was something wrong with the 'managed-gitops-test-data/deployment-permutations-*' repo that I created (some of the ConfigMaps were invalid), so I fixed that and made sure the tests were up to date with it, plus some misc refactoring.

PR #159, opened 2022-06-24
Create a new token secret for service account if the default secret is absent

chetan-rns:
#### Description:
- From k8s 1.24 onwards service accounts will not get a default token secret. Instead, we need to create a token secret of type `kubernetes.io/service-account-token` and point it to our service account. 
- Add unit test to check the above behavior

https://issues.redhat.com/browse/GITOPSRVCE-171


jgwest (approved):

PR #162, opened 2022-06-27
Add GitOpsDeployments details in binding for components.

jparsai:


appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -48,20 +49,16 @@ type ApplicationSnapshotEnvironmentBindingReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *ApplicationSnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
 
-	fmt.Println("ApplicationSnapshotEventBinding event: ", req)
+	log := log.FromContext(ctx).WithValues("name", req.Name, "namespace", req.Namespace)
+	defer log.V(sharedutil.LogLevel_Debug).Info("Application Reconcile() complete.")
```
drpaneas, line 54: I think "Application Reconcile()" is too generic, it could be an ArgoCD application for example, but from what I understand the purpose of this log message is to signal the end of the ApplicationSnapshotEnvironmentBindingReconciler loop, right? If so, why not use this name instead?

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -48,20 +49,16 @@ type ApplicationSnapshotEnvironmentBindingReconciler struct {
...
-
-// Proof-of-concept Reconcile. Comment out the above Reconcile, and rename this to Reconcile, when starting working on this.
-func (r *ApplicationSnapshotEnvironmentBindingReconciler) ReconcilePOC(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	binding := &appstudioshared.ApplicationSnapshotEnvironmentBinding{}
 
 	if err := r.Client.Get(ctx, req.NamespacedName, binding); err != nil {
 		// Binding doesn't exist: it was deleted.
 		// Owner refs will ensure the GitOpsDeployments are deleted, so no work to do.
+		log.Error(err, "Error occurred while retrieving Binding for '"+req.NamespacedName.String()+"'")
 		return ctrl.Result{}, nil
```
drpaneas, line 62: Not code you wrote, but part of the block you are touching. If the error is not `nil` then I guess the return here should be `return ctrl.Result{}, err` instead of `return ctrl.Result{}, nil`. Also, given the tests didn't catch this, try adding a test if it's not too much of work.

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
jparsai: Sorry, my bad, I missed it. It should be Application Snapshot Environment Binding Reconciler()

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -126,33 +126,37 @@ func (r *ApplicationSnapshotEnvironmentBindingReconciler) ReconcilePOC(ctx conte
...
 		// B) The GitOpsDeployment is exactly as expected, so return
+		log.V(sharedutil.LogLevel_Debug).Info("GitOpsDeployment already exists.")
 		return nil
 	}
 
 	// C) The GitOpsDeployment should be updated to be consistent with what we expect
 	actualGitOpsDeployment.Spec = expectedGitopsDeployment.Spec
 
 	if err := k8sClient.Update(ctx, &actualGitOpsDeployment); err != nil {
+		log.Error(err, "unable to update "+actualGitOpsDeployment.Name)
```
drpaneas, line 159: I would also add the kind of the CR you are trying to update, .e.g `unable to update actualGitOpsDeployment " + actualGitOpsDeployment.Name` because in the code it's obvious but not in the logs.

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -92,18 +89,19 @@ func (r *ApplicationSnapshotEnvironmentBindingReconciler) ReconcilePOC(ctx conte
 	var firstErr error
 	// For each deployment, check if it exists, and if it has the expected content.
 	// - If not, create/update it.
-	for _, expectedGitOpsDeployment := range expectedDeployments {
+	for i, expectedGitOpsDeployment := range expectedDeployments {
 
 		if err := processExpectedGitOpsDeployment(ctx, expectedGitOpsDeployment, *binding, r.Client); err != nil {
 
 			if firstErr != nil {
+				log.Error(err, "Error occurred while processing expected GitOpsDeployment for Binding '"+binding.Name)
```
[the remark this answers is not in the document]
jgwest: Agreed @drpaneas @jparsai , as discussed in standup, my original proposal isn't great. What about something like this, but open to better suggestions on how to combine a bunch of errors: 
```go
	var allErrors error
	// var firstErr error
	// For each deployment, check if it exists, and if it has the expected content.
	// - If not, create/update it.
	for _, expectedGitOpsDeployment := range expectedDeployments {

		if err := processExpectedGitOpsDeployment(ctx, expectedGitOpsDeployment, *binding, r.Client); err != nil {

			log.Log.Error(err, "unable to process expected gitops deployment")

			if allErrors == nil {
				allErrors = err
			} else {
				allErrors = fmt.Errorf("%v. error: %v", err, allErrors)
			}

		} else {
			// No error: add to status
			/* (...) */
		}
	}

	/* (...) */

	if allErrors != nil {
		return ctrl.Result{}, fmt.Errorf("unable to process expected GitOpsDeployment: %v", firstErr)
	}
```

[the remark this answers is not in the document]
drpaneas: If you want to accumulate (combine a bunch of) errors you can use [error wrapping ](https://go.dev/doc/go1.13#error_wrapping)directly as of Go 1.13. See: https://stackoverflow.com/questions/33470649/combine-multiple-error-strings

Alternatively you can use a `[]string` and store the error string using the `err.Error()` (which return the string msg) and combine them (e.g. `allErrorsStr := strings.Join(allErrors, "\n")`) them using and then check the `len()` of this variable you can check if you had at least one.

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
jparsai: Returning error instead of nil, I also tried to test this using E2E, but it cant be, because Reconcile will throw this error if Binding CR is not found and in E2E if Binding is deleted then Reconcile wont be triggered to throw the error.

samyak-jn (approved):
/lgtm

tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
```diff
@@ -0,0 +1,384 @@
...
+
+			expectedGitOpsDeployments := []appstudiosharedv1.BindingStatusGitOpsDeployment{
+				{ComponentName: binding.Spec.Components[0].Name, GitOpsDeployment: gitOpsDeploymentNameFirst},
+				{ComponentName: binding.Spec.Components[1].Name, GitOpsDeployment: gitOpsDeploymentNameSecond},
+			}
+
+			Eventually(binding, "2m", "1s").Should(bindingFixture.HaveStatusGitOpsDeployments(expectedGitOpsDeployments))
+
+			//====================================================
+			By("Verify that GitOpsDeployment CR created by GitOps-Service is having metadata as given in Binding.")
```
jgwest, line 58: Think this can be 'is having source spec' or 'is having spec source' as given in Binding, likewise in other places where the term metadata is used. This is because it's the .spec.source field of the GitOpsDeployment that the test is actually waiting for.

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+			// Trigger Reconciler
+			_, err = bindingReconciler.Reconcile(ctx, request)
+			Expect(err).To(BeNil())
+
+			// Check status field after calling Reconciler
+			bindingSecond := &appstudiosharedv1.ApplicationSnapshotEnvironmentBinding{}
+			err = bindingReconciler.Client.Get(ctx, request.NamespacedName, bindingSecond)
+
+			Expect(err).To(BeNil())
```
jgwest, line 117: For a couple of these tests, can you verify that the `.metadata.ownerReference` field of the GitOpsDeployment resources are equal to the Binding? That way, we can be sure the `GitOpsDeployment` will be deleted by k8s once the `OwnerReference` is.

tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
```diff
@@ -0,0 +1,384 @@
...
+	appstudiosharedv1 "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend/apis/managed-gitops/v1alpha1"
+	bindingFixture "github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture/binding"
+	gitopsDeplFixture "github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture/gitopsdeployment"
+
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("ApplicationSnapshotEnvironmentBinding Reconciler E2E tests", func() {
```
jgwest, line 21: For a couple of these tests, can you verify that the `.metadata.ownerReference` field of the GitOpsDeployment resources are equal to the Binding? That way, we can be sure the `GitOpsDeployment` will be deleted by k8s once the `OwnerReference` is.

jgwest (changes requested):
Great stuff, @jparsai , and thanks for your patience in how long it took to review. Just some very minor tweaks then good to go.

appstudio-controller/controllers/appstudio.redhat.com/applicationsnapshotenvironmentbinding_controller.go
```diff
@@ -88,22 +86,26 @@ func (r *ApplicationSnapshotEnvironmentBindingReconciler) ReconcilePOC(ctx conte
...
+			if allErrors == nil {
+				allErrors = fmt.Errorf("%s Error: %w", errorMessage, err)
+			} else {
+				allErrors = fmt.Errorf("%s\n%s Error: %w", allErrors.Error(), errorMessage, err)
 			}
 		} else {
 			// No error: add to status
 			statusField = append(statusField, appstudioshared.BindingStatusGitOpsDeployment{
-				ComponentName:    "", // GITOPSRVCE-156: TODO: use the actual component name
+				ComponentName:    binding.Spec.Components[i].Name,
```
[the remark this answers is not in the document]
jparsai: Yes I agree, earlier `Components` field was inside `Spec` but later it was moved into `Status`, I was supposed to change it from `binding.Spec.Components[i]` to `binding.Status.Components[i]`, but I missed it.

jgwest:
Looks like the E2E test fails because [the timeout we are using is too short](https://github.com/openshift/release/blob/f65abe36b7a5f1563b08744687ee815912cd0f28/ci-operator/config/redhat-appstudio/managed-gitops/redhat-appstudio-managed-gitops-main.yaml#L56). I will follow up with Samyak on this, and merge in the mean time.

PR #163, opened 2022-06-27
Update manifests: update health ports, and add zap logging parameter

jgwest:
#### Description:
- Update health ports so they are consistent with what port the controller is actually listening on, to fix health check issues.
- Add zap time encoding parameter, to improve the time reporting format of the controller when running on staging cluster


drpaneas:
what where the health issues?

/lgtm

samyak-jn (approved):

jgwest:
@drpaneas when running on staging cluster, the health/readiness checks were occasionally failing due to the ports being wrong.

PR #166, opened 2022-06-29
Update make targets to use k8s configurations instead of docker while setting up on openshift

samyak-jn:
Hello folks :hugs:,

#### Description:
- Update make targets to use k8s configurations instead of docker while setting up on openshift (required by openshift-ci)!


samyak-jn:
Hi team,

I'm merging this PR as per our conversation from yesterday's standup. I have verified that the tests are passing on both the GitHub actions and ci is running as expected. The changes are isolated and don't affect any piece of functionality.

PR #169, opened 2022-07-04
Delete orphaned ArgoCD Apps in Namespace Reconciler.

jparsai:
#### Description:
This PR is to introduce a logic in Namespace Reconciler to identify orphaned ArgoCD applications and delete them.

#### Link to JIRA Story:
https://issues.redhat.com/browse/GITOPSRVCE-170

#### Type of changes

1. Moved Namespace Reconciler and related functions into a new file `namespace_reconciler.go`
2. Modified existing Namespace Reconciler function to introduce function to to identify and delete orphaned ArgoCD applications.
3. Added unit test cases.


cluster-agent/controllers/argoproj.io/namespace_reconciler.go
```diff
@@ -0,0 +1,335 @@
...
+
+	// Iterate through all application IDs collected from DB.
+	for id := range processedApplicationIds {
+		// Iterate through ArgoCD all and remove corresponding application from list.
+		for i, application := range argoApplications {
+			if application.Labels["databaseID"] == id {
+				argoApplications = dropProcessedApplicationId(argoApplications, i)
+			}
+		}
+	}
```
chetan-rns, line 313: Since we are using a map for `processedApplicationIds`, can we avoid iterating both the loops?

For example:

```
for _, application := range argoApplications {
    if _, ok := processedApplicationIds[application.Labels["databaseID"]]; !ok {
	    if err := controllers.DeleteArgoCDApplication(ctx, application, client, log); err != nil {
		    log.Error(err, "unable to delete an orphaned ArgoCD application "+application.Name)
	    } else {
		    log.Info("Deleted orphaned ArgoCD application " + application.Name)
	    }
    }
}
```
Also the map `processedApplicationIds` might be empty and we would still remove the orphaned applications.

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
```diff
@@ -0,0 +1,335 @@
...
+	return argoApplications
+}
+
+func dropProcessedApplicationId(argoApplications []appv1.Application, i int) []appv1.Application {
+	argoApplications[i] = argoApplications[len(argoApplications)-1]
+	return argoApplications[:len(argoApplications)-1]
+}
+
+func deleteApplication(ctx context.Context, client client.Client, argoApplications []appv1.Application, log logr.Logger) {
+	for _, application := range argoApplications {
```
chetan-rns, line 328: I wonder if we could directly delete the application in the previous function without looping through them again?

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
```diff
@@ -0,0 +1,335 @@
...
+		// This ensures that at least 'namespaceReconcilerInterval' time elapses from the end of one rum to the beginning of another.
+		r.startTimerForNextCycle()
+	}()
+
+}
+
+func runNamespaceReconcile(ctx context.Context, dbQueries db.DatabaseQueries, client client.Client, log logr.Logger) {
+
+	// Fetch list of ArgoCD applications to be used later
+	processedApplicationIds := make(map[string]string)
```
chetan-rns, line 61: Nit: Since we don't use the value of this map anywhere, can we change the type to `bool`? 
```suggestion
	processedApplicationIds := make(map[string]bool)
```

jparsai: Yes It can be anything, values of map are is not useful in current situation, so it can be of any type.
Just for curiosity, is there any reason for choosing bool?

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
jparsai: Can you elaborate a bit more, please?
In previous function I am removing all those applications from list, which are having entry in DB. As result of this all orphaned applications remaining in list, so in this function I am going through that list in deleting applications from ArgoCD one by one.

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
chetan-rns: Instead of removing elements from the array and then deleting the remaining ones from DB inside `deleteApplication()`, what if we directly delete them from the DB inside the loop? Then we don't have to update the array right? The above example tries to do something similar

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
chetan-rns: Same as [above](https://github.com/redhat-appstudio/managed-gitops/pull/169#discussion_r919839837). If we delete the orphaned applications from the DB in the previous function itself we don't have to call  `deleteApplication()` and loop again.

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
chetan-rns: I agree in this case it doesn't matter much. I prefer using bool if we are only checking if a key exists and don't care about the value. So I suggested it as a Nit. Please feel free to ignore

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
jparsai: My bad, I misunderstood here. Changed the logic as you suggested. PTAL.

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
jgwest: FWIW I tend to do something like this: an `interface{}` for the value, then a comment explaining the use:
 ```
 // map: applications IDs seen (string) -> (map value not used)
processedApplicationIDs := make(map[string]interface{})
 ```
 But what y'all have here is fine. In Java we have `HashSet`, but no built-in equivalent in Go AFAIK :stuck_out_tongue:

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
```diff
@@ -0,0 +1,321 @@
...
+
+	// Start a goroutine, because DeleteArgoCDApplication() function from cluster-agent/controllers may take some time to delete application.
+	go deleteOrphanedApplications(argoApplications, processedApplicationIds, ctx, client, log)
+
+	log.Info(fmt.Sprintf("Namespace Reconciler finished an iteration at %s. "+
+		"Next iteration will be triggered after %v Minutes", time.Now().String(), namespaceReconcilerInterval))
+}
+
+// compareApplications compares Application objects, since both objects are of different types we can not use == operator for comparison.
+func compareApplications(applicationFromArgoCD appv1.Application, applicationFromDB fauxargocd.FauxApplication, log logr.Logger) bool {
```
drpaneas, line 192: I had similar implementation but then as @jgwest noticed, this logic will issue an update to the object per different field. In other words, if there are e.g. 3 differences, it will correct the object 3 times, aka it will send 3 API calls. Ideally we would like to minimize the API calls. So if we can send one call and update it all at once it would be preferable. 

Take a look at my implementation for a very similar function: https://github.com/redhat-appstudio/managed-gitops/pull/168/files#diff-a387a638ba7c6a4d686c64d88c5b9c9a28a7b289ed665fed4baed16c99c6ab37R246-R250

jparsai: Yes, there is no harm in doing this, but here it is not exactly same case. In compareApplications() if any of the field is different, it will return from there only and will not compare any other field. After this a new Operation will be created and all out of sync fields will be updated at once. Means if there are 3 different fields still only 1 Operation will be created, which will update all 3 fields in by making 1 API call.
https://github.com/redhat-appstudio/managed-gitops/blob/f4121710a35ef41d156da68bbb54c22748bc6cd5/cluster-agent/controllers/argoproj.io/namespace_reconciler.go#L152

drpaneas: O I've got it now ;) One small caveat in that case you are printing only the first log. So if there are more than one diffs you miss the rest.

drpaneas (approved):
/lgtm

Thanks @jparsai

PR #171, opened 2022-07-05
Add UNIQUE contraint to explicitly prevent duplicates on tables that don't support/want them

jgwest:
#### Description:
- Based on a couple of Slack conversations where folks were hitting issues with duplicate values in the database where they weren't expected (mainly around APICRtoDatabaseMapping).
- Since we have the power of a relational database at our disposable, we can use the `UNIQUE` constraint to prevent duplicates, thus preventing race conditions in these fields.
- I've added UNIQUE constraints to a few of the tables, in order to express the existing requirements around a lack of duplicates.


chetan-rns:
I couldn't reproduce the issue. But the constraints look good to me. Should we handle the error received when this constraint rejects a duplicate entry? 

For example, do we get an existing object and continue when we receive a duplicate entry error?

https://github.com/redhat-appstudio/managed-gitops/blob/45b53e5f01fe8f441a20b450b266172dd628a431/backend/eventloop/application_event_loop/application_event_runner_syncruns.go#L243

jgwest:
No review comments, merging.

PR #173, opened 2022-07-06
Fix argocd installation script to use the correct namespace and port

chetan-rns:
#### Description:
- Argo CD namespace should be read from the variable `ARGO_CD_NAMESPACE`
- The port used for port-forwarding is already used by backend metrics. Use a different port number 4000


jparsai (approved):

drpaneas (approved):
/lgtm

jgwest (approved):
:+1:, thanks @chetan-rns

PR #176, opened 2022-07-07
Avoid creating duplicate Operations.

jparsai:
#### Description:
This PR is to add the logic to avoid creating Operation for a resource, if it is already present in cluster, but waiting to be completed.

#### JIRA Ticket
https://issues.redhat.com/browse/GITOPSRVCE-176

#### Type of changes

-  Modified CreateOperation() function to return existing operation if an In Progress Operation is already available in cluster.
- Added unit test cases.

### DIFF
--- a/backend/eventloop/eventlooptypes/types.go
+++ b/backend/eventloop/eventlooptypes/types.go
@@ -13,7 +13,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	sharedv1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
+	operation "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
 	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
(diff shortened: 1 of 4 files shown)


backend/eventloop/eventlooptypes/types.go
```diff
@@ -13,7 +13,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	operation "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
+	managedgitopsbackendsharedv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
```
drpaneas, line 16: this is too long and as a result later when you use it the lines tend to get longer as well. You can just pick another shorter name for this (eg. `shared`) instead of `managedgitopsbackendsharedv1alpha1`

drpaneas (commented):
Code looks good and testing also good. I have only one question related to the name of the operation. Apart from that, it's good to go. :D

jgwest (approved):
LGTM, thanks @jparsai ! I realized that I had some draft review comments in the GitHub Web UI that I forgot to hit 'Submit' on (from last week), so rather than delays things further I added the changes as a new commit. Feel free to let me know if anything seems incorrect.

PR #181, opened 2022-07-20
Add CI checks for duplicate pkg

drpaneas:
#### Description:
- Add check for duplicate packages

No

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -34,7 +34,7 @@ jobs:
           go-version-file: './backend/go.mod'
       - name: "Install dupimport"
         run: |
-          go install github.com/gostaticanalysis/dupimport/cmd/dupimport
+          go install github.com/gostaticanalysis/dupimport/cmd/dupimport@latest
       - name: "Ensure 'backend' has no duplicate imports"
         run: |
           cd $GITHUB_WORKSPACE/backend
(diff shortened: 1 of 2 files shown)


samyak-jn (commented):
/lgtm

.github/workflows/static.yml
```diff
@@ -22,6 +22,39 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Setup Golang
+        uses: actions/setup-go@v3.2.0
+        with:
+          go-version-file: './backend/go.mod'
+      - name: "Install dupimport"
+        run: |
+          go install github.com/gostaticanalysis/dupimport/cmd/dupimport@latest
```
jgwest, line 37: Rather than `@latest`, can we use `@37d4ab484168ec60101dcc9bf644b54f910c2a25` (e.g. the most recent commit) to prevent new commits from breaking or otherwise subverting our build/test jobs?

jgwest (changes requested):
LGTM, one small tweak then good to go.

.github/workflows/static.yml
drpaneas: Done

jgwest (approved):
LGTM, thanks @drpaneas for jumping on figuring out how to do this! It's always great when we can get GitHub's PR-checking robots to do our checking for us :smile:

PR #186, opened 2022-07-28
Prevent deletion of Argo CD Applications that are missing 'databaseID' field, unit tests for DeleteArgoCDApplication, fix issue with '/' in database secret

jgwest:
#### Description:
- Fix some issue that were causing problems on staging cluster:
- Fix issue with '/' in database secret
- Prevent deletion of Argo CD Applications that are missing 'databaseID' field
    - Unit tests for DeleteArgoCDApplication


jgwest:
Thanks @jparsai!

PR #187, opened 2022-07-30
Use go-pg's shared connection pool logic and update GitOps Service k8s manifests

jgwest:
#### Description:
- Fixing some issues seen on the AppStudio staging cluster:
    - A 'too many connections' errors from PostgreSQL: fortunately we can use go-pg in-built connection sharing to solve this
    - Appstudio-controller is OOM-ing, so we can increase the memory limits
    - During stress unit tests, Argo CD is OOMing, so increase the memory limit of that
- Switch to using go-pg's existing connection pooling logic: one connection pool for 'verbose' logging, and one for without verbose logging
- Unit tests for above.
- Increase memory limits of appstudio-controller `Deployment` resource

### DIFF
--- a/backend-shared/config/db/queries.go
+++ b/backend-shared/config/db/queries.go
@@ -255,10 +255,11 @@ const (
 	sharedDBPoolMapKey_verbose  = "verbose"
 )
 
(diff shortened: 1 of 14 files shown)


jgwest:
Merging w/o reviews to fix current issues with appstudio-staging cluster.

PR #189, opened 2022-08-04
Update backend/appstudio-controller K8s manifests

jgwest:
#### Description:
- Increase the liveness/readiness probe duration of backend
- Increase the memory limit of appstudio-controller


jgwest:
Merging as manifest updates only

PR #195, opened 2022-08-09
Move application info cache to cluster-agent, and cleanup code. Update DeleteArgoCDApplication logic to avoid Error log

jgwest:
#### Description:
- Update Argo CD Application resource deletion logic to wait 2 minutes from the _current time_ for the application to delete, rather than from the deletion timestamp.
- Move application info cache to cluster-agent, and cleanup code
- Update DeleteArgoCDApplication logic to avoid Warning log

### DIFF
--- a/cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
+++ b/cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
@@ -21,7 +21,7 @@ import (
 //
 // The cache:
 // - will always return the most recent ApplicationState value from the database
-// - in constrast, the cache will return a value for Application that is at most 60 seconds old
+// - in contrast, the cache will return a value for an Application that is at most 60 seconds old
 //   (the Application in the cache state will be eventually consistent with the database)
 //     - since it is eventually consistent, the calling code needs to be aware of this in its logic.
 
--- a/cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
+++ b/cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
@@ -1,4 +1,4 @@
-package util
+package application_info_cache
 
 import (
 	"context"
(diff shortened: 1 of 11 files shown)


cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
```diff
@@ -11,6 +11,20 @@ import (
...
+// of the time these reconciliations are changes that might not be useful for us.
+//
+// To avoid unnecessary read/write to the database, the cache is being implemented in backend-shared which will be used
+// by cluster-agent's Application contrller reconciliation.
+//
+// The cache should only be used by 'application_controller.go'.
+//
+// The cache:
+// - will always return the most recent ApplicationState value from the database
+// - in constrast, the cache will return a value for Application that is at most 60 seconds old
```
samyak-jn, line 24:
```suggestion
// - in contrast, the cache will return a value for an Application that is at most 60 seconds old
```

samyak-jn (approved):
/lgtm

The changes are very neat, and of course, making our codebase more cleaner and easier to navigate makes more sense! Thanks, @jgwest :rocket:

cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
```diff
@@ -11,6 +11,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+// The cache exists because we do want to overwhelm the controller (or the database) with various operations frequently: most
```
drpaneas, line 14:
```suggestion
// The cache exists because we do not want to overwhelm the controller (or the database) with various operations frequently: most
```

cluster-agent/controllers/util.go
```diff
@@ -133,13 +137,13 @@ func DeleteArgoCDApplication(ctx context.Context, appFromList appv1.Application,
 
 		backoff.Reset()
 
-		// Add up to X minutes (eg 2 minutes) to wait for the deletion
-		expirationTime = expirationTime.Add(WaitForArgoCDToPerformNonfinalizerDeleteTimeout)
+		// Wait 2 minutes from the current time for the application to delete, before reporting an error.
```
drpaneas, line 140: I'm sure you have a good reason. Can you share some information why this is needed?

drpaneas (changes requested):

drpaneas:
The only reason I requested changes is for a comment that is missing (I think) a `not`, thus is changing the whole meaning. Apart from that good to go. I would be great if you can answer the other two question I have as well (no blockers).

cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache.go
jgwest: :laughing: , whoops, thanks for catching this.

cluster-agent/controllers/argoproj.io/application_info_cache/application_info_cache_test.go
```diff
@@ -18,7 +17,7 @@ var _ = Describe("application_info_cache Test", func() {
 			Expect(err).To(BeNil())
 
 			ctx := context.Background()
-			asc := util.NewApplicationInfoCache()
+			asc := NewApplicationInfoCache()
```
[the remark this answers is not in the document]
jgwest: @drpaneas ah, I think ApplicationInfoCache used to be called ApplicationSharedCache. I have updated the variable name in all the test functions.

cluster-agent/controllers/util.go
jgwest: @drpaneas, good question. Once the finalizer is removed from the Application, Kubernetes should naturally clean up (delete) the Application, so we wait for it to do so.

drpaneas (approved):
/lgtm

PR #199, opened 2022-08-10
Expose ComparedTo field of Argo CD Application via GitOpsDeploymentStatus

Rizwana777:
#### Description:
- This PR is about exposing comparedTo field of ArgoCD Application via GitopsDeploymentStatus field
- Added new field in database called "reconciled_state"
- Added new field in migration list
- Updated logic in cluster-agent to store reconciledState fields in database.
- Added logic in backend to retrieve reconciledState from database and assign it to gitopsDeploymentStatus.ReconciledState fields.

https://issues.redhat.com/browse/GITOPSRVCE-181)]


Rizwana777:
@jgwest  PTAL and please correct me if I have done migrations wrong, Thank you

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -734,6 +734,20 @@ func (a *applicationEventLoopRunner_Action) applicationEventRunner_handleUpdateD
...
+	if err != nil {
+		a.log.Error(err, "unable to retrieve comparedTo field in ApplicationState")
+		return err
+	}
+
+	// Update gitopsDeployment status with reconciledState
+	gitopsDeployment.Status.ReconciledState.Source.Path = comparedTo.Source.Path
+	gitopsDeployment.Status.ReconciledState.Source.RepoURL = comparedTo.Source.RepoURL
+	gitopsDeployment.Status.ReconciledState.Source.Branch = comparedTo.Source.TargetRevision
+	gitopsDeployment.Status.ReconciledState.Destination.Name = comparedTo.Destination.Name
```
jgwest, line 748: When I try out the PR, and I view the `reconciledState` in the `GitOpsDeployment` resource, it looks like this:

    reconciledState:
      destination:
        name: 5c59aecf-6ae1-4801-b52a-4641f464e7ef <--- the managedenvironment_id of the ManagedEnviroment row
        namespace: jane
      source:
        branch: ""
        path: environments/overlays/dev
        repoURL: ""


Rather than displaying the database primary key for the user (we generally want to keep our database internals hidden), we should instead convert this "5c59aecf-(...)" value into a reference to the name of the corresponding GitOpsDeploymentManagedEnvironment CR in the environment. (This is included as part of the story criteria).
- For example, let's say we have a GitOpsDeploymentManagedEnvironment name `my-env`, defined in the namespace.
- That GDManagedEnvironment resource corresponds to the `5c59aecf-6ae1-4801-b52a-4641f464e7ef` ManagedEnvironment row.
- Therefore, in .reconciledState.destination.name, we should use the value `my-env`

In order to get the name of the GitOpsDeploymentManagedEnvironment resource corresponding to the ManagedEnvironment table, we can use the `APICRToDatabaseMapping` table. It is the responsibility of this table to maintain the relationship from K8s API resources (like GitOpsDeploymentManagedEnvironment) <-> corresponding database resources (row in ManagedEnvironment table).

I checked, and it looks like we don't have an existing database query function for this, so I wrote one (which hopefully works, I've not tested it).

You can add this to apicrtodatabasemapping.go (you'll also need to add it to the `ApplicationScopedQueries` interface in `queries.go`)

```go
// GetAPICRForDatabaseUID retrieves the name/namespace/uid of an API Resources (such as GitOpsDeploymentManagedEnvironment) based on the
// primary key of the corresponding database row (for example, ManagedEnvironment)
func (dbq *PostgreSQLDatabaseQueries) GetAPICRForDatabaseUID(ctx context.Context, obj *APICRToDatabaseMapping) error {

	if err := validateQueryParamsEntity(obj, dbq); err != nil {
		return err
	}

	if err := isEmptyValues("GetAPICRForDatabaseUID",
		"APIResourceType", obj.APIResourceType,
		"DBRelationType", obj.DBRelationType,
		"DBRelationKey", obj.DBRelationKey); err != nil {
		return err
	}

	var result []APICRToDatabaseMapping

	if err := dbq.dbConnection.Model(&result).
		// TODO: GITOPSRVCE-68 - PERF - Add a DB index for this
		Where("atdbm.api_resource_type = ?", obj.APIResourceType).
		Where("atdbm.db_relation_type = ?", obj.DBRelationType).
		Where("atdbm.db_relation_key = ?", obj.DBRelationKey).
		Context(ctx).
		Select(); err != nil {

		return fmt.Errorf("error on retrieving database mapping for APICRToDatabase: %v", err)
	}

	if len(result) == 0 {
		return NewResultNotFoundError(fmt.Sprintf("unable to retrieve APICRToDatabase mapping for %s:%s",
			obj.APIResourceType, obj.DBRelationKey))
	}

	if len(result) > 1 {
		return fmt.Errorf("unexpected number of results when retrieving APICRToDatabase mapping for %s:%s",
			obj.APIResourceType, obj.DBRelationKey)
	}

	*obj = result[0]

	return nil

}
``

Then, in backend, when you read the value of the managedEnvironment ID from the ReconciledState field of the ApplicationState database table, you can convert the value of this field to the corresponding 
- You can use the above function to translate the database ID -> CR name
- If no match is found (for example, because the managedenvironment was deleted), just leave the field empty (rather than returning an error)

File: utilities/db-migration/migrations/000007_v7.down.sql [hunk 2]
```diff
@@ -0,0 +1 @@
+ALTER TABLE ApplicationState DROP COLUMN reconciled_state;
```
jgwest, line 1: :+1:

utilities/db-migration/migrations/000007_v7.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE ApplicationState ADD COLUMN reconciled_state VARCHAR ( 4096 );
```
jgwest, line 1: :+1:

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -411,6 +411,77 @@ var _ = Describe("Application Controller", func() {
 			Expect(result).ToNot(BeNil())
 
 		})
+
+		It("Test to check whether reconciled state is stored in database", func() {
```
jgwest, line 415: FYI the unit tests appear to be failing in the CI job, in these functions.

backend-shared/apis/managed-gitops/v1alpha1/gitopsdeployment_types.go
```diff
@@ -75,6 +75,26 @@ type ResourceStatus struct {
...
+	// Path contains path from .status.Sync.CompareTo field of ArgoCD Application
+	Path    string `json:"path"`
+	RepoURL string `json:"repoURL"`
+	Branch  string `json:"branch"`
+}
+
+// GitOpsDeploymentDestination contains the information of .status.Sync.CompareTo.Destination field of ArgoCD Application
+type GitOpsDeploymentDestination struct {
+	Name      string `json:"name"`
+	NameSpace string `json:"namespace"`
```
jgwest, line 95: The name of this field should be `Namespace`, rather than `NameSpace`

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -411,6 +411,77 @@ var _ = Describe("Application Controller", func() {
...
+			By("Call reconcile function")
+			result, err := reconciler.Reconcile(ctx, newRequest(namespace, name))
+			Expect(err).To(BeNil())
+			Expect(result).ToNot(BeNil())
+
+			err = reconciler.DB.GetApplicationStateById(ctx, &applicationStateDB)
+			Expect(err).ToNot(BeNil())
+			By("Verify ReconciledState is stored in db")
+			Expect(applicationStateDB.ReconciledState).ToNot(BeNil())
+
```
jgwest, line 483: I think we should have a second unit test that ensures that when:
- Argo CD Application already exists
- Application row already exists
- ApplicationState row already exists
- ApplicatoinState already has a non-empty value in ReconciledState

Test that when:
- the Argo CD Application comparedTo field changes to a new value
That:
- The existing value in ReconciledState is updated to the new value.

cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -242,3 +266,52 @@ func compressResourceData(resources []appv1.ResourceStatus) ([]byte, error) {
...
+		comparedTo.Destination.Name = managedEnvID
+	}
+
+	// 4) Convert the fauxArgoCD object into JSON
+	comparedToBytes, err := json.Marshal(comparedTo)
+	if err != nil {
+		return "", fmt.Errorf("SEVERE: unable to convert comparedTo to JSON")
+	}
+
+	// 5) Finally, store the JSON in the database.
```
jgwest, line 313: Nit pick: `// 5) Finally, store the JSON in the database.` -> `// 5) Finally, store the JSON in the database object.`

jgwest (changes requested):
Hey @Rizwana777, the logic and migration code looks great! I've add some comments below re: updating the value of the `name` field in the GitOpsDeployment resource's ReconciledState field. Basically, we should hide the database ID, and instead use the name of the GitOpsDeploymentManagedEnvironment for thie field. Hopefully this should make those story criteria more clear.

I've also requested an additional unit test, see below.

Finally, the story criteria include E2E tests, which I didn't see as present in the PR (correct me if I am wrong!).

I think there are a couple things we can test for in the E2E tests, for this story. See 'A:' and 'B:' below.

## A: Test that creating/updating a local GitOpsDeployment causes the reconciledState field to be updated.

1) Create a GitOpsDeployment with an empty destination field, like this:
```yaml
apiVersion: managed-gitops.redhat.com/v1alpha1
kind: GitOpsDeployment
metadata:
  name: gitops-depl
  namespace: (e2e namespace)
spec:
  source:
    repoURL: https://github.com/redhat-appstudio/gitops-repository-template
    path: environments/overlays/dev

#  destination: {}

  type: automated
```

2) Ensure that once created, the .status.reconciledState field eventually contains the correct values the equal to the `spec` field defined above.

3) Next, update a fields in `source:`.

4) The .status.reconciledState field should eventually match the updated value you set in step 3.

We already have E2E tests that create, update, and delete a GitOpsDeployment in gitopsDeployment_test.go. You should be able to update the existing tests to include the behaviour described above.


## B) Test that creating/updating a GitOpsDeployment (but that targets a managed environment), causes the reconciledState field to be updated.

The same steps as A), but using a GitOpsDeployment that targets a managed environment. 

For this, you should be able to add the reconciledState-checking logic into `managed_environment_test.go` (checking for reconciledState after create), rather than writing a new test for this.

backend-shared/util/argocd/util.go
```diff
@@ -1,12 +1,57 @@
...
+
+	// ArgoCDDefaultDestinationInCluster is 'in-cluster' which is the spec destination value that Argo CD recognizes
+	// as indicating that Argo CD should deply to the local cluster (the cluster that Argo CD is installed on).
+	ArgoCDDefaultDestinationInCluster = "in-cluster"
+)
 
 // GenerateArgoCDClusterSecretName generates the name of the Argo CD cluster secret (and the name of the server within Argo CD).
 func GenerateArgoCDClusterSecretName(managedEnv db.ManagedEnvironment) string {
-	return "managed-env-" + managedEnv.Managedenvironment_id
+	return "managedEnvPrefix" + managedEnv.Managedenvironment_id
```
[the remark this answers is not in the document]
Rizwana777: Yes , you are correct I will update this 😅

cluster-agent/controllers/argoproj.io/application_controller_test.go
Rizwana777: I will check this @jgwest I did some rebase and forgot top test it properly again

Rizwana777:
Thank you for the review @jgwest I will update the changes.

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -411,6 +411,77 @@ var _ = Describe("Application Controller", func() {
 			Expect(result).ToNot(BeNil())
 
 		})
+
+		It("Test to check whether reconciled state is stored in database", func() {
+			By("Close database connection")
+			defer dbQueries.CloseDatabase()
+			defer testTeardown()
```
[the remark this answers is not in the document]
Rizwana777:
```
err = db.SetupForTestingDBGinkgo()
			Expect(err).To(BeNil())
```
@jgwest  This is called in BeforeEach loop, so I have not added here.

cluster-agent/controllers/argoproj.io/application_controller.go
```diff
@@ -242,3 +266,52 @@ func compressResourceData(resources []appv1.ResourceStatus) ([]byte, error) {
 
 	return buffer.Bytes(), nil
 }
+
+// storeInComparedToFieldInApplicationState will read 'comparedTo' field of an Argo CD Application, and write the
+// correct value into 'applicationState'.
+func storeInComparedToFieldInApplicationState(argoCDAppParam appv1.Application, applicationState db.ApplicationState) (string, error) {
+
+	comparedToParam := argoCDAppParam.Status.Sync.ComparedTo
```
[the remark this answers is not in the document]
Rizwana777: yes @jgwest thank you for this

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -225,16 +242,25 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 			Expect(gitOpsDeploymentResource.Spec.Source.RepoURL).To(Equal(repoURL))
 			Expect(gitOpsDeploymentResource.Spec.Source.Path).To(Equal("environments/overlays/dev"))
 
+			expectedReconciledStateField := managedgitopsv1alpha1.ReconciledState{
+				Source: managedgitopsv1alpha1.GitOpsDeploymentSource{
+					RepoURL: gitOpsDeploymentResource.Spec.Source.RepoURL,
+					Path:    gitOpsDeploymentResource.Spec.Source.Path,
+				},
+			}
```
jgwest, line 250: Likewise, shouldn't the `Destination:` fields be defined here as well?

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -364,11 +390,20 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 			err = k8s.Get(&gitOpsDeploymentResource)
 			Expect(err).To(Succeed())
 
+
+			expectedReconciledStateField := managedgitopsv1alpha1.ReconciledState{
+				Source: managedgitopsv1alpha1.GitOpsDeploymentSource{
+					RepoURL: gitOpsDeploymentResource.Spec.Source.RepoURL,
+					Path:    gitOpsDeploymentResource.Spec.Source.Path,
+				},
+			}
```
jgwest, line 399: Destination fields here as well.

tests-e2e/core/managed_environment_test.go
```diff
@@ -79,12 +79,20 @@ var _ = Describe("GitOpsDeployment Managed Environment E2E tests", func() {
 			// 	return false
 			// }, "60s", "1s").Should(BeTrue())
 
-			By("ensuring GitOpsDeployment should have expected health and status")
+			By("ensuring GitOpsDeployment should have expected health and status and reconciledState")
+
+			expectedReconciledStateField := managedgitopsv1alpha1.ReconciledState{
```
jgwest, line 84: I would likewise expect a Destination field defined here

backend/eventloop/application_event_loop/application_event_runner_test.go
```diff
@@ -778,13 +779,17 @@ var _ = Describe("ApplicationEventLoop Test", func() {
...
+			Expect(err).To(BeNil())
+
 			applicationState := &db.ApplicationState{
 				Applicationstate_application_id: deplToAppMapping.Application_id,
 				Health:                          string(managedgitopsv1alpha1.HeathStatusCodeHealthy),
 				Sync_Status:                     string(managedgitopsv1alpha1.SyncStatusCodeSynced),
 				Revision:                        "abcdefg",
 				Message:                         "Success",
 				Resources:                       buffer.Bytes(),
+				ReconciledState:                 reconciledStateString,
```
jgwest, line 792: Also in this test, we should check that the GitOpsDeployment's status field has the correct ReconciledState values, similar to how we are checking other status values.

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -736,6 +736,34 @@ func (a *applicationEventLoopRunner_Action) applicationEventRunner_handleUpdateD
...
+		apiCRToDBMapping := &db.APICRToDatabaseMapping{
+			APIResourceType: db.APICRToDatabaseMapping_ResourceType_GitOpsDeploymentManagedEnvironment,
+			DBRelationType:  db.APICRToDatabaseMapping_DBRelationType_ManagedEnvironment,
+			DBRelationKey:   comparedTo.Destination.Name,
+		}
+		err = dbQueries.GetAPICRForDatabaseUID(ctx, apiCRToDBMapping)
+		if err != nil {
+			comparedTo.Destination.Name = ""
+		} else {
+			comparedTo.Destination.Name = apiCRToDBMapping.APIResourceName
```
jgwest, line 756: It looks like this line is not currently covered by the unit tests. We should be able to create a new test, based on  `Should update correct status of deployment after calling DeploymentStatusTick handler.` test, which:
1) Creates a GitOpsDeploymentManagedEnvironment in the k8s fake client
    - See the existing test named `reconciles a GitOpsDeployment that references a managed environment, then delete the managed env and reconciles`, for an example of how to create a ManagedEnvironment from a unit test.
2) Creates a ManagedEnvironment database entry based on that GitOpsDeploymentManagedEnvironment
3) Creates a APICRToDatabaseMapping based, that connects the `GitOpsDeploymentManagedEnvironment` in step 1, with the `ManagedEnvironment` row in step 2.
4) Updates the applicationState.ReconciledState field to contain a fake comparedTo field from Argo CD, which points to the uid of the managed environment.
5) Calls `applicationEventRunner_handleUpdateDeploymentStatusTick`

Then, verifies that the .status.reconciledState value of the `GitOpsDeployment` resource correctly references the name of the GitOpsDeploymentManagedEnvironment resource.

jgwest (changes requested):
Hi @Rizwana777 , thanks for making the requested updates. I've got some further review comments related to unit tests and E2E tests (and also FYI the E2E tests are failing, and my guess is the review comments would help shed some light on that). Let me know what you think!

cluster-agent/controllers/argoproj.io/application_controller_test.go
```diff
@@ -405,11 +519,26 @@ var _ = Describe("Application Controller", func() {
...
+			applicationStateget := &db.ApplicationState{
+				Applicationstate_application_id: applicationDB.Application_id,
+			}
+
+			By("Verify ReconcidedState in ApplicationState is updated in database")
+			err = reconciler.DB.GetApplicationStateById(ctx, applicationStateget)
+			Expect(err).To(BeNil())
+
+			By("Checking for sync status to verify whether Application CR and database is in sync with each other and it indicates that reconciledState is updated")
+			Expect(applicationStateget.Sync_Status).To(Equal("Synced"))
```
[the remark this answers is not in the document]
Rizwana777: If Application is in sync with database so that reconciledStated is also in sync, I thought like this, please correct me if I am wrong

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -185,6 +185,23 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 					gitopsDeplFixture.HaveResources(expectedResourceStatusList),
 				),
 			)
+
+			expectedReconciledStateField := managedgitopsv1alpha1.ReconciledState{
+				Source: managedgitopsv1alpha1.GitOpsDeploymentSource{
```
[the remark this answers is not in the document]
Rizwana777: As you mentioned in the comment of e2e tests, `Create a GitOpsDeployment with an empty destination field, like this:` I didn't added destination field. please correct me if I have taken this wrong.

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -394,6 +429,7 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
 					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
 					gitopsDeplFixture.HaveResources(expectedResourceStatusList),
+					gitopsDeplFixture.HaveReconciledState(expectedReconciledStateField),
```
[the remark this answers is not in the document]
Rizwana777: yes, I will update this

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -185,6 +185,23 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
...
+			expectedReconciledStateField := managedgitopsv1alpha1.ReconciledState{
+				Source: managedgitopsv1alpha1.GitOpsDeploymentSource{
+					RepoURL: gitOpsDeploymentResource.Spec.Source.RepoURL,
+					Path:    gitOpsDeploymentResource.Spec.Source.Path,
+				},
+			}
+
+			Eventually(gitOpsDeploymentResource, ArgoCDReconcileWaitTime, "1s").Should(
+				SatisfyAll(
+					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
```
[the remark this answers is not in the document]
Rizwana777: Yes , I can check this in previous Eventually

Rizwana777:
@jgwest  Updated changes, PTAL

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #202, opened 2022-08-17
[DEVHAS-148] Add CommitID field to ApplicationSnapshotEnvironmentBinding status

johnmcollier:
#### Description:
This PR adds a new status field to the `BindingComponentGitOpsRepository` section of the ApplicationSnapshotEnvironmentBinding status to include the commit ID for the most recent commit for the Component in its GitOps repository

https://issues.redhat.com/browse/DEVHAS-148


jgwest:
Hey @johnmcollier , thanks for the PR! (BTW, don't worry about fixing the CI job, I'll handle that during merge)

A couple questions come to mind:

### Question 1

I'm curious what the thinking was re: adding a status to each of the components (`SnapshotEnvironmentBinding.status.components[*].commitID`), rather than directly in `SnapshotEnvironmentBindingStatus.status` itself? It's true that some commit might only touch the YAML of a subset of components, but I would think that's not usually the case (and/or it's not a distinction that is useful in practice). I'm not saying it's right or wrong, just curious re: the requirements here.

### Question 2

It seems (correct me if I'm wrong!) like this field can't reliably be used to tell if the contents of the Binding (snapshot, env vars, etc) is up-to-date with Git, which is (I presume) what its purpose is.

Here's a potential race condition that might break this logic:

1) SnapshotEnvironmentBinding.status.components[*].commitID has value of commit `abc`

2) A change is made to one of the fields SnapshotEnvironmentBinding, for example, updating the Snapshot.
- For example, because the user wants to deploy a new version of their application to an Environment.
- This should cause the GitOps Repository to be updated by HAS (in a later step)

3) However, before HAS has a chance to update the GitOps repository, and before HAS updates the `commitID` field, another controller reads the `.status.components[*].commitID` of the SnapshotEnvironmentBinding and sees git commit `abc` (the old value)
- It sees the old commitID `abc`, and incorrectly assumes that commit `abc` contains the update made in step 2 (the snapshot value update)
- But, this is incorrect: Since HAS has not had a chance to update the GitOps repository yet, and thus has not updated the `commitID`, the `commitID` is still referring to the old version.

4) HAS finally updates the `commitID` field, but it is too late because another controller has already made an incorrect assumption in step 3.

johnmcollier:
Discussed on slack with @jgwest - a new story in the DEVHAS jira project will capture the promotion related requirements

jgwest:
New story is: https://issues.redhat.com/browse/DEVHAS-159

jgwest (approved):
LGTM, thanks @johnmcollier!

PR #203, opened 2022-08-18
:recycle: Use any instead of interface{}

drpaneas:
#### Description:

This is a cosmetic change to match the idiomatic way of writing Go >= 1.18 (thus it's breaking if you try to compile with previous versions). Generics aside, if the existing code uses `interface{}`, we can now replace it with `any` throughout, just as the **Go standard library has done**. So, what this PR is doing is something like this:

```go
func Print(args ...interface{}) {
```

can now be refactored very straightforwardly to:

```go
func Print(args ...any) {
```

_NOTE_: This doesn't mean the function is now turned into a generic, nor it has any other effect on it at all. The `any` shortcut it simply a **type alias** for `interface{}`, so Gophers can (and they should IMHO) to use the shorter name anywhere.

In brief, the new name for `interface{}` is `any`.

None


jgwest (approved):
LGTM, thanks @drpaneas, and thanks for helping improve code quality by taking advantage of new Go language improvements :tada:.

PR #204, opened 2022-08-23
Switch to defaulting to new Kubernetes behaviour for Secret/ServiceAccount

jgwest:
#### Description:
- Update the Secret/ServiceAccount creation logic to work with the logic that will work with newer versions of Kubernetes. I've tested it with both OpenShift 4.10 and OpenShift 4.11 (the two versions we support).
- Update the unit test logic which mocks  the ServiceAccount/Secret behaviour
- This removes the failover delay in the previous code (`wait.ErrWaitTimeout`), which would have increased the time to run the E2E test suite when running on OpenShift 4.11 and KCP.

### DIFF
--- a/backend-shared/util/service_account.go
+++ b/backend-shared/util/service_account.go
@@ -2,7 +2,6 @@ package util
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


chetan-rns:
@jgwest  Please remove unused functions to fix lint checks. Everything else looks good to me!

samyak-jn (commented):
/lgtm as well, tested!
Thanks @jgwest :taco:

PR #210, opened 2022-08-31
Update ci dockerfile to match the kcp requirements (Needed: #GITOPSRVC-187)

samyak-jn:
#### Description:
- Update the ci Dockerfile to match the kcp requirements, this PR is intended to add commands to install yq package in the ci's Dockerfile which is required by ckcp.

Ref: Check [logs](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/31857/rehearse-31857-pull-ci-redhat-appstudio-managed-gitops-main-managed-gitops-e2e-tests-on-kcp/1565006131774885888#1:build-log.txt%3A37) from the release PR for the above requirement.

### DIFF
--- a/.ci/openshift-ci/Dockerfile
+++ b/.ci/openshift-ci/Dockerfile
@@ -5,8 +5,10 @@ ENV GOFLAGS=""
 
 SHELL ["/bin/bash", "-c"]
 
-# Install kubectl, postgresql-server
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && \
+# Install yq, kubectl, postgresql-server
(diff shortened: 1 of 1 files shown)


jgwest (approved):
LGTM too, thanks @samyak-jn!

drpaneas:
Some comments:

1. It's not recommended to use different binaries to do the same job, as the maintainers of container images have a tendency to shrink the size down, it's quite common to see binaries that were present at one version, are now not available in the system. That said, I would prefer to either use `wget` or `curl`, but not both.
2. For the case of `kubectl` we are fetching the latest from GitHub. For the case of `yq` you are fetching a specific version. Is there any specific reason for this? e.g. why not fetching the latest `yq` here as well? [1]
3. If `openshift_dev_setup.sh` requires a binary then it should install it.


[1] https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64

PR #212, opened 2022-09-02
Add a constraint to check if tests are running in ci

samyak-jn:
#### Description:
- Add a constraint to check, if the script is running in openshift ci env, if yes, then use k8s resource instead of docker utility.

#### Link to JIRA Story (if applicable): Needed for [GITOPSRVC-187](https://issues.redhat.com/browse/GITOPSRVCE-187)

### DIFF
--- a/kcp/ckcp/setup-ckcp-on-openshift.sh
+++ b/kcp/ckcp/setup-ckcp-on-openshift.sh
@@ -185,10 +185,27 @@ test-gitops-service-e2e-in-kcp() {
   kubectl create ns kube-system || true
   make start-e2e &
   make test-e2e
+} 
+
+test-gitops-service-e2e-in-kcp-in-ci() {
+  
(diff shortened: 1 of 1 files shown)


drpaneas:
I don't understand what's that but seems isolated/harmless and if you need this /lgtm from me as well.

samyak-jn:
Hi @drpaneas :wave: 

To elaborate on why this check is needed?
When we run our ckcp script to test our e2e tests with the kcp environment now (before this) it uses a docker utility which is against the best practices for openshift ci (to create a container within a container); to rectify this, I'm removing any use of docker usage when we are running on openshift ci. The openshift ci uses a flag OPENSHIFT_CI="true" which is this constraint about^

samyak-jn:
Merging this isolated change, needed to trigger CI on release repo, thanks folks for reviewing!

PR #214, opened 2022-09-07
Add more constraints a/c to KCP to make openshift-ci happy (pt: 3/n)

samyak-jn:
#### Description:
- In this PR, there is a check added in `create-dev-env.sh` to skip the check of postgresql-staging pods while running in kcp kubeconfig, as "pods" as a resource is not known at KCP; The actual pods are running in workload end, and hence, changes are made accordingly.

#### Link to JIRA Story (if applicable): Needed [#GITOPSRVC-187](https://issues.redhat.com/browse/GITOPSRVCE-187)

### DIFF
--- a/tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
+++ b/tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
@@ -362,6 +362,11 @@ var _ = Describe("ApplicationSnapshotEnvironmentBinding Reconciler E2E tests", f
 
 		It("should create a GitOpsDeployment that references cluster credentials specified in Environment", func() {
 
+			// ToDo: solve GITOPSRVC-217, and remove this constraint
+			if fixture.IsRunningAgainstKCP() {
+				Skip("Skipping this test because of race condition when running on KCP based env")
+			}
+
 			By("creating second managed environment Secret")
 			secret := corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -146,64 +146,76 @@ if [ "$1" = "kube-auto" ]; then
   done
   echo " * Postgres secret has been created."
 
-   # Wait until postgres pod is running
-  echo " * Wait until Postgres pod is running"
-  counter=0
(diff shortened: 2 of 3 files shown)


tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
```diff
@@ -362,6 +362,10 @@ var _ = Describe("ApplicationSnapshotEnvironmentBinding Reconciler E2E tests", f
 
 		It("should create a GitOpsDeployment that references cluster credentials specified in Environment", func() {
 
+			if fixture.IsRunningAgainstKCP() {
+				Skip("Skipping this test until we support running gitops operator with KCP")
```
jgwest, line 366: I think this error message should be updated to indicate it is a race condition when running on KCP, rather than related to gitops operator, unless I am misunderstanding the issue.

Can you also open a quick JIRA task with a title like 'chore: fix race condition in applicationsnapshotenvrionmentbinding_test.go', pointing to this comment?

jgwest (changes requested):
One minor tweak then good to go.

drpaneas:
Also note: before merging, have a look at our docs at https://github.com/redhat-appstudio/managed-gitops/blob/main/docs/development.md and update it if this PR changes anything related to that. Also I think we need a CI doc to document what is happening in the CI because if we start to diverge between: local test, openshift test, KCP test, I'm afraid we will get lost.

tests-e2e/core/applicationsnapshotenvironmentbinding_test.go
samyak-jn: Aha absolutely I'll do both, silly me :)

samyak-jn:
Checked the document from my end, as well as opened the following stories to address the above-mentioned comments.

[GITOPSRVCE-217](https://issues.redhat.com/browse/GITOPSRVCE-217): chore: fix race condition in applicationsnapshotenvrionmentbinding_test.go
[GITOPSRVCE-218](https://issues.redhat.com/browse/GITOPSRVCE-218): [Docs]: CI/Testing document for distinguishing CI/Local test on KCP vs Non KCP environment

jgwest (approved):
LGTM, thanks @samyak-jn!

PR #217, opened 2022-09-07
Update ckcp setup local script to reflect new changes from the repo (pt: 4/n)

samyak-jn:
#### Description:

- Pulled out the latest changes from ckcp, integrated with the changes required for managed-gitops in `openshift-dev-env.sh` carefully, and tried maintaining the uniformity as much as possible
- Added a stable commit hash which is tested on clusterbot cluster which focuses on release-0.7; with every new release we need to update the script accordingly on ci as well as in our managed-gitops repo
- Added a better way for the cleanup function

To test the changes just run the `make kcp-test-local-e2e` target against a clusterbot cluster (v4.10/v4.11)

### DIFF
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -147,7 +147,7 @@ if [ "$1" = "kube-auto" ]; then
   echo " * Postgres secret has been created."
 
   # Wait until postgres pod is running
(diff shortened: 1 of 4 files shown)


jgwest (approved):
LGTM, thanks @samyak-jn!

PR #222, opened 2022-09-11
Update CPS script to read the KUBECONFIG path as input

chetan-rns:
Remove the CPS KUBECONFIG file and update the script to read the file path as input.


jgwest (approved):
LGTM too, thanks @chetan-rns

PR #223, opened 2022-09-14
Let ci pod use it's local kubeconfig automatically

samyak-jn:
#### Description:
- Let CI pod use its local kubeconfig, instead of explicitly defining "" as default
- Slack Thread for reference, [here](https://coreos.slack.com/archives/CBN38N3MW/p1662790897362269)

#### Link to JIRA Story (if applicable): Needed [#GITOPSRVC-187](https://issues.redhat.com/browse/GITOPSRVCE-187)

### DIFF
--- a/kcp/ckcp/setup-ckcp-on-openshift.sh
+++ b/kcp/ckcp/setup-ckcp-on-openshift.sh
@@ -8,6 +8,7 @@ SCRIPT_DIR="$(
   pwd
 )"
 
+ADMIN_KUBECONFIG="$KUBECONFIG"
 export GITOPS_IN_KCP="true"
 export DISABLE_KCP_VIRTUAL_WORKSPACE="true"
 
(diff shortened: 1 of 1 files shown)


kcp/ckcp/setup-ckcp-on-openshift.sh
```diff
@@ -192,15 +192,14 @@ test-gitops-service-e2e-in-kcp() {
 
 test-gitops-service-e2e-in-kcp-in-ci() {
   
-  export KUBECONFIG=${TMP_DIR}/ckcp-ckcp.default.managed-gitops-compute.kubeconfig
   printf "The Kubeconfig being used for this is:" $KUBECONFIG
```
chetan-rns, line 195: We are printing the KUBECONFIG path here before setting its value. Should it go after line https://github.com/redhat-appstudio/managed-gitops/blob/661336ce4fdda9485af815092124f243e4a0b2b2/kcp/ckcp/setup-ckcp-on-openshift.sh#L197

chetan-rns (changes requested):

kcp/ckcp/setup-ckcp-on-openshift.sh
samyak-jn: Yes, we should found a better way to deal with this coming right up in another commit o/

chetan-rns (approved):
/lgtm

PR #225, opened 2022-09-16
Add a check for postgres pod running and port-forwarding status for CI

samyak-jn:
#### Description:
- It is being observed that while in openshift-ci testing, it starts to port-forward before the Postgres pod is in the Running state; hence, to resolve this, added a check for postgres pod to be in running status if the port is being occupied and finally if the port-forwarding is working, thanks to the create-dev-env script.

Log for ref:
image


chetan-rns (approved):
/lgtm

samyak-jn:
Thanks, @chetan-rns for reviewing, and merging this urgently due to the test on ci, will update more on daily.
CC: @jgwest

PR #226, opened 2022-09-16
Claim GitOpsDeployment resource from backend APIExport using identity hash

chetan-rns:
#### Description:
- Currently, the GitOpsDeployment resource is part of both Backend and Appstudio APIExports. KCP doesn't allow a resource to be part of multiple APIBindings in a workspace and one of the bindings will fail with the below error. Instead, we can claim the duplicate resource from an existing APIExport using an identity hash. This PR adds a script to claim the GitOpsDeployment resource from the backend APIExport.
```shell
'Unable to bind APIs: naming conflict with a bound API gitopsrvc-backend-shared, spec.names.plural=gitopsdeployments is forbidden'
```
- Remove unused APIExport at manifests/dependencies


jgwest (approved):
LGTM, thanks @chetan-rns!

PR #227, opened 2022-09-20
Kustomize overlay with GitOps Service KCP Manifests (Solves: #GITOPSRVC-209)

samyak-jn:
#### Description:
- a kustomize YAML file that points to all the manifests in our repository that are responsible for deploying to KCP
- For now, the kustomisation contains, APIExoorts, APIResourceSchema, Namespace creation for `kube-system`, `gitops-service-argocd`, RBAC's, Workloads, Routes, and a hack for `in-cluster-config` secret mount.
- We have scripts ready for testing against local kcp (ckcp) or using kcp service (CPS), both of the scripts were using some hacks in order to work ArgoCD with sync targets. One of which is included in the kustomisation file which is mounting a secret for `argocd-cluster-config`, located [here](https://github.com/samyak-jn/managed-gitops/blob/kustomize-kcp-manifests-for-gitops/manifests/kcp-resources/argocd-cluster-config.yaml). For another [workaround](https://github.com/samyak-jn/managed-gitops/blob/kustomize-kcp-manifests-for-gitops/kcp/setup-kcp-cps.sh#L67-117) where we were mounting KCP kubeconfig to argocd-server and argocd-controller, we still need to check what's a better way to do this in such scenarios.

- Note: This is the first initial draft, which hasn't been tested yet in production or testing. There are some ToDo(s) within the file that needs to be taken care of. Also, PR #226 needs to go in before this^

#### Link to JIRA Story (if applicable): [GITOPSRVCE-209](https://issues.redhat.com/browse/GITOPSRVCE-209)

### DIFF
--- a/manifests/kcp-resources/argocd-cluster-config.yaml
+++ b/manifests/kcp-resources/argocd-cluster-config.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argocd-cluster-config
+  labels:
+    argocd.argoproj.io/secret-type: cluster
+type: Opaque
+stringData:
+  name: in-cluster
+  namespace: "$ARGOCD_NAMESPACE"
+  server: https://kubernetes.default.svc
+  config: |
+    {
+      "tlsClientConfig": {
+        "insecure": true
+      }
+    }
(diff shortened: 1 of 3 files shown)


jgwest (approved):
LGTM, thanks @samyak-jn!

PR #234, opened 2022-09-29
Sanitize errors into user facing and developer facing errors

raghavi101:
#### Description:
Sanitize user-facing errors from the GitOps service errors. The general intuition followed here is that wherever `managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
` is imported in backend and appstudio-component, the errors are distributed into two : the ones that will be user facing `userError` and the ones that will be developer facing `devError`. Furthermore, for any `k8sClient` calls for Get, Update, Create, Delete are thought to be errors that a user must receive. There might also be miscellaneous locations where the user errors would be present which are thought as a requirement.
https://issues.redhat.com/browse/GITOPSRVCE-77

### DIFF
--- a/backend-shared/util/errors/error.go
+++ b/backend-shared/util/errors/error.go
@@ -1,0 +1,82 @@
+package errors
+
+import (
+	"context"
+	"fmt"
+	"runtime/debug"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// UnknownError should be used if there is no useful information we can provide to the user related to the
+	// context of the error: for example, if the error is related to specific technical internals of the GitOpsService
+	// that we should not expose to the user (for example: 'unable to connect to database')
+	UnknownError = "an unknown error occurred"
+)
+
+// UserError is an error message that contains both:
+// - A GitOps-Service-developer-focused error message
+// - A user-focused error message
+//
+// We differentiate the two, because User Error messages should be sanitized to ensure they do not refer to
+// internal technical details, whereas developed-focused-errors do not have this restriction.
+//
+// Likewise, User Errors tend to contain helpful tips to help the user fix the error. These kind of tips
+// are not (as) useful to GitOps Service developers.
+//
+// A UserError must have both defined. However, if no user error can be provided, use UnknownError as a generic failure message.
+type UserError interface {
+	DevError() error
+	UserError() string
+}
+
+type userErrorImpl struct {
+	userError string
+	devError  error // may be nil
+}
+
+// DevError returns a non-user-facing error, or nil
+func (ue userErrorImpl) DevError() error {
+	if ue.devError != nil {
+		return ue.devError
+	}
+
+	return nil
+}
+
+// DevError return a user-facing error string, or ""
+func (ue userErrorImpl) UserError() string {
+	return ue.userError
+}
+
+// NewDevOnlyError is used if there is no meaningful information that we can provide the user on why the error occurred
+// and/or how to fix it. This function should only be used when the error is primarily referencing internal GitOps Service
+// technical detials.
+func NewDevOnlyError(devError error) UserError {
+	return NewUserDevError(UnknownError, devError)
+}
+
+// NewUserDevError is used in cases where we can provide both a user-focused error, and a developer-focused error.
+// This function should be used when an error occurs, and we can provide both user and dev errors around the context of that error.
+func NewUserDevError(userErrorString string, devError error) UserError {
+
+	log := log.FromContext(context.Background())
+
+	if userErrorString == "" {
+		diagnosticMsg := fmt.Sprintf("SEVERE: Asked to create an Error with nil user error message. devError message is '%v'.", devError)
+		log.Error(nil, diagnosticMsg)
+		debug.PrintStack()
+	}
+
+	if devError == nil {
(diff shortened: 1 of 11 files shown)


backend-shared/util/errors/error.go
```diff
@@ -0,0 +1,82 @@
+package errors
```
jgwest, line 1: Let's change the package from `errors` to `gitopsErrors`, so that it doesn't conflict with the existing built-in `errors` packaged, as noted by Panos in his JIRA story comment.

You will likely also need to update the other code that references this package, and rename the parent folder of this file.

backend-shared/util/errors/error.go
```diff
@@ -0,0 +1,82 @@
...
+		diagnosticMsg := fmt.Sprintf("SEVERE: Asked to create an Error with nil dev error message. userError message is '%v'.", userErrorString)
+		log.Error(nil, diagnosticMsg)
+		debug.PrintStack()
+	}
+
+	return userErrorImpl{
+		userError: userErrorString,
+		devError:  devError,
+	}
+}
```
jgwest, line 82: I think we can also add something like this, which was suggested by Panos on the JIRA story.
 
```go
// GitOpsErrorType can be used with Print to output the error message, for debugging purposes
type GitOpsErrorType int

const (
	DevOnly  GitOpsErrorType = iota
	UserOnly GitOpsErrorType = iota
	All      GitOpsErrorType = iota
)

// Print will output the error message, for debugging purposes
func Print(err UserError, filter GitOpsErrorType) {
	switch filter {
	case DevOnly:
		fmt.Println(err.DevError())
	case UserOnly:
		fmt.Println(err.UserError())
	case All:
		fmt.Println(err.UserError())
		fmt.Println(err.DevError())
	}
}
```

jgwest (changes requested):
Looks good @raghavi101, sorry for the delay in reviewing, and thanks for your patience :sweat_smile: .

I've pushed a change to your branch, you will need to `git pull` on your local branch to pick up my changes.

Changes I made:
- For `Reconcile(...)` functions, the error that they return isn't returned to the user, so fortunately we don't need to modify those. The error they output is just printed to the stderr/stdout (where only developers/admins can see it), so these don't need to be sanitized.
- For `application_event_runner_syncruns.go`, I added a tiny bit of code, plus some cleanup, and some TODOs for whoever works on GITOPSRVCE-44 (but don't worry about them for the current story) 
- I've removed the user error around the inform managed env function (forget the name), as AFAICT the error was not being returned to the user.
    - We can revisit these functions again at a later time, on another story, when we start looking at error reporting for managedenvironments.  
- It looks like the user catch panic utility function isn't need any more once the above changes were made, so I removed it.

Take a look at my changes and making sure I haven't broken anything :smile: .

After that, I added a couple requests below for changes, which are based on Panos's suggestions in the story. Once those are in, we should be good to go.

jgwest:
CI jobs are failing: I think that's just because the PR needs to be rebased.

jgwest:
@raghavi101 

I found this error in the build logs, which is causing the CI failure:

```
------------------------------
• [FAILED] [322.689 seconds]
GitOpsDeployment Condition Tests
/go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_resource_test.go:13
  Errors are set properly in Status.Conditions field of GitOpsDeployment
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_resource_test.go:15
    [It] ensures that GitOpsDeployment .status.condition field is populated with the corrrect error when an illegal GitOpsDeployment is created
    /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_resource_test.go:17

  Begin Captured GinkgoWriter Output >>
    STEP: create an invalid GitOpsDeployment application 10/19/22 10:17:13.429
  << End Captured GinkgoWriter Output

  Timed out after 300.001s.
  Expected
      <bool>: false
  to be true
  In [It] at: /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_resource_test.go:52
------------------------------
```

GitOpsDeploymentCondition: false / Expected: [{ErrorOccurred unable to locate managed environment for new application 0001-01-01 00:00:00 +0000 UTC <nil> True ErrorOccurred}] / Actual: [{ErrorOccurred an unknown error occurred 2022-10-19 10:22:02 +0000 UTC 2022-10-19 10:17:13 +0000 UTC True ErrorOccurred}]


It looks like like 46 of gitopsdeployment_resource_test.go needs to change to `an unknown error occurred`:

```go
			expectedConditions := []managedgitopsv1alpha1.GitOpsDeploymentCondition{
				{
					Type:    managedgitopsv1alpha1.GitOpsDeploymentConditionErrorOccurred,
					Message: "unable to locate managed environment for new application", // <====== this line needs to change
					Status:  managedgitopsv1alpha1.GitOpsConditionStatusTrue,
					Reason:  managedgitopsv1alpha1.GitopsDeploymentReasonErrorOccurred,
				},
			}
```

Then I think it will work.

raghavi101:
@jgwest Yes you were right, I have made changes based on your suggestion and all the CI tests are passing. Now it should be good to go! Thank you!

PR #242, opened 2022-10-10
Refactor CR names of AppStudio-shared components.

jparsai:
### Description
This PR is to rename CRs, Structs, go files, comments, implementation, usage of AppStudio-shared resources and remove `Application` prefix.

### Jira Ticket 
https://issues.redhat.com/browse/GITOPSRVCE-193

### Note
Git is not able to map some of the renamed files with actual file and showing different names before and after refactoring.
Ex: Like in this case `applicationsnapshots.yaml` is renamed to `snapshots.yaml` and `applicationpromotionruns.yaml` is renamed to `promotionruns.yaml`, the content of both files is correct but Git comparison is showing it wrong like this: 
```
cainjection_in_applicationsnapshots.yaml → cainjection_in_promotionruns.yaml
cainjection_in_applicationpromotionruns.yaml → cainjection_in_snapshots.yaml
```


raghavi101:
Hi @jparsai everything looks great! 
Just wanted to bring to your attention that within `appstudio-controller` there are places where I see `Application` prefixes to `Promotion` and `Snapshot` but if that is not required to be changed then everything is good to go.

jparsai:
Got it, it is in Reconciler names and Comments.

raghavi101:
Good to go! Thanks @jparsai

jgwest (changes requested):
Looks great @jparsai! I found a few more cases where the old API names are still used.

If you grep for these terms:
- applicationpromotionrun
- applicationsnapshot
- applicationsnapshotenvironmentbinding

e.g. `grep -r -i "applicationpromotionrun"`

jgwest (approved):
Looks great, thanks @jparsai!

PR #246, opened 2022-10-18
Enhance cleanup for KCP E2E

drpaneas:
#### Description:
This PR is extending the cleanup functionallity for KCP E2E. Especially, it sends a SIGKILL signal followed by SIGTERM in case the processes were not stopped. If all fails, it prints more verbose information about the processes and tries (for one last time) to kill any Go processes (since the Makefile is running `go run ./main.go --this --that --flags`).

The reason for going down that route is because we have experienced a strange behavior for E2E KCP tests, were the controllers were not stopped, thus the CI was timing out -- resulting into a test failure. From the related Slack thread:

> In the [build log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/redhat-appstudio_managed-gitops/237/pull-ci-redhat-appstudio-managed-gitops-main-managed-gitops-e2e-tests-on-kcp/1582082265867882496/build-log.txt) for the test, you can see that:
The test suite failed at 2022-10-17 20:08:53Z (which should be the end of the job)
But, for some reason the controllers weren't killed until 2022-10-17T 21:24:31Z

I have also added a 5 minute timeout for ArgoCD deploy.sh which seems to take 2 hours.

None

### DIFF
--- a/kcp/ckcp/setup-ckcp-on-openshift.sh
+++ b/kcp/ckcp/setup-ckcp-on-openshift.sh
@@ -3,18 +3,22 @@
 set -o errexit
 set -o nounset
 
+echo "========================================="
+echo "Running setup-ckcp-on-openshift.sh script"
+echo "========================================="
+
+# Variables
+# ~~~~~~~~~
+
 SCRIPT_DIR="$(
   cd "$(dirname "$0")" >/dev/null
   pwd
 )"
 
-echo "========================================="
-echo "Running setup-ckcp-on-openshift.sh script"
-echo "========================================="
-
 ARGOCD_MANIFEST="$SCRIPT_DIR/../../manifests/kcp/argocd/install-argocd.yaml"
 ARGOCD_NAMESPACE="gitops-service-argocd"
-export TMP_DIR="$(mktemp -d -t kcp-gitops-service.XXXXXXXXX)"
+TMP_DIR="$(mktemp -d -t kcp-gitops-service.XXXXXXXXX)"
+export TMP_DIR
 export GITOPS_IN_KCP="true"
 export DISABLE_KCP_VIRTUAL_WORKSPACE="true"
 export COMPATIBLE_GO_VERSION="1.18"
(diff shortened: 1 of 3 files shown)


kcp/ckcp/setup-ckcp-on-openshift.sh
```diff
@@ -15,8 +15,54 @@ ARGOCD_MANIFEST="$SCRIPT_DIR/../../manifests/kcp/argocd/install-argocd.yaml"
...
+  for SERVICE in "${proc[@]}"
+  do
+    if pgrep -x "$SERVICE" >/dev/null
+    then
+      echo "$SERVICE is still running..."
+      echo "Checking for Go processes..."
+      pgrep go
+    fi
+  done
+
```
samyak-jn, line 65: Thanks @drpaneas :rocket:  This looks good, just a suggestion instead of this can we just introduce a while() loop or wait.poll() which takes var serviceActive as a boolean, and unless it becomes false we keep running the inner loop until either it becomes false or timeouts and flag is still true;

samyak-jn (commented):
Thanks, Panos!
Just a suggestion added, retesting once to see if the error is reciprocated or was just a CI gitch!

kcp/ckcp/setup-ckcp-on-openshift.sh
samyak-jn: BTW fwiw lol I forgot this is in bash, so yeah same logic with bash options^

drpaneas: This is a good suggestion in general but I'm not if favor of polishing this particular piece of code yet, simply because we do not know what is the root problem here (this code is trying to unravel tha problem). Let's find out first what causes the failure and then we can refactor this to be optimized for sure. Although there is no much gain here, because implementing an "until" loop in bash with booleans (bash doesn't like bools) it's going to complicate the code. This is running on the CI and worst case it will "waste" 10 seconds or 20 -- it's cheap and imho doesn't worth doing it. At least not now, until we know what works and what not.

drpaneas: What we can do to not over-complicate things:

1. Create a separate cleanup.sh script.
2. From this script, call: `timeout 5m bash cleanup.sh`

WDYT?

drpaneas:
Some things I've noticed:

1. Goreman tries to kill the processes and if they don't die [within 10 seconds ](https://github.com/mattn/goreman/blob/29e405014c0287260257dc7b21298587229f42d2/proc.go#L77) it exits 1 (that is the error we get).

The problem is _not_ that.

2. According to Operator-SDK docs: `Stop the operator by pressing ctrl+c in the terminal tab or window the operator is running in`. This will always result into Exit 1.

```bash
cd backend; 
go run ./main.go --zap-log-level info --zap-time-encoding=rfc3339nano
<PRESS CTRL+C>

# ---
^C2022-10-19T02:06:12.131136+02:00	INFO	Stopping and waiting for non leader election runnables
2022-10-19T02:06:12.131175+02:00	INFO	Stopping and waiting for leader election runnables
2022-10-19T02:06:12.131191+02:00	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "gitopsdeploymentsyncrun", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentSyncRun"}
2022-10-19T02:06:12.131217+02:00	INFO	All workers finished	{"controller": "gitopsdeploymentsyncrun", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentSyncRun"}
2022-10-19T02:06:12.131208+02:00	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "gitopsdeploymentrepositorycredential", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentRepositoryCredential"}
2022-10-19T02:06:12.131211+02:00	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "gitopsdeploymentmanagedenvironment", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentManagedEnvironment"}
2022-10-19T02:06:12.131245+02:00	INFO	All workers finished	{"controller": "gitopsdeploymentmanagedenvironment", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentManagedEnvironment"}
2022-10-19T02:06:12.131205+02:00	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "gitopsdeployment", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeployment"}
2022-10-19T02:06:12.131256+02:00	INFO	All workers finished	{"controller": "gitopsdeployment", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeployment"}
2022-10-19T02:06:12.131235+02:00	INFO	All workers finished	{"controller": "gitopsdeploymentrepositorycredential", "controllerGroup": "managed-gitops.redhat.com", "controllerKind": "GitOpsDeploymentRepositoryCredential"}
2022-10-19T02:06:12.131271+02:00	INFO	Stopping and waiting for caches
2022-10-19T02:06:12.131376+02:00	INFO	Stopping and waiting for webhooks
2022-10-19T02:06:12.131428+02:00	INFO	Wait completed, proceeding to shutdown the manager
# ---

$ echo $?
1
```

That is the problem.

Suggestions:

1. Build the bundle the use OLM (overkill for every PR)
2. Build the container and push into the container registry (that would be ideal for E2E tests). In this case we need to be careful because Goreman errors out if the process is not killed within 10 sec, while Operator-SDK errors if gracefull shutdown is not happening within 30 sec.
3. Keep doing what we do. I don't understand why we need to cleanup this, but if we must do it, then running `goreman run stop-all || exit 0` should work. If you see the controllers are indeed killed properly (along with their `main` listening PORT as well). The problem is the return code only.
4. Modify Operator-SDK upstream to return `0` on CTRL+C.
5. Modify our controller to catch CTRL+C and return `0`.


I will go with option 3 here.
Let me know

drpaneas:
Something is wrong with this PR. I will close it and inspect it further and re-submit.

PR #249, opened 2022-10-21
Fix race condition in application event runner when gitopsdeployment shares a name (seen when gitops service is stopped)

jgwest:
#### Description:
- The job of the preprocess event loop was to ensure that each incoming request had an associatedGitOpsDeploymentUID.
- This associatedGitOpsDeploymentUID was then used to shard GitOpsDeployment/SyncRun runners.
- However, sharding on UID, rather than on name, led to the issue seen in this bug (and may cause other issues too).
- Thus this PR remove the preprocess event loop and associatedGitOpsDeplUID, and cleans up code that used them
- (I will update the Google docs documentation too, once this merges)


samyak-jn (approved):
/lgtm, in my review I checked that UID is replaced with the name and namespace (for gitopsDepl)

PR #253, opened 2022-10-27
Check if the APIs are synced in a workspace

chetan-rns:
#### Description:
- Create SyncTarget for a user workspace.
- Currently, we check if a SyncTarget is ready by waiting until the deployment is in the `Ready` state. But the APIs may not be synced yet leading to an error when applying argocd resources. Add a step to check whether the required CRDs are synced by the Syncer.


jgwest (approved):
Works great :tada:, thanks @chetan-rns!

PR #254, opened 2022-10-31
Fix serviceaccount test by searching token based on annotations

akram:
#### Description:
Fix serviceaccount test by searching token based on annotations. This allows test to run on `kind` or `k3s` as the implementation of token links with service account is different.


jgwest (approved):
LGTM too, thanks @akram!  Merging w/o passing KCP test due to issue Samyak mentioned above.

PR #257, opened 2022-11-02
Update manifests to add support for AppStudio KCP staging 

jgwest:
#### Description:
- Updates to controller `Deployment`s and `Roles` to support KCP virtual workspaces
- There was a hardcoded reference to the APIExportName inside backend-shared, I have updated it to use the value provided as a parameter in backend's main.go. 
- Split the existing Argo CD install yaml into individual per-resource files (I used kubernetes-split-yaml for this).


jgwest:
Looks good @drpaneas , thanks for the quick turnaround! Following up on questions from standup...

The `redhat-appstudio/appliation-api` package is a merge of the APIs from the HAS team (Application, Component, ComponentDetectionQuery), and the AppStudio/HACBS Environment API from our team. Because these two APIs used to be separate, there was some lightweight refactoring work that John C et al did initially to get them to play well together.

### Remaining changes for this PR:

Feel free to delete `applicationpromotionrun_types.go`, `applicationsnapshot_types.go` in your PR. I don't see any fields in there that we need.

Snapshot looks good as is.

SnapshotEnvironmentBinding (SEB) looks good, but the HAS folks made this change to spec, which we need to ensure is merged into SEB:
From:
```go
Spec   SnapshotEnvironmentBindingSpec   `json:"spec,omitempty"`
```
to
```go
Spec   SnapshotEnvironmentBindingSpec   `json:"spec"`
```

### Dealing with conflicting types name 

If there are any structs in our code that conflict with structs with the same name in HAS code (HAS code is `application_*`, `component*`, etc), then feel free to refactor the struct names of OUR types. 

For example, it looks like HAS folks renamed OUR `ComponentStatus` to `BindingComponentStatus` in order to avoid the issue of conflicting type names between their and our APIs. This would be an example of a refactoring change we want to preserve with this PR.

In general, we should preserve the type name changes they made (e.g. change our code to use BindingComponentStatus as the new name for our ComponentStatus).


### Out of scope for changes in this PR/story:
	
`application_*` (note the underscore) are owned by the HAS folks, and shouldn't be touched by this PR/story. These are related to AppStudio `Application` resource.

Likewise the `component_types.go` shouldn't need to change, this is owned by HAS folks as well.


@johnmcollier  will be reviewing as well, and will keep us honest, to make sure we aren't breaking anything they added/own :smile: .

jgwest:
Merging without reviews: this PR is KCP only, so I've not chased anyone to review it. I'm mostly merging it to preserve the work in case we pick up KCP support in the medium term. Will clean up all the YAML (both KCP and non-KCP) on a separate story.

PR #266, opened 2022-11-15
Add support for syncing/terminating sync of an Argo CD Application to cluster-agent

jgwest:
#### Description:
- Ensure that only one operation is running at a time for a particular resource (Application/SyncOperation/etc)
- Sugar: `true/false` -> `shouldRetryTrue/False` for functions returning a should retry value 
- Add processing of SyncOperation resource, which will cause Argo CD to do a manual sync (or to terminate a manual sync), based on the contents of the SyncOperation resource in the DB.
- Unit/E2E tests to be worked on elsewhere, this is being pushed to unblock team members

### DIFF
--- a/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
+++ b/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	appv1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
@@ -15,6 +16,7 @@ import (
 	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
 	argosharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util/argocd"
 	"github.com/redhat-appstudio/managed-gitops/cluster-agent/controllers"
+	"github.com/redhat-appstudio/managed-gitops/cluster-agent/utils"
 	corev1 "k8s.io/api/core/v1"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -44,6 +46,13 @@ import (
 type OperationEventLoop struct {
 	eventLoopInputChannel chan operationEventLoopEvent
 }
+
+// Functions that return a boolean indicating whether the request should be retried, should use these constants
+// to improve code readability, rather than just using 'true'/'false'
+const (
+	shouldRetryTrue  = true
+	shouldRetryFalse = false
+)
 
 func NewOperationEventLoop() *OperationEventLoop {
 	channel := make(chan operationEventLoopEvent)
@@ -78,10 +87,47 @@ func operationEventLoopRouter(input chan operationEventLoopEvent) {
 
 	log.Info("controllerEventLoopRouter started")
 
+	dbQueries, err := db.NewSharedProductionPostgresDBQueries(true)
+	if err != nil {
+		log.Error(nil, "SEVERE: Controller event loop exited before startup.")
+		return
+	}
+
+	credentialService := utils.NewCredentialService(nil, false)
+
 	for {
 		newEvent := <-input
 
-		mapKey := newEvent.request.Name + "-" + newEvent.request.Namespace
+		// Generate the map key (which controls task concurrency) by retrieving the Operation from the database
+		// that corresponds to the Operation custom resource from the event.
+		var mapKey string
+		_, err := sharedutil.CatchPanic(func() error {
+
+			dbOperation, err := getDBOperationForEvent(ctx, newEvent, dbQueries, log)
+			if err != nil {
+				return err
+			}
+
+			if dbOperation == nil {
+				// If information could not be retrieved from the database about the operation, then just skip processing the event
+				return nil
+			}
+
+			// Only concurrently process one Application/SyncOperation at a time
+			mapKey = dbOperation.Instance_id + "-" + string(dbOperation.Resource_type) + "-" + dbOperation.Resource_id
+
+			return nil
+		})
+
+		if err != nil {
+			log.Error(err, "unable to acquire information for operation from DB")
+			continue
+		}
+
+		if mapKey == "" {
+			log.Info("mapkey was nil for request, continuing.")
+			continue
+		}
 
 		// Queue a new task in the task retry loop for our event.
 		task := &processOperationEventTask{
@@ -89,7 +135,8 @@ func operationEventLoopRouter(input chan operationEventLoopEvent) {
 				request: newEvent.request,
 				client:  newEvent.client,
 			},
-			log: log,
+			log:               log,
+			credentialService: credentialService,
 		}
 		taskRetryLoop.AddTaskIfNotPresent(mapKey, task, sharedutil.ExponentialBackoff{Factor: 2, Min: time.Millisecond * 200, Max: time.Second * 10, Jitter: true})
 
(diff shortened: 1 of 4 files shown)


cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -319,14 +426,174 @@ func (task *processOperationEventTask) internalPerformTask(taskContext context.C
...
+		log.Error(nil, "SEVERE: unexpected desired state in SyncOperation DB entry")
+		return shouldRetryFalse, nil
+	}
+}
+
+// returns shouldRetry, error
+func terminateExistingOperation(ctx context.Context, dbApplication *db.Application, argoCDNamespace corev1.Namespace,
+	credentialService *utils.CredentialService, eventClient client.Client, log logr.Logger) (bool, error) {
+
+	if err := utils.TerminateOperation(ctx, dbApplication.Name, argoCDNamespace, credentialService, eventClient, time.Duration(5*time.Minute), log); err != nil {
```
drpaneas, line 510: what happens in a situation where this doesn't get deleted (for whatever reason) ? Given that we process on operation at a time, would that create a block the queue for the rest?

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -319,14 +426,174 @@ func (task *processOperationEventTask) internalPerformTask(taskContext context.C
...
+		log.Error(err, "unable to terminate operation: "+dbApplication.Name)
+		return shouldRetryTrue, err
+	}
+
+	log.Info("Successfully terminated operations for application '" + dbApplication.Name + "'")
```
drpaneas, line 515: operationS ? Is this a typo or you are terminating a couple of them here?

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -319,14 +426,174 @@ func (task *processOperationEventTask) internalPerformTask(taskContext context.C
...
+
+	return shouldRetryFalse, nil
+
+}
+
+// returns shouldRetry, error
+func runAppSync(ctx context.Context, dbOperation db.Operation, dbSyncOperation db.SyncOperation, dbApplication *db.Application,
```
drpaneas, line 522: Same code smell here, too many args for a function.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -319,14 +426,174 @@ func (task *processOperationEventTask) internalPerformTask(taskContext context.C
...
+	// Start the AppSync operation in a separate thread.
+	go func() {
+		err = utils.AppSync(cancellableCtx, dbApplication.Name, dbSyncOperation.Revision, argoCDNamespace.Name, eventClient,
+			credentialsService, false)
+
+		var failed bool
+		if err != nil {
+			log.Error(err, "app sync failed on application '"+dbApplication.Name+"'")
+			failed = true
+		} else {
```
drpaneas, line 543: You don't need this. The default value is `false`.

drpaneas (commented):
Is this up for review? If not, consider this as a preliminary:

*  My main concern is related to the implementation of the `runAppSync()` function. It has outer and inner for-loops mixed with go routines and backoffs. The comment (`returns shouldRetry, error`) is not sufficient to me. Can you to explain how this is supposed to work? 
* I've made some comments in the code about re-thinking (extending maybe?) some sanity checks and using the true/false sugar const in more places (if it's applicable).
* The unit-tests are failing in the CI.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -45,6 +47,13 @@ type OperationEventLoop struct {
 	eventLoopInputChannel chan operationEventLoopEvent
 }
 
+// Functions that return a boolean indicating whether the request should be retried, should use these constants
+// to improve code readability, rather than just using 'true'/'false'
+const (
+	shouldRetryTrue  = true
```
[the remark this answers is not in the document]
jgwest: Good thinking, I tried it out, and I think I still like that `shouldRetry` provides additional information on the context of the retry:
* `retry / noRetry` = Does this mean that a retry was already attempted? Does this mean a retry should be attempted?
* `shouldRetryTrue / shouldRetryFalse` = Indicates specifically what retry means here: that it is instructing the caller what should be done (a retry).

I suspect this is entirely about personal preference, and in this case to me it feels like the slightly longer name is a good tradeoff for the information it provides.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -78,29 +87,132 @@ func operationEventLoopRouter(input chan operationEventLoopEvent) {
...
+			if db.IsResultNotFoundError(err) {
+				return true, nil
+			} else {
+				// some other generic error
+				log.Error(err, "Unable to retrieve operation due to generic error: "+dbOperation.Operation_id)
+				return false, err
+			}
+		}
+
+		if strings.TrimSpace(dbOperation.Instance_id) == "" && strings.TrimSpace(dbOperation.Resource_id) == "" &&
```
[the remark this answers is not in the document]
jgwest: Whoops, good catch!

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -319,14 +426,174 @@ func (task *processOperationEventTask) internalPerformTask(taskContext context.C
...
 		log.Error(nil, "SEVERE: unrecognized resource type: "+string(dbOperation.Resource_type))
 		return &dbOperation, shouldRetryFalse, nil
 	}
 
 }
 
-// processOoperation_ManagedEnvironment handles an Operation that targets an Application.
+// Process a SyncOperation database entry, that was pointed to by an Operation CR.
+// returns shouldRetry, error
+func processOperation_SyncOperation(ctx context.Context, dbOperation db.Operation, crOperation operation.Operation, dbQueries db.DatabaseQueries,
```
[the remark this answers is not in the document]
jgwest: :+1: lots of opportunities elsewhere in the codebase to clean this up as well. I went with 4,5,6,7,8, as these values are all contextual environment values that can be shared between multiple functions.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
jgwest: Processing of each operation runs in its own goroutine, sharded by the resource that the operation is pointing to. See `taskRetryLoop.AddTaskIfNotPresent(...)` in this file for where the goroutine is started.

So for example, in this case the following operations would be processed as follows
Operation A: points to Application row A1
Operation B: points to Application row A2
Operation C: points to Application row A1
Operation D: points to SyncOperation row SO1
Operation E: points to SyncOperation row SO1
Operation F: points to SyncOperation row SO2

B, and F would run on their own goroutines, in parallel.
A/C, and D/E would would run sequentially (e.g. first A, then C).

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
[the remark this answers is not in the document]
drpaneas: >  I think I still like that `shouldRetry` provides additional information on the context of the `retry`.
> `retry` / `noRetry` = Does this mean that a retry was already attempted? Does this mean a retry should be attempted?

Indeed, having a `const retry` accessible to the whole pkg might be interpreted in many ways and also misused in certain places. A solution for that would be to limit the scope, and use it only in the place intended to be used. For example,  if I see `return retry` in the `reconcile()` function, it's obvious to me (_and I hope to everyone else, otherwise my suggestion is pointless here_) that the reconciler has to re-reconcile this one more time. Same thing if I see `return noRetry` or `return !retry`, my first guest is that the Reconciler should not try this again. I'm familiar with the concept of the reconciler, so that's where I get the _missing_ context. (it either needs to retry something (`true`) or it doesn't (`false`). Sometimes it gives us the reason (e.g. `err`) or sometimes it doesn't (`null`). It's the same thing with `i` and `v` in a traditional `for` loop: we all know that they stand for the current `index` value, within this limited scope, but as a standalone names they make no sense.


> (Now, OTOH whether to do shouldRetry/shouldNotRetry vs shouldRetryTrue/shouldRetryFalse is another question 🤔) I suspect this is entirely about personal preference, and in this case to me it feels like the slightly longer name is a good tradeoff for the information it provides.

You are making the value part of the name, so it's like saying `const thisNumberIs5 = 5` or `const thirtyFive = 35` which is superfluous. If the name of the variable is not providing the necessary context and you need to embed the value as part of the name, then something is off -- I am afraid that this is making the code less maintainable (e.g. `if !shouldRetryFalse`).

While I am writing this comment, I came up with another, we could say `return tryAgain` which is imperative form of a verb, instructing the caller exactly what to do: try again one more time. The opposite form would be `!tryAgain`.

But anyway, the last thing I want is to get into an endless debate with naming vars. Given my suggestion has already created some confusion for you, while yours hasn't created any confusion to me (just nitpicking here), let's go with yours.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
drpaneas: all clear now, thx

chetan-rns (approved):
/lgtm

chetan-rns:
Thanks, @jgwest I'll include unit/e2e tests in my next PR.

PR #272, opened 2022-11-30
Update backend service resource limit for appstudio controller

jgwest:
#### Description:
- Update backend service resource limit for appstudio controller


jgwest:
Merging w/o review comments to unblock appstudio staging cluster users, and because this is a YAML-only change.

PR #273, opened 2022-11-30
Add connection status condition to managed environment

jopit:

#### Description:
- Updated the reconcile code for managed environments to add a status condition indicating if the connection to the managed environment's cluster succeeded or not.
- Added unit tests
- Added e2e tests


tests-e2e/core/managed_environment_status_test.go
```diff
@@ -0,0 +1,233 @@
...
+			Expect(managedEnv.Status.Conditions).To(HaveLen(1))
+			condition := managedEnv.Status.Conditions[0]
+			Expect(condition.Status).To(Equal(metav1.ConditionTrue))
+			Expect(condition.Reason).To(Equal("Succeeded"))
+			Expect(condition.Message).To(BeEmpty())
+		})
+	})
+})
+
+func HaveStatusCondition(conditionType string) matcher.GomegaMatcher {
```
jparsai, line 171: Can it be moved to tests-e2e/fixture/managedenvironment/fixture.go? since we have similar functions for other CRs defined in same location.

jparsai (commented):
LGTM, I think one minor change can be done here.

jgwest (changes requested):
Looks great, really nice, thanks John! Some minor tweaks, and one possible new story to come out of this. See below for details.

**1) moving the Reason field values into string constants:**

For the "DatabaseError" string constant, and the other strings we are using in the Reason field (such as "UnableToCreateClient"), can we move these to be string contants rather than repeated strings?

Nothing fancy, just something like:
```go
const (
    ConditionReason_DatabaseError = "DatabaseError" 
    ConditionReason_InvalidSecretType = "InvalidSecretType"
    // (...)

    // (open to any other suggestions on constant name format!)
)
```
And then we can reference those constants in the code/tests.

**2) Since this wasn't part of the original story criteria (and probably should have been), I'm happy to move this requirement into a new story, or if you'd like you can tackle it here that works too. Up to you, just let me know**

Since the conditions that we are setting on the ManagedEnvironment are part of our public API (outside users/consumers of the API can see the messages), the condition messages are subject to at least a couple of restrictions:
- The human readable messages must provide information for users on how to solve the problem (IF it is possible for them to do anything: if not, just return a generic error)
- We can't leak any internal service details via these messages

For example: we don't want to return condition error messages (that we return to users) to contain database keys (as these might be used, when combined with a privilege escalation bug, to DOS our service or maliciously attack other users. Note: the primary key for a database UID is not secret/privleged by itself, but it is useful when attempting to exploit other bugs)

I went into a bit more detail here re: user facing vs developer facing errors on https://issues.redhat.com/browse/GITOPSRVCE-77 (and you can see the PR on there for an example). And yes, I definitely think we should find a more permanent place to hold this information.

jgwest (approved):
LGTM, thanks again @jopit!

PR #274, opened 2022-12-02
Add ManagedEnvironments to appstudio-controller-rbac, update RBAC missing tags

jgwest:
#### Description:
- Add missing RBAC tags, and regenerate role.yaml
- Update the standalone appstudio RBAC yaml (which will be deprecated soon by another story)


jgwest:
Merging: no code reviews, but only RBAC YAML changes.

PR #275, opened 2022-12-05
Re-enable creation of GitOpsDeployments for Applications

jgwest:
#### Description:
- In the M7 timeframe, the GitOps Service would create a GitOpsDeployment whenever an AppStudio Application was created.
- As part of adding support for KCP, and in support of enabling the Environment API, we disabled this behaviour in the GitOps Service.
- However, since KCP plans have changed, and non-KCP replacements are not yet completed, we need to re-enable the automatic creation of GitOpsDeployments for AppSTudio Applications.


chetan-rns (approved):
/lgtm

PR #280, opened 2022-12-09
Add created_on field to resource database row

Rizwana777:
Description:
- Added created_on field and updated db-migrations

Link to JIRA Story (if applicable):
https://issues.redhat.com/browse/GITOPSRVCE-301


drpaneas:
The code looks to me. What I don't know though is if there any other place where we need this. e.g. what about the appstudio controller?

db-schema.sql
```diff
@@ -98,7 +98,9 @@ CREATE TABLE ManagedEnvironment (
 	-- pointer to credentials for the cluster
 	-- Foreign key to: ClusterCredentials.clustercredentials_cred_id
 	clustercredentials_id VARCHAR (48) NOT NULL,
-	CONSTRAINT fk_cluster_credential FOREIGN KEY (clustercredentials_id) REFERENCES ClusterCredentials(clustercredentials_cred_id) ON DELETE NO ACTION ON UPDATE NO ACTION
+	CONSTRAINT fk_cluster_credential FOREIGN KEY (clustercredentials_id) REFERENCES ClusterCredentials(clustercredentials_cred_id) ON DELETE NO ACTION ON UPDATE NO ACTION,
+
+	created_on TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```
jgwest, line 103: Can you add a quick comment to the new `created_on` fields in this file, indicating what they do?

backend-shared/config/db/application_test.go
```diff
@@ -41,6 +43,8 @@ var _ = Describe("Application Test", func() {
 
 		err = dbq.GetApplicationById(ctx, &applicationget)
 		Expect(err).To(BeNil())
+		Expect(applicationput.Created_on).To(BeAssignableToTypeOf(timestamp))
```
jgwest, line 46: Rather than checking if it's assignable to a timestamp, lets instead check if the time field is within the last 5 minutes. Something like this:
```
Expect(applicationput.Created_on.After(time.Now().Add(time.Minute*-5))).To(BeTrue(), "Create on should be within the last 5 minutes")
```
(feel free to check my logic in the statement, I haven't tested it. )

Presuming you agree, can you you make the change in all the tests where we are are doing a create, then a get?

backend-shared/config/db/application_test.go
```diff
@@ -51,6 +55,7 @@ var _ = Describe("Application Test", func() {
 			Engine_instance_inst_id: gitopsEngineInstance.Gitopsengineinstance_id,
 			Managed_environment_id:  managedEnvironment.Managedenvironment_id,
 			SeqID:                   int64(seq),
+			Created_on:              applicationget.Created_on,
```
jgwest, line 58: It would be good to instead assign Created_on a new value, and make sure that update correctly updates this field to a new value.

Presuming you agree, we should make the same change to the other tests as well.

jgwest (changes requested):
Thanks @Rizwana777, some minor tweaks re: tests, then good to go.

backend-shared/config/db/application_test.go
Rizwana777: @jgwest I assumed that we don't have to add  new value to Created_on field for update as Created_on field contains a time  when the Application was `created` for eg, if Application was created at 9:05 PM then Created_on field should be 9:05 PM , and if we are updating the same Application at 9:15 PM then we can keep the Created_on field as it is rather than addding a new value(9:15 PM) because Application was created at 9:05 PM(as Created_on field specifies that when the Application was created ), I think we can add new field called Updated_on so that we can see when this Application was updated, please correct me if I am assuming this wrong

jgwest: @Rizwana777 yup, that's fair, thanks for the explanation! We can leave it is as is.

jgwest (approved):
Looks great, thanks @Rizwana777! And thanks for reviewing @drpaneas.

PR #285, opened 2022-12-15
Delete top-level resources first as part of e2e cleanup

jopit:

#### Description:
Add code to the e2e test cleanup code to delete application, environment and snapshot environment binding resources before deleting gitops service resources, since the former are likely owners of the latter and so should be deleted first.


jgwest (approved):
LGTM, thanks @jopit!

PR #287, opened 2022-12-21
Add new argoCD instance

raghavi101:
#### Description:
Add cluster-agent support for creating new Argo CD instances, via GitOpsEngineInstance Operations

https://issues.redhat.com/browse/GITOPSRVCE-175


cluster-agent/utils/argocd_create.go
```diff
@@ -217,12 +217,23 @@ func CreateNamespaceScopedArgoCD(ctx context.Context, argocdCRName string, names
 			ResourceExclusions: string(resourceExclusions),
 		},
 	}
-
+	fmt.Println(namespace)
```
wtam2018, line 220: It looks like you are trying to create the namespace of the OperatorCR which comes from the request task. It is the namespace of the resource that we are reconciling. That namespace must already exist, right?

jgwest:
@raghavi101 

Made some tweaks with my latest commit, feel free to poke holes in my logic :smile: 

- Added new verification steps to unit tests
- Switched to calling `operations.CreateOperation()` to create the Operation db row and CR.
- Improvements to `SetupForTestingDBGinkgo` to handle this new functionality
- Clean up go.mod, and bump up to Argo CD Operator 0.5
- Remove dependency from Argo CD Operator from the backend component
- Misc logic improvements

jgwest:
My current hypothesis is that the remaining E2E test issue will be fixed by: https://github.com/redhat-appstudio/managed-gitops/pull/304

raghavi101:
@jgwest Based on the issue https://github.com/redhat-appstudio/managed-gitops/pull/304 is solving, I think your hypothesis makes sense. Should we test it after https://github.com/redhat-appstudio/managed-gitops/pull/304 gets merged?

tests-e2e/core/argocd_gitopsengine_instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+	dbq, err := db.NewUnsafePostgresDBQueries(true, true)
+	Expect(err).To(BeNil())
+
+	defer dbq.CloseDatabase()
+	var clusterAccess []db.ClusterAccess
+	err = dbq.UnsafeListAllClusterAccess(ctx, &clusterAccess)
+	Expect(err).To(BeNil())
+
+	for _, clusterAccess := range clusterAccess {
+		if clusterAccess.SeqID >= 0 {
```
jgwest, line 97: @raghavi101 Can you explain the significance of checking if `SeqID >= 0`? 

I would have thought that ALL the entries in the database table would have a SeqID >= 0 (since this value should be autogenerated by postgresql in all cases, and since the value starts at 0), which means this code will just delete all the ClusterAccesses, and all the Operations (rather than only those that are specific to the E2E tests).

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -915,9 +926,38 @@ func deleteArgoCDApplicationOfDeletedApplicationRow(ctx context.Context, dbAppli
...
+
+	log := opConfig.log.WithValues("Gitopsengineinstance_id", dbGitopsEngineInstance.Gitopsengineinstance_id)
+
+	err := opConfig.dbQueries.GetGitopsEngineInstanceById(ctx, dbGitopsEngineInstance)
+	if err != nil {
+		log.Error(err, "Unable to retrieve database GitopsEngineInstance row from database")
+		return shouldRetryTrue, err
+	} else {
+		// The ArgoCD CR name is given the name of the namespace it is being created in
+		fmt.Println(dbGitopsEngineInstance.Namespace_name)
```
jgwest, line 951: A few `fmt.Println` snuck into the code, and should be removed, like this one. If you look through the PR tab and search for `fmt.Println` you can find them all.

tests-e2e/core/argocd_gitopsengine_instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+	Context("ArgoCD instance gets created from an operation's gitopsEngineInstance resource-type", func() {
+
+		BeforeEach(func() {
+			By("Delete old namespaces, and kube-system resources")
+			Expect(fixture.EnsureCleanSlate()).To(Succeed())
+
+		})
+
+		It("ensures that a standalone ArgoCD gets created successfully when an operation CR of resource-type GitOpsEngineInstance is created", func() {
+			err := unsafeCleanDBWhereIDisAutoGenerated()
```
jgwest, line 33: Do you by chance know which other E2E test was interfering with this test?

jgwest (changes requested):

tests-e2e/core/argocd_gitopsengine_instance_test.go
raghavi101: I observed two things, either a test was affecting the current test or the current test was affecting the other tests. In sequence, whichever was being tested _before_ and _after_ the current test should be the answer here. I haven't seen a recurring pattern of tests on ci but I might be wrong.

tests-e2e/core/argocd_gitopsengine_instance_test.go
raghavi101: @jgwest This would be hazardous but I could not find a way other than this 
- which will clean the gitopsEngineInstance values. As we are using GetOrCreate for gitopsEngineInstance, if the code GETS a gitopsEngineInstance it will pick that up, namespace will be as present in its db entry overall discarding the namespace passed in the e2e and the test would fail
- after running the test there will be only one gitopsEngineInstance entry left and other tests will get affected because the altered gitopsEngineInstance will be picked and the db entries will be mismatched as required by those tests

wtam2018:
@jgwest please do a final review.

jgwest (approved):
LGTM, thanks for the detailed investigations @raghavi101, and thanks for your patience in getting this across the finish line!

PR #290, opened 2022-12-23
Report errors to GitOpsDeploymentSyncRun status conditions

chetan-rns:
#### Description:
- Set conditions based on the errors received while handling GitOpsDeploymentSyncRun
- Include unit tests

https://issues.redhat.com/browse/GITOPSRVCE-44

### DIFF
--- a/backend/eventloop/application_event_loop/application_event_runner_syncruns.go
+++ b/backend/eventloop/application_event_loop/application_event_runner_syncruns.go
@@ -33,12 +33,6 @@ func (action *applicationEventLoopRunner_Action) applicationEventRunner_handleSy
 	// Handle all GitOpsDeploymentSyncRun related events
 	err := action.applicationEventRunner_handleSyncRunModifiedInternal(ctx, dbQueries)
 
-	// TODO: GITOPSRVCE-44: Get the SyncRunModified object from k8s, so we can update it if necessary
-
-	// TODO: GITOPSRVCE-44: Add the ErrorOccurred condition to the syncRunModified
-
-	// TODO: GITOPSRVCE-44: If no user error (just dev error), then output generic error occurred
-
 	syncRunCR, clientErr := getGitOpsDeploymentSyncRun(ctx, action.workspaceClient, action.eventResourceName, action.eventResourceNamespace)
 	if clientErr != nil {
 		if !apierr.IsNotFound(clientErr) {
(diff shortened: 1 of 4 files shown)


backend/eventloop/application_event_loop/application_event_runner_syncruns.go
```diff
@@ -33,20 +33,93 @@ func (action *applicationEventLoopRunner_Action) applicationEventRunner_handleSy
...
+		if !apierr.IsNotFound(clientErr) {
+			return fmt.Errorf("unable to get GitOpsDeploymentSyncRun: %v", err)
+		}
+		return nil
+	}
 
+	conditionType := managedgitopsv1alpha1.GitOpsDeploymentSyncRunConditionErrorOccurred
 	if err != nil {
+		// set the condition only for user errors
+		if err.UserError() != "" {
```
drpaneas, line 47: I don't understand thing. Can you explain what you are trying to achieve?

chetan-rns: The `err` here contains both user-focussed error and developer-focussed error messages. We want only user errors to be reported in the status and log dev-related errors.

backend/eventloop/application_event_loop/application_event_runner_syncruns.go
```diff
@@ -33,20 +33,93 @@ func (action *applicationEventLoopRunner_Action) applicationEventRunner_handleSy
...
+	if clientErr != nil {
+		if !apierr.IsNotFound(clientErr) {
+			return fmt.Errorf("unable to get GitOpsDeploymentSyncRun: %v", err)
+		}
+		return nil
+	}
 
+	conditionType := managedgitopsv1alpha1.GitOpsDeploymentSyncRunConditionErrorOccurred
 	if err != nil {
+		// set the condition only for user errors
```
jgwest, line 46: In the case where an error occurred, but it is a dev-only error, I think we should still set the Error Ocurred condition, BUT we should use a generic error message like 'an internal error occurred'.

This is what I did for `GitOpsDeployment`s:
https://github.com/redhat-appstudio/managed-gitops/blob/6197011ff97aa22cefb45e9a96a9022ab8243d2d/backend/eventloop/application_event_loop/application_event_runner_deployments.go#L910

It's not perfect, because it presents the user with an error that they have no way to resolve (they should probably contact support), but it does at least tell them that something is wrong. WDYT?

jgwest (changes requested):

jgwest (approved):

PR #294, opened 2023-01-04
Remove all uses of metadata.Namespace field of Namespace resources

jgwest:
#### Description:
- When creating a Namespace, you should not specify the .metadata.Namespace field, e.g.

```yaml
kind: Namespace
metadata:
  name: my-namespce
  namespace: my-namespace  # this field is not require,d or correct, for Namespaces.
```
- Namespaces are not namespace-scoped, and thus should not have a `namespace:` field.
- That one shouldn't do this is pretty obvious, but there were a number of places in the code (primarily in unit tests), where we WERE doing this, and relying on this field
- This PR removes references to the field, and updates tests as needed.


drpaneas (approved):
/lgtm

PR #301, opened 2023-01-10
Move db code from config/db to db in backend-shared

chetan-rns:
#### Description:
- Move db code from config/db to db in backend-shared
- Fix import paths

https://issues.redhat.com/browse/GITOPSRVCE-310


jgwest (approved):
Great stuff, thanks @chetan-rns!

PR #302, opened 2023-01-10
Sanitize messages for managed environment connection status

jopit:

#### Description:

- Updated the code that sets the connection condition status to ensure the user messages don't expose internal details


jgwest (approved):
Looks great, thanks @jopit!

PR #303, opened 2023-01-10
Update push.yaml to support new kustomize repository URL format

jgwest:
#### Description:
- The infra-deployments repository uses this value to determine how to update the repository with new container images/commits, [for example](https://github.com/redhat-appstudio/infra-deployments/pull/1108/files)


chetan-rns (approved):
/lgtm

chetan-rns:
I'm assuming the revision parameter will always be a full SHA. If not, we can use `[0-9a-f]{5,40}` to be on the safer side

PR #304, opened 2023-01-11
compareApplications function panic on nil in Argo CD Application fields.

jparsai:
#### Description:
This PR is to fix a bug where compareApplications panics if Argo CD app and nil in Spec.SyncPolicy fields.

https://issues.redhat.com/browse/GITOPSRVCE-344


cluster-agent/controllers/argoproj.io/namespace_reconciler.go
```diff
@@ -269,21 +269,26 @@ func compareApplications(applicationFromArgoCD appv1.Application, applicationFro
...
-	if applicationFromArgoCD.Spec.SyncPolicy.Automated.SelfHeal != applicationFromDB.Spec.SyncPolicy.Automated.SelfHeal {
-		log.Info("SelfHeal field in ArgoCD and DB entry is not in Sync.")
-		isAutomatedSelfHealUpdateNeeded = true
-	}
-
 	var isAutomatedAllowEmptyUpdateNeeded bool
-	if applicationFromArgoCD.Spec.SyncPolicy.Automated.AllowEmpty != applicationFromDB.Spec.SyncPolicy.Automated.AllowEmpty {
-		log.Info("AllowEmpty field in ArgoCD and DB entry is not in Sync.")
-		isAutomatedAllowEmptyUpdateNeeded = true
+
```
jgwest, line 274: I think this will miss the case where, for example:
- syncPolicy is nil in the database
- syncPolicy is not-nil on the cluster
- or vice versa

E.g. we should handle the case where at least one of them is nil.

With the new version of the logic, it will return false, when it should return true.

It also could not hurt to add some additional unit tests for this function to catch this.

jgwest (changes requested):

cluster-agent/controllers/argoproj.io/namespace_reconciler.go
jparsai: Logic is changed now, PTAL.

jgwest:
@jparsai, FYI I pushed a new commit that does the following:
- I took the existing reflect.DeepEqual()-based logic we were using in operation_event_loop.go, and moved it into controllers/util.go. 
    - It makes sense to combine both places where we are comparing Argo CD Application with database Application
- Next, I moved both pieces of code (the operation event loop, and the namespace reconciler) to use it.
- Next, I ported over your tests.
- Finally, your test discovered a bug in the original operation_event_loop.go reflect.DeepEqual()-based logic, related to comparing empty []string slices, so I fixed that.

Please take a look at my changes, see if you can see if I am breaking anything and/or feel free to poke any holes in my logic!

jparsai:
> @jparsai, FYI I pushed a new commit that does the following:
> 
> * I took the existing reflect.DeepEqual()-based logic we were using in operation_event_loop.go, and moved it into controllers/util.go.
>   
>   * It makes sense to combine both places where we are comparing Argo CD Application with database Application
> * Next, I refactored both pieces of code (the operation event loop compare logic, and the namespace reconciler compre logic) to use it.
> * Next, I ported over your tests.
> * Finally, your test discovered a bug in the original operation_event_loop.go reflect.DeepEqual()-based logic (sweat_smile) , related to comparing empty []string slices, so I fixed that.
> 
> Please take a look at my changes, see if you can see if I am breaking anything and/or feel free to poke any holes in my logic!

@jgwest this seems to be a better logic, Thank you for updating it.

PR #306, opened 2023-01-11
Enable Prometheus metrics and contribute initial Grafana/Prometheus setup

jgwest:
#### Description:
- Enable metrics scraping of controller metrics via ServiceMonitors
- Disable the use of auth proxy on controllers, so that Prometheus can actually access the metrics :stuck_out_tongue: 
- Contribute an initial developer-only Grafana/Prometheus install script, which GitOps Service developers can use to build Grafana dashboards, and to monitor their GitOp Service installs on K8s
    - To install Grafana and enable Prometheus, run `manifests/scripts/metrics-dashboard/run.sh`
#### Link to JIRA Story (if applicable): [GITOPSRVCE-339](https://issues.redhat.com/browse/GITOPSRVCE-339)

### DIFF
--- a/manifests/scripts/metrics-dashboard/README.md
+++ b/manifests/scripts/metrics-dashboard/README.md
@@ -1,7 +1,14 @@
 # Enable Prometheus logging of GitOps Service in 'gitops' namespace, and install Grafana 
 
+This shell script will:
+- Enable (Prometheus) user monitoring of projects on the cluster. This allows us to use the OpenShift cluster's prometheus, rather than installing our own
+- Install a new Grafana instance into 'grafana' namespace. OpenShift doesn't have a built-in Grafana instance (it has its own metrics graphing mechanism via OpenShift Console)
+- Enables Prometheus to scrape resources from OpenShift GitOps
+- Creates the [Argo CD default Grafana Dashboard](https://argo-cd.readthedocs.io/en/stable/operator-manual/metrics/#dashboards)
+
+
 ### How to Use
 1) Acquire an OpenShift Cluster (for example, cluster bot) and log into it.
-2) Install OpenShift GitOps operator to it.
+2) Install OpenShift GitOps operator to it. (This is required because the install script will attempt to enable Prometheus integration on GitOps Operator)
+    - For example, by running `make install-argocd-openshift`
 3) Run the `run.sh` install script in this directory.
-
--- a/manifests/scripts/metrics-dashboard/README.md
+++ b/manifests/scripts/metrics-dashboard/README.md
@@ -1,0 +1,7 @@
+# Enable Prometheus logging of GitOps Service in 'gitops' namespace, and install Grafana 
+
+### How to Use
+1) Acquire an OpenShift Cluster (for example, cluster bot) and log into it.
+2) Install OpenShift GitOps operator to it.
+3) Run the `run.sh` install script in this directory.
+
--- a/manifests/scripts/metrics-dashboard/prometheus/openshift-operators-service-monitor.yaml
+++ b/manifests/scripts/metrics-dashboard/prometheus/openshift-operators-service-monitor.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    name: gitops-operator
+  name: gitops-operator-metrics-in-gitops
+  namespace: openshift-gitops
+spec:
+  endpoints:
+  - bearerTokenSecret:
+      key: ""
+    port: http-metrics
+  - bearerTokenSecret:
+      key: ""
+    port: cr-metrics
+  namespaceSelector:
+    matchNames:
+      - openshift-operators
+  selector:
+    matchLabels:
+      name: gitops-operator
--- a/manifests/scripts/metrics-dashboard/run.sh
+++ b/manifests/scripts/metrics-dashboard/run.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+# -----------------
+
+# Using OpenShift's Prometheus instance requires us to enable user workload monitoring via the openshift-monitoring ConfigMap:
+# https://docs.openshift.com/container-platform/4.11/monitoring/enabling-monitoring-for-user-defined-projects.html)
+# - BUT, the way I'm doing it here is a brittle hack :P
+kubectl apply -f prometheus/enable-user-workload-monitoring.yaml
+
+# Allow Prometheus process to view the openshift-operators/gitops namespaces
+kubectl apply -f prometheus/prometheus-roles-for-openshift-operators.yaml -n openshift-operators
+kubectl apply -f prometheus/prometheus-roles-for-gitops-ns.yaml -n gitops
+
+# -----------------
+echo 
+echo "* Kick off an OLM install of Grafana, and wait for it to complete"
+
+kubectl create ns grafana > /dev/null 2>&1
+
+kubectl apply -f grafana/grafana-operator-group.yaml -n grafana
+kubectl apply -f grafana/grafana-subscription.yaml -n grafana
+
+echo
+echo "* Waiting for Grafana CRDs to exist"
+
+while : ; do
+  kubectl get customresourcedefinition/grafanas.integreatly.org  > /dev/null 2>&1 && break
+  sleep 1s
+done
+
+# -----------------
+
+echo "* Waiting for Grafana route"
+
+while : ; do
+  kubectl get route/grafana-route -n grafana -o yaml  > /dev/null 2>&1 && break
+  sleep 1s
+done
+
+export HOSTNAME=`kubectl get route/grafana-route -n grafana -o yaml  | grep "    host:" | cut -c11-`
+export ADMIN_SECRET_VALUE=`openssl rand -hex 10`
+
+echo
+echo "* Grafana route is: https://$HOSTNAME"
+echo "  Username: user"
+echo "  Password: $ADMIN_SECRET_VALUE"
+echo
+
+TMP_DIR=`mktemp -d`
+
+# Substitute the cluster domain into the Grafana Ingress CR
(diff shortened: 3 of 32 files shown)


manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
jopit, line 1: Is this script and the associated yaml files only used to enable us to create our grafana dashboard (i.e. for [GITOPSRVCE-340](https://issues.redhat.com/browse/GITOPSRVCE-340))?

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# -----------------
+
+# Using OpenShift's Prometheus instance requires us to enable user workload monitoring via the openshift-monitoring ConfigMap:
+# https://docs.openshift.com/container-platform/4.11/monitoring/enabling-monitoring-for-user-defined-projects.html)
+# - BUT, the way I'm doing it here is a brittle hack :P
+kubectl apply -f prometheus/enable-user-workload-monitoring.yaml
```
jopit, line 8: It's probably possible to use the `openshift.io/cluster-monitoring` label on the various namespaces to enable prometheus scraping without having to enable user workload monitoring.  But it may not be worth doing right now.

jopit (approved):
A couple of questions, comments, but it basically lgtm.

manifests/scripts/metrics-dashboard/README.md
```diff
@@ -0,0 +1,7 @@
+# Enable Prometheus logging of GitOps Service in 'gitops' namespace, and install Grafana 
+
+### How to Use
+1) Acquire an OpenShift Cluster (for example, cluster bot) and log into it.
+2) Install OpenShift GitOps operator to it.
```
jopit, line 5: Should that be the GitOps service operator?  If not, when does that operator get installed?

jopit (approved):

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+
+kubectl apply -f grafana/grafana-operator-group.yaml -n grafana
+kubectl apply -f grafana/grafana-subscription.yaml -n grafana
+
+echo
+echo "* Waiting for Grafana CRDs to exist"
+
+while : ; do
+  kubectl get customresourcedefinition/grafanas.integreatly.org  > /dev/null 2>&1 && break
+  sleep 1s
```
jopit, line 28: On mac, `sleep 1s` produces an error, it wants `sleep 1`

jopit (changes requested):

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+done
+
+# -----------------
+
+echo "* Waiting for Grafana route"
+
+while : ; do
+  kubectl get route/grafana-route -n grafana -o yaml  > /dev/null 2>&1 && break
+  sleep 1s
```
jopit, line 37: ditto

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# The kubectl equivalent to: 'oc adm policy add-cluster-role-to-user cluster-monitoring-view -z grafana-serviceaccount'
+kubectl apply -f grafana/grafana-cluster-role-binding.yaml -n grafana
+
+# -----------------
+echo
+echo "* Waiting for Grafana service account token secret to exist"
+while : ; do
+  kubectl get secrets -n grafana | grep "grafana-serviceaccount-token"  > /dev/null 2>&1 && break
+  sleep 1s
```
jopit, line 69: ditto

manifests/scripts/metrics-dashboard/run.sh
jgwest: @jopit correct. It's primarily for us to run locally on our own machines, in order to work on new metrics or Grafana dashboards. (Thus it doesn't need to be production ready, etc. Just for our use as devs)

In the Stone Soup staging and production environments, they have they own mechanism for setting up Prometheus and Grafana: https://github.com/redhat-appstudio/infra-deployments/tree/main/components/monitoring

manifests/scripts/metrics-dashboard/run.sh
jgwest: @jopit That would definitely be much easier... does OpenShift support using this label for user-defined projects? E.g. perhaps it's something that technically works, but isn't for users to use?

I only ask because I don't see any reference to this label on [user monitoring page](https://docs.openshift.com/container-platform/4.11/monitoring/enabling-monitoring-for-user-defined-projects.html), or any other references in the docs that indicate it can be used this way, which makes me wonder :thinking: .

jopit: Right it's not for users, its for Red Hat. But we are Red Hat.  Not sure what the actual restrictions on using it are, the PM's said we should use it for the gitops operator to set up monitoring for it's argo-cd instances.

jgwest: Ah, gotcha, yup, that makes sense.

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+echo "* Grafana route is: https://$HOSTNAME"
+echo "  Username: user"
+echo "  Password: $ADMIN_SECRET_VALUE"
+echo
+
+TMP_DIR=`mktemp -d`
+
+# Substitute the cluster domain into the Grafana Ingress CR
+cp -f grafana/grafana-cr.yaml $TMP_DIR/grafana-cr-resolved.yaml
+sed -i 's/HOSTNAME/'$HOSTNAME'/g' $TMP_DIR/grafana-cr-resolved.yaml
```
jopit, line 53: On mac, `sed -i` requires a backup suffix, for example `sed -i.bak`, which will also work on linux I believe.

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+sed -i 's/ADMIN_SECRET_VALUE/'$ADMIN_SECRET_VALUE'/g' $TMP_DIR/grafana-cr-resolved.yaml
```
jopit, line 54: ditto sed

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+echo
+echo "* Applying GrafanaDataSource, using Grafana Service Account Token"
+
+GRAFANA_SECRET=`kubectl -n grafana get secrets | grep "grafana-serviceaccount-token" |  cut -d ' ' -f 1`
+
+GRAFANA_SA_TOKEN=`kubectl -n grafana get secret $GRAFANA_SECRET -o jsonpath={.data.token} | base64 -d`
+
+cp -f grafana/grafana-data-source.yaml  $TMP_DIR/grafana-data-source-resolved.yaml
+
+sed -i 's/GRAFANA_SA_TOKEN/'$GRAFANA_SA_TOKEN'/g' $TMP_DIR/grafana-data-source-resolved.yaml
```
jopit, line 81: ditto sed

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+kubectl apply -f dashboards/argo-cd/grafana-argo-dashboard.yaml
+
+# Create a ServiceMonitor for GitOps Operator, in the openshift-gitops namespace
+# - This SHOULD instead be created in the openshift-operators namespace (which is where the gitops-operator lives),
+#   but the prometheus-operator process has a hardcoded list of namespaces that it checks, openshift-operators
+#   is not on it.
+
+echo "* Create GitOps Operator ServiceMonitor"
+
+kubectl apply -f prometheus/openshift-operators-service-monitor.yaml -n openshift-gitops
```
jopit, line 104: Aren't we supposed to set up monitoring for the gitops-service-operator rather than the gitops-operator?

manifests/scripts/metrics-dashboard/prometheus/openshift-operators-service-monitor.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
```
jopit, line 1: Shouldn't the service monitor be for the gitops-service-operator rather than the gitops-operator?  Or am I totally confuse about what we're doing?

manifests/scripts/metrics-dashboard/run.sh
```diff
@@ -0,0 +1,109 @@
...
+while : ; do
+  kubectl get customresourcedefinition/grafanas.integreatly.org  > /dev/null 2>&1 && break
+  sleep 1s
+done
+
+# -----------------
+
+echo "* Waiting for Grafana route"
+
+while : ; do
```
jopit, line 35: When I ran the script, the grafana route didn't get created automatically.  I had to create it manually:
```
oc create route edge grafana-route -n grafana --service grafana-operator-controller-manager-metrics-service
```

jgwest: Hmm, not sure how it was working before :thinking:.

manifests/scripts/metrics-dashboard/run.sh
jgwest: I've updated the README.md with more context on what is being setup.

manifests/scripts/metrics-dashboard/README.md
jgwest: I've updated the README.md with more information about what is being exposed.

jopit (approved):
Tested, the updated script on a mac, works fine.  lgtm

PR #307, opened 2023-01-16
Refactor CleanupOperation in backend-shared

chetan-rns:
#### Description:
- Fix the CleanupOperation() in backend-shared
- Add/Fix unit tests

https://issues.redhat.com/browse/GITOPSRVCE-290

### DIFF
--- a/backend-shared/util/operations/types.go
+++ b/backend-shared/util/operations/types.go
@@ -27,7 +27,7 @@ const (
 const operationCRNamePattern = "operation-%s"
 
 // CreateOperation will create an Operation CR on the target GitOpsEngine cluster, and a corresponding entry in the
-// database. It will then wait for that operatio to complete (if waitForOperation is true)
+// database. It will then wait for that operation to complete (if waitForOperation is true)
 // - In order to avoid intermittent issues, the Operation could will keep trying for 60 seconds.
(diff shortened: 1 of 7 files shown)


backend-shared/util/operations/types.go
```diff
@@ -171,26 +171,28 @@ func createOperationInternal(ctx context.Context, waitForOperation bool, dbOpera
 
 }
 
-// cleanupOperation cleans up the database entry and (optionally) the CR, once an operation has concluded.
+// cleanupOperation cleans up the operation CR and (optionally) the database entry, once an operation has concluded.
 func CleanupOperation(ctx context.Context, dbOperation db.Operation, k8sOperation managedgitopsv1alpha1.Operation, operationNamespace string,
```
chetan-rns, line 175: In several unit tests, we are checking if the Operation rows are created. So, I'm assuming the original intention behind [commenting](https://github.com/redhat-appstudio/managed-gitops/pull/307/files#diff-acb15696e94beedbf9da214a7eee8ce27d9734bce6e5a8ea3ab2f2ac0e165a4cL180-L187) the DB entry deletion was to test the creation of Operation rows.

I've refactored the code to make the DB entry deletion optional instead of the other way around. By doing so, the existing unit tests still pass. Let me know if there will be any unwanted outcomes with this logic.

jgwest (approved):
Looks great, thanks @chetan-rns! I see no obvious issues with the behavior changes.

PR #309, opened 2023-01-17
Patch backend component on stonesoup cluster: restore controller limits to original values

jgwest:
#### Description:
- Restore the controller limits to their original values: they were unintentionally reduced by the latest manifest update.


jgwest:
Merging w/o reviews, as this a YAML change only.

PR #312, opened 2023-01-18
Fix continual increment of deployment failures metric

jopit:

#### Description:
- Change `AddOrUpdateGitOpsDeployment()` to check if an entry for the deployment already exists before setting its error condition to `false`.


jgwest (approved):
Great catch, great fix, great to get a unit test to verify and keep it from regressing again, thanks @jopit!

PR #317, opened 2023-01-24
Add Stonesoup-specific Prometheus YAML

jgwest:
#### Description:
- The Stonesoup staging cluster users its own Prometheus instance, which is not compatible with the OpenShift-based user-workload instance.
- I've thus customized the YAML to work for both cases. :crossed_fingers:


jgwest:
Merging w/o review comments, since it's YAML only. I'm sure the YAML/kustomize organization could be improved, but first step is getting it working :smile:.

PR #321, opened 2023-01-25
Add RBAC to allow controllers to create TokenReviews, add missing appstudio-controller deployment patch

jgwest:
#### Description:
- Add RBAC to allow controllers to create TokenReviews (this was failing on Stonesoup cluster to missing roles/rolebindings)
- Add missing appstudio-controller deployment patch (this was inadvertently excluded).


jgwest:
Merging w/o review comments or tests, as the change is YAML only. The E2E tests on stonesoup cluster should catch any issues.

PR #323, opened 2023-01-25
upgrade db-migrate in backend

saumeya:
#### Description:


jannfis:
Where do all the changes to the `go.mod` files come from, i.e. moving around dependencies, commenting out some here and there? This looks weird. Is it intentional?

PR #325, opened 2023-01-25
Move dashboard into base/monitoring, add ConfigMap generator for Stonesoup

jgwest:
#### Description:
- In order to expose our dashboard JSON on the Stonesoup staging cluster, I'm [following their example](https://github.com/redhat-appstudio/service-provider-integration-operator/tree/main/config/monitoring/base) of converting the dashboard JSON into a ConfigMap. 
- At present, this will only work on Stonesoup cluster, but we can also add automatic dashboard import into our own Grafana/Prometheus script with a separate PR


jgwest:
Merging w/o review comments/CI, as YAML only and the YAML is not linked from this repository.

PR #328, opened 2023-01-28
Handle references to both ManagedEnv/Secrets in ManagedEnv controller

jgwest:
#### Description:
- When K8s resource are processed by `Reconcile()` of the ManagedEnvironment controller, in some cases they will be Secrets, and in some cases they will be GitOpsDeploymentManagedEnvironments.
- This is because the controller watches both types of resources via Watch.
- The old logic did not properly handle the above fact: it would not miss valid requests in some cases. This PR fixes that logic.
- I'd also added a filter to filter out Secret events in `openshift-*` and `kube-system` Namespaces, which will never be supported namespaces for our API requests.


jgwest:
/retest

Test failed due to https://github.com/redhat-appstudio/managed-gitops/pull/337

PR #329, opened 2023-01-28
Skip reconciliation of events for resources in Namespaces that are being deleted

jgwest:
#### Description:
- Skip reconciliation of events for resources in Namespaces that are being deleted
- We want to avoid the case where the GitOps Service/Argo CD are keeping the Namespace from being deleted by creating/recreating resources as they are deleted.


flacatus (approved):

jgwest:
Thanks for reviewing @chetan-rns and @flacatus!

PR #333, opened 2023-02-01
Update request/limits for backend and cluster-agent components, for Stonesoup

jgwest:
#### Description:
- Update request/limits for backend/cluster-agent components when running on Stonesoup
- Reduce requests/limits when not running on Stonesoup (e.g. on dev's machines)


jgwest:
Merging w/o comments as this is YAML only.

PR #334, opened 2023-02-01
Add printcolumn tags for sync and health status

chetan-rns:
#### Description:
- Add kubebuilder print column tags to show sync and health status columns
```shell
$ kubectl get gitopsdeployments.managed-gitops.redhat.com -A
NAMESPACE   NAME          SYNC STATUS   HEALTH STATUS
jane        gitops-depl   Synced        Healthy
jgw         gitops-depl   Synced        Degraded
```


jgwest (approved):
Thanks for adding this, this should be very helpful for consuming the API, much faster than having to do `kubectl get` on the individual GitOpsDeployments. LGTM!

PR #337, opened 2023-02-01
Fix test race condition in SnapshotEnvironmentBinding E2E test

jgwest:
#### Description:
- Fix a race condition I noticed in the SnapshotEnvironmentBinding tests: we should wait for the GitOpsDeployment to exist (and not exist), rather than assuming it that it will be created quickly.

### DIFF
--- a/tests-e2e/core/snapshotenvironmentbinding_test.go
+++ b/tests-e2e/core/snapshotenvironmentbinding_test.go
@@ -414,11 +414,11 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler E2E tests", func() {
 
 			// Check no GitOpsDeployment CR found with default name (longer name).
 			gitOpsDeployment := buildGitOpsDeploymentObjectMeta(gitOpsDeploymentName, binding.Namespace)
-			Consistently(gitOpsDeployment, "30s", "1s").ShouldNot(k8s.ExistByName(k8sClient), "wait 30s for the object not to exist")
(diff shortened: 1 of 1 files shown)


tests-e2e/core/snapshotenvironmentbinding_test.go
```diff
@@ -414,13 +414,11 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler E2E tests", func() {
...
-			err = k8s.Get(&gitOpsDeployment, k8sClient)
-			Expect(apierr.IsNotFound(err)).To(BeTrue())
+			Consistently(gitOpsDeployment, "30s", "1s").ShouldNot(k8s.ExistByName(k8sClient), "wait 30s for the object not to exist")
 
-			// Check GitOpsDeployment is created with short name).
+			// Check GitOpsDeployment is created with short name.
 			gitOpsDeployment.Name = binding.Name + "-" + binding.Spec.Components[0].Name
-			err = k8s.Get(&gitOpsDeployment, k8sClient)
-			Expect(err).To(Succeed())
+			Eventually(gitOpsDeployment, "2m", "1s").Should(k8s.ExistByName(k8sClient))
```
anandf, line 421: `gitOpsDeployment` should be changed to `& gitOpsDeployment`. I ran the test with this change and the tests passed after prefixing the `&`. Same change required in `L417` as well.

jgwest:
Yup, that fixed it, thanks for investigating and reviewing @anandf :tada:

PR #338, opened 2023-02-02
Remove Type field from Unit tests and e2e tests

Rizwana777:
#### Description:
Remove Type field from Unit tests and e2e tests as it is not required
Here is the PR in Application API where I have added omitempty for Type field in Environment Spec
https://github.com/redhat-appstudio/application-api/pull/28 

https://issues.redhat.com/browse/GITOPSRVCE-356


jgwest (approved):
LGTM, thanks @Rizwana777!

PR #340, opened 2023-02-06
Security: namespace consistency of OperationCR with GitopsEngineInstance

raghavi101:
#### Description:
Check for cluster-agent to only respond to Operation CRs that are created in the same namespace as the target Argo CD (GitopsEngineInstance)

- With this PR the `cluster-agent` component will not process further any Operation that is not in the namespace consistent with GitopsEngineInstance's namespace. 

- For `backend` component, based on my investigation it looks like the operation CR namespace is being [hardcoded](https://github.com/redhat-appstudio/managed-gitops/blob/main/backend/eventloop/application_event_loop/application_event_runner_deployments.go#L183) to default namespace - `gitops-service-argocd`. Adding a check in this component seems redundant (~can be removed once reviewers agree~). Now removed

- If a user tries to create an OperationCR in any namespace, the cluster-agent component will not process it as long as a database entry is found matching the operationID. Assuming that somehow it is possible to create a database entry for operation, the e2e test in this PR first creates an operationDB entry and then an operationCR. The cluster-agent correctly discards the operation that is not is the namespace as that of GitopsEngineInstane's namespace.

- This PR also includes changes in the unit tests present in managed-gitops/cluster-agent/controllers/managed-gitops/eventloop as part of [398](managed-gitops/cluster-agent/controllers/managed-gitops/eventloop)

https://issues.redhat.com/browse/GITOPSRVCE-362
https://issues.redhat.com/browse/GITOPSRVCE-398


backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
```diff
@@ -291,3 +291,203 @@ var _ = Describe("Application Event Runner Deployments to check SyncPolicy.SyncO
 
 	})
 })
+
```
raghavi101, line 294: This test will be removed/kept based on reviews whether backend needs this test or not

raghavi101:
retest

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -346,6 +346,12 @@ func (a applicationEventLoopRunner_Action) handleNewGitOpsDeplEvent(ctx context.
 
 	ctx = sharedutil.RemoveKCPClusterFromContext(ctx)
 	waitForOperation := !a.testOnlySkipCreateOperation // if it's for a unit test, we don't wait for the operation
+
```
raghavi101, line 349: Backend component additional code check will be removed in next commit

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -346,6 +346,12 @@ func (a applicationEventLoopRunner_Action) handleNewGitOpsDeplEvent(ctx context.
...
+	err = CheckOperationCRNamespace(&operationNamespace, engineInstance)
```
raghavi101, line 350: Backend component additional code check will be removed in next commit

jgwest (changes requested):

backend/eventloop/application_event_loop/application_event_runner_deployments.go
raghavi101: Commit, based on discussion that a code patch which checks operatonCR namespace in **_backend_** is redundant, I would remove all changes that I pushed in this PR in **_backend_** component.

raghavi101: But if you think that the namespace check is required in backend, I would keep the changes and also refine them based on your review comments that I see.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -573,17 +594,11 @@ var _ = Describe("Operation Controller", func() {
 				retry, err := task.PerformTask(ctx)
 				Expect(err).To(BeNil())
 				Expect(retry).To(BeFalse())
-
-				By("If no error was returned, and retry is false, then verify that the 'state' field of the Operation row is Completed")
```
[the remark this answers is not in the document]
raghavi101: I think this part got removed accidentally in the refactoring, adding it back. Thanks for seeing the missing code!

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -616,6 +631,14 @@ var _ = Describe("Operation Controller", func() {
 				By("Close database connection")
 				defer dbQueries.CloseDatabase()
 				defer testTeardown()
+				namespaceToCreate := &corev1.Namespace{
```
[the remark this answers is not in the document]
raghavi101: I tried to move them to BeforEach() but it is breaking some of the tests and I couldn't find a permanent fix to them. I think for now I'll keep them in the respective It() blocks

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1521,7 +1571,124 @@ var _ = Describe("Operation Controller", func() {
 				Expect(isRunning).To(BeFalse())
 			})
 		})
+
+		Context("Operation namespace consistency test", func() {
+
+			It("Operation namespace should match gitopsEngineInstance's namespace", func() {
```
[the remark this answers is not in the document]
raghavi101: Yes there might be tests that check this condition, infact each test does the same process for different cases so I'm fairly positive these two tests can be removed.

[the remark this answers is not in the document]
raghavi101: Eg : [here](https://github.com/redhat-appstudio/managed-gitops/blob/main/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go#L509)

backend/eventloop/db_reconciler_test.go
```diff
@@ -547,7 +557,7 @@ var _ = Describe("DB Reconciler Test", func() {
 				repoCredentialCr = managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredential{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "test-repo-" + string(uuid.NewUUID()),
-						Namespace: "test-k8s-namespace",
+						Namespace: "argocd",
```
jgwest, line 560: Is it necessary to create this in `argocd`? I think it should instead be created in either the old value, or in the namespace defined in `apiNamespace`.

backend/eventloop/db_reconciler_test.go
```diff
@@ -741,7 +751,7 @@ var _ = Describe("DB Reconciler Test", func() {
 				gitopsDeplSyncRunCr = managedgitopsv1alpha1.GitOpsDeploymentSyncRun{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "test-gitopsdeployment-syncrun",
-						Namespace: "test-k8s-namespace",
+						Namespace: "argocd",
```
jgwest, line 754: As above.

backend/eventloop/shared_resource_loop/sharedresourceloop_managedenv_test.go
```diff
@@ -512,7 +512,15 @@ var _ = Describe("SharedResourceEventLoop Test", func() {
...
-			_, _, _, instance, _, err := db.CreateSampleData(dbQueries)
+			_, _, engineCluster, _, _, err := db.CreateSampleData(dbQueries)
+			Expect(err).To(BeNil())
+			instance := &db.GitopsEngineInstance{
+				Gitopsengineinstance_id: "test-fake-instance-id",
+				Namespace_name:          "gitops-service-argocd",
+				Namespace_uid:           "test-fake-instance-namespace-914",
+				EngineCluster_id:        engineCluster.Gitopsenginecluster_id,
+			}
+			err = dbQueries.CreateGitopsEngineInstance(ctx, instance)
```
jgwest, line 523: Possible to use the GitOpsEngineInstance coming from `CreateSampleData`?

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -99,14 +102,20 @@ var _ = Describe("Operation Controller", func() {
 				},
 			}
 
+			namespaceToCreate := &corev1.Namespace{
```
jgwest, line 105: It looks like we're creating an Argo CD namespace here. If you're going to do that, you can remove the existing argocdNamespace field for this
```go
var argocdNamespace *corev1.Namespace
``` 
in this file.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1521,7 +1526,182 @@ var _ = Describe("Operation Controller", func() {
...
+					Spec: managedgitopsv1alpha1.OperationSpec{
+						OperationID: operationDB.Operation_id,
+					},
+				}
+
+				err = task.event.client.Create(ctx, operationCR)
+				Expect(err).To(BeNil())
+
+				retry, err := task.PerformTask(ctx)
+				Expect(err).ToNot(BeNil())
```
jgwest, line 1686: Can we double check here that the error that is returned here, contains `OperationCR namespace did not match with existing namespace of GitopsEngineInstance`? That way we know the test is causing and receiving the correct error.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1521,7 +1526,182 @@ var _ = Describe("Operation Controller", func() {
...
+				Expect(retry).To(BeFalse())
+
+				By("Verifying whether Application CR is created")
```
jgwest, line 1689:
```
By("Verifying whether Application CR is created")
```
Would change this to something like 'verifying the application CR is not created', which I believe is our purpose here

tests-e2e/core/operation_namespace_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+		BeforeEach(func() {
+
+			By("deleting the namespace before the test starts, so that the code can create it")
+			config, err = fixture.GetSystemKubeConfig()
+			if err != nil {
+				panic(err)
+			}
+			config, err = fixture.GetSystemKubeConfig()
+			Expect(err).To(BeNil())
```
jgwest, line 54: I think these two lines are the same...

tests-e2e/core/operation_namespace_test.go
```diff
@@ -0,0 +1,332 @@
...
+			err = dbQueries.CreateOperation(ctx, operationDB, operationDB.Operation_owner_user_id)
+			Expect(err).To(BeNil())
+
+			operationCR = &managedgitopsv1alpha1.Operation{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      operationName,
+					Namespace: namespaceToTarget.Name,
+				},
+				Spec: managedgitopsv1alpha1.OperationSpec{
+					OperationID: "test-operation",
```
jgwest, line 150: Can use `operationDB.Operation_id` here, so that it's obvious what the purpose of this value is, and where this value is coming from.

tests-e2e/core/operation_namespace_test.go
```diff
@@ -0,0 +1,332 @@
...
+					Name:      "operation",
+					Namespace: "my-user",
+				},
+			}
+
+			Eventually(func() bool {
+				isComplete, _ := dboperations.IsOperationComplete(ctx, operationDB, dbQueries)
+				fmt.Println("- Operation state result achieved : ", isComplete)
+				return isComplete
+			}, "1m", "5s").Should(BeFalse())
```
jgwest, line 267: I think this should be a `Consistently`, rather than an `Eventually`. With `Consistently`, we want to keep checking, every 5 seconds, for a minute, to make sure that the Operation is never acted upon.

jgwest (changes requested):
Looks great @raghavi101, just a few minor tweaks I happened to notice while reviewing the updates, then good to go.

backend/eventloop/db_reconciler_test.go
```diff
@@ -451,7 +451,17 @@ var _ = Describe("DB Reconciler Test", func() {
...
+				Expect(err).To(BeNil())
+				_, _, engineCluster, _, _, err := db.CreateSampleData(dbQueries)
+				Expect(err).To(BeNil())
+				gitopsEngineInstance := &db.GitopsEngineInstance{
+					Gitopsengineinstance_id: "test-fake-instance-id",
+					Namespace_name:          "gitops-service-argocd",
+					Namespace_uid:           "test-fake-instance-namespace-914",
+					EngineCluster_id:        engineCluster.Gitopsenginecluster_id,
+				}
+				err = dbQueries.CreateGitopsEngineInstance(ctx, gitopsEngineInstance)
```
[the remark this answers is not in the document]
raghavi101: gitopsEngineInstance here needs to be in the hard coded namespace : gitops-service-argocd ([here](https://github.com/redhat-appstudio/managed-gitops/blob/main/backend/eventloop/application_event_loop/application_event_runner_deployments.go#L184)) or else the test would fail that is why not using the generic function.

backend/eventloop/db_reconciler_test.go
raghavi101: `argocd` is required because we're using 
```
_, _, _, gitopsEngineInstance, _, err := db.CreateSampleData(dbq)
Expect(err).To(BeNil())
```
this will create a gitopsEngineInstance in `argocd` namepsace, I have tried to alter the ns to some other ns for eg. the default ns `gitops-service-argocd` but it clashes with the `argoCDNamespace` creation during fake client setup. So went with `argocd` which isn't used elsewhere to get good validity.

backend/eventloop/shared_resource_loop/sharedresourceloop_managedenv_test.go
raghavi101: same reason as above

raghavi101:
Hi @jgwest PTAL, made all changes

jgwest (approved):
Looks great, thanks again @raghavi101!

PR #341, opened 2023-02-06
Add support for resource cleanup finalizer for GitOpsDeployment 

chetan-rns:
#### Description:
- Adds support for the deletion finalizer `resources-finalizer.managed-gitops.redhat.com`. 
- In the presence of the finalizer, GitOpsDeployment will wait for the deletion of all the child resources. Once all the dependencies are removed, the backend will remove the deletion finalizer and the GitOpsDeployment CR will be deleted.
- Unit tests and e2e tests 

https://issues.redhat.com/browse/GITOPSRVCE-360

### DIFF
--- a/backend-shared/db/utils.go
+++ b/backend-shared/db/utils.go
@@ -460,8 +460,10 @@ func SetupForTestingDBGinkgo() error {
 	for _, managedEnvironment := range managedEnvironments {
 		if strings.HasPrefix(managedEnvironment.Managedenvironment_id, "test-") {
 			rowsAffected, err := dbq.DeleteManagedEnvironmentById(ctx, managedEnvironment.Managedenvironment_id)
-			Expect(rowsAffected).Should(Equal(1))
 			Expect(err).To(BeNil())
+			if err == nil {
+				Expect(rowsAffected).Should(Equal(1))
+			}
 		}
 	}
 
--- a/backend-shared/db/utils.go
+++ b/backend-shared/db/utils.go
@@ -461,9 +461,7 @@ func SetupForTestingDBGinkgo() error {
 		if strings.HasPrefix(managedEnvironment.Managedenvironment_id, "test-") {
 			rowsAffected, err := dbq.DeleteManagedEnvironmentById(ctx, managedEnvironment.Managedenvironment_id)
 			Expect(err).To(BeNil())
-			if err == nil {
-				Expect(rowsAffected).Should(Equal(1))
-			}
+			Expect(rowsAffected).Should(Equal(1))
 		}
 	}
 
--- a/backend/eventloop/application_event_loop/application_event_runner_deployments.go
+++ b/backend/eventloop/application_event_loop/application_event_runner_deployments.go
@@ -42,10 +42,6 @@ const (
 	deploymentModifiedResult_NoChange deploymentModifiedResult = "noChangeInApp"
 
 	prunePropagationPolicy = "PrunePropagationPolicy=background"
-
-	// deletionFinalizer will indicate the GitOpsDeployment to wait until all its dependencies are removed.
-	// In the absence of this finalizer, GitOpsDeployment will be deleted first and its dependencies will be removed in the background.
-	deletionFinalizer = "resources-finalizer.managed-gitops.redhat.com"
 )
 
 // This file is responsible for processing events related to GitOpsDeployment CR.
@@ -419,7 +415,7 @@ func (a applicationEventLoopRunner_Action) handleDeleteGitOpsDeplEvent(ctx conte
 	if allErrors == nil {
 		if gitopsDepl != nil && isGitOpsDeploymentDeleted(gitopsDepl) {
 			// remove the finalizer if all the dependencies are cleaned up
-			if err := removeFinalizerIfExist(ctx, a.workspaceClient, gitopsDepl, deletionFinalizer); err != nil {
+			if err := removeFinalizerIfExist(ctx, a.workspaceClient, gitopsDepl, managedgitopsv1alpha1.DeletionFinalizer); err != nil {
 				a.log.Error(err, "failed to remove the deletion finalizer from GitOpsDeployment", "name", gitopsDepl.Name, "namespace", gitopsDepl.Namespace)
 				return false, gitopserrors.NewDevOnlyError(err)
 			}
--- a/backend/eventloop/application_event_loop/application_event_runner_deployments.go
+++ b/backend/eventloop/application_event_loop/application_event_runner_deployments.go
@@ -413,7 +413,7 @@ func (a applicationEventLoopRunner_Action) handleDeleteGitOpsDeplEvent(ctx conte
 	}
 
 	if allErrors == nil {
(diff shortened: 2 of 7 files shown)


backend/eventloop/application_event_loop/application_event_runner_test.go
```diff
@@ -1689,6 +1716,121 @@ var _ = Describe("ApplicationEventLoop Test", func() {
...
+				}
+			}
+			Expect(gitopsDeplUpdated).To(BeFalse())
+		})
+
+		It("should not update if the GitOpsDeployment is already deleted", func() {
+			err := k8sClient.Delete(ctx, gitopsDepl)
+			Expect(err).To(BeNil())
+
+			err = removeFinalizerIfExist(ctx, k8sClient, gitopsDepl, deletionFinalizer)
```
jopit, line 1796: Even though it's a little strange, the test should probably set the finalizer on the in-memory `gitopsDepl` before calling `removeFinalizerIfExists` (and not call `k8sClient.Update()`).  This will ensure that the reason it doesn't get updated is because it's been deleted, and not because it doesn't have the finalizer set.

jopit (changes requested):

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -617,6 +617,62 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
...
+					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
+					gitopsDeplFixture.HaveResources(expectedResourceStatusList),
+				),
+			)
+			By("delete the GitOpsDeployment resource")
+
+			err = k8sClient.Delete(ctx,&gitOpsDeploymentResource)
+			Expect(err).To(Succeed())
+
+			By("verify if the finalizer has prevented the GitOpsDeployment from deletion")
```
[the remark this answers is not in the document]
chetan-rns: Good catch. I don't know if there is a better way to verify this. But for now, I've added a finalizer to one of the dependent resources to prevent them from being deleted.

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -40,7 +41,11 @@ const (
 	deploymentModifiedResult_Updated  deploymentModifiedResult = "updatedApp"
 	deploymentModifiedResult_NoChange deploymentModifiedResult = "noChangeInApp"
 
-	prunePropagationPolicy = "PrunePropagationPolicy=background"
+	prunePropagationPolicy = "PrunePropagationPolicy=foreground"
```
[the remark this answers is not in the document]
chetan-rns: I've reverted it. I assumed that we had to delete resources in the foreground. But it's not a requirement in the story, so keeping the existing behaviour

jopit (approved):
lgtm

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -412,6 +417,13 @@ func (a applicationEventLoopRunner_Action) handleDeleteGitOpsDeplEvent(ctx conte
 	}
 
 	if allErrors == nil {
+		if gitopsDepl != nil && isGitOpsDeploymentDeleted(gitopsDepl) {
```
jgwest, line 420: The if statement here is actually doing:
`if gitopsDepl != nil && (gitopsDepl == nil || gitopsDepl.DeletionTimestamp != nil)` which seems a bit odd. But it's logically correct, so maybe that's fine :shrug:.

jgwest (changes requested):

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -420,6 +432,54 @@ func (a applicationEventLoopRunner_Action) handleDeleteGitOpsDeplEvent(ctx conte
...
+}
+
+// we consider the GitOpsDeployment as deleted if:
+// 1. the GitOpsDeployment object is not found.
+// 2. the GitOpsDeployment is under deletion and the deletiontimestamp is set.
+func isGitOpsDeploymentDeleted(gitopsDepl *managedgitopsv1alpha1.GitOpsDeployment) bool {
+	return gitopsDepl == nil || gitopsDepl.DeletionTimestamp != nil
+}
+
+func removeItemFromSlice(item string, items []string) []string {
```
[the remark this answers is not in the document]
chetan-rns: Good catch! I forgot that slices are actually pointers to the underlying array and copying a slice and modifying it will update the original. To keep it simple I've updated the function to:
```
func removeItemFromSlice(item string, items []string) []string {
	result := []string{}
	for _, i := range items {
		if i != item {
			result = append(result, i)
		}
	}

	return result
}
```

backend-shared/db/utils.go
```diff
@@ -458,8 +458,10 @@ func SetupForTestingDBGinkgo() error {
 	for _, managedEnvironment := range managedEnvironments {
 		if strings.HasPrefix(managedEnvironment.Managedenvironment_id, "test-") {
 			rowsAffected, err := dbq.DeleteManagedEnvironmentById(ctx, managedEnvironment.Managedenvironment_id)
-			Expect(rowsAffected).Should(Equal(1))
 			Expect(err).To(BeNil())
+			if err == nil {
```
anandf, line 462: Is this `if` check required ? if err was not nil, it would have already failed in the previous line. `Expect(err).To(BeNil())`

chetan-rns: Not required. I've removed it.

backend/eventloop/application_event_loop/application_event_runner_deployments.go
chetan-rns: Updated it

backend-shared/db/utils.go
anandf: Thanks. LGTM

tests-e2e/core/gitopsdeployment_test.go
```diff
@@ -617,6 +618,96 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
...
+
+			Eventually(gitOpsDeploymentResource, ArgoCDReconcileWaitTime, "1s").Should(
+				SatisfyAll(
+					gitopsDeplFixture.HaveSyncStatusCode(managedgitopsv1alpha1.SyncStatusCodeSynced),
+					gitopsDeplFixture.HaveHealthStatusCode(managedgitopsv1alpha1.HeathStatusCodeHealthy),
+					gitopsDeplFixture.HaveResources(expectedResourceStatusList),
+				),
+			)
+
+			updateFinalizer := func(c client.Object, op func(client.Object)) {
```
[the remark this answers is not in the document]
chetan-rns: I've created a separate function `UpdateWithoutConflict` in `k8s/fixture.go`. Let me know if we need to refactor the original `Update()` to handle conflicts.

jgwest (approved):
Looks great, thanks again @chetan-rns, and thanks @jopit and @anandf for reviewing!

PR #342, opened 2023-02-06
Webhook implementation for CRs.

jparsai:
# Description
This PR enables Webhooks which are implemented in `application-api` to avoid changes in immutable fields.

PR to add webhooks in application-api:- https://github.com/redhat-appstudio/application-api/pull/26
**Note**
This PR also removes few existing test cases which are updating Spec fields of CR and now that is not allowed by Webhooks.

# Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-211

### DIFF
--- a/tests-e2e/core/promotionrun_controller_test.go
+++ b/tests-e2e/core/promotionrun_controller_test.go
@@ -1,14 +1,11 @@
 package core
 
 import (
-	"context"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	appstudiocontroller "github.com/redhat-appstudio/managed-gitops/appstudio-controller/controllers/appstudio.redhat.com"
 	"github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture"
 	"github.com/redhat-appstudio/managed-gitops/tests-e2e/fixture/k8s"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appstudiosharedv1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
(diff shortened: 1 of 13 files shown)


jgwest (changes requested):
Looks great, thanks @jparsai!

One request: it couldn't hurt to have some quick E2E tests that test that the webhooks work as expected (similar to the unit tests in application-api):

SnapshotEnvironmentBinding:
- Create a SnapshotEnvironmentBinding
- Try to modify `.spec.application` and/or `.spec.environment`
    -  Should return an error when trying to modify these

Snapshot:
- Create a Snapshot
- Try to modify `.spec.application` and `.spec.components`
    -  Should return an error when trying to modify these, as it is blocked by the webhook

(And for PromotionRun, it looks like we already have tests for this)

tests-e2e/core/managed_environment_test.go
```diff
@@ -218,6 +219,7 @@ var _ = Describe("Environment E2E tests", func() {
 							AllowInsecureSkipTLSVerify: true,
 						},
 					},
+					Type: appstudiosharedv1.EnvironmentType_POC,
```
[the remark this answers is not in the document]
jparsai: Yes, actually this field was mandatory earlier, so I had to add it for E2Es to pass, but now I see Rizwana changed it and now it is not required. I removed it in PR.

jparsai:
> Looks great, thanks @jparsai!
> 
> One request: it couldn't hurt to have some quick E2E tests that test that the webhooks work as expected (similar to the unit tests in application-api):
> 
> SnapshotEnvironmentBinding:
> 
> * Create a SnapshotEnvironmentBinding
> * Try to modify `.spec.application` and/or `.spec.environment`
>   
>   * Should return an error when trying to modify these
> 
> Snapshot:
> 
> * Create a Snapshot
> * Try to modify `.spec.application` and `.spec.components`
>   
>   * Should return an error when trying to modify these, as it is blocked by the webhook
> 
> (And for PromotionRun, it looks like we already have tests for this)

Yes, I do have E2E for those. They all are in promotionrun_controller_test.go, since this file is using Snapshot and SnapshotEnvironmentBinding CRs as well, so I thought of grouping all tests at one place.

Snapshot :- https://github.com/redhat-appstudio/managed-gitops/blob/37a260ea091aa6bacb1ea9c33760dbd3b76fb1a8/tests-e2e/core/promotionrun_controller_test.go#L158

SnapshotEnvironmentBinding :- https://github.com/redhat-appstudio/managed-gitops/blob/37a260ea091aa6bacb1ea9c33760dbd3b76fb1a8/tests-e2e/core/promotionrun_controller_test.go#L192

PromotionRun :- https://github.com/redhat-appstudio/managed-gitops/blob/37a260ea091aa6bacb1ea9c33760dbd3b76fb1a8/tests-e2e/core/promotionrun_controller_test.go#L218

tests-e2e/core/promotionrun_controller_test.go
```diff
@@ -143,6 +147,118 @@ var _ = Describe("Application Promotion Run E2E Tests.", func() {
 			}
 
 			Eventually(promotionRun, "3m", "1s").Should(promotionRunFixture.HaveStatusComplete(expectedPromotionRunStatus))
+
+			envVariable := os.Getenv("DISABLE_APPSTUDIO_WEBHOOK")
+			// Don't run webhook checks if it is disabled.
+			if envVariable != "" && !strings.EqualFold(envVariable, "true") {
+				By("Validate CR Webhooks.")
```
jgwest, line 154: Would it be possible to move the webhook tests into their own file, e.g. `webhook_tests.go`, something like that? In my earlier PR comment, I was confused as I had assumed that the tests in `promotionrun_controller_test.go` were only related to the promotionrun CR. But as you mentioned, this block of tests is testing webhooks for all components.

tests-e2e/go.mod
```diff
@@ -7,11 +7,11 @@ require (
 	github.com/argoproj/argo-cd/v2 v2.5.4
 	github.com/argoproj/gitops-engine v0.7.1-0.20221004132320-98ccd3d43fd9
 	github.com/openshift/api v3.9.1-0.20190916204813-cdbe64fb0c91+incompatible
-	github.com/redhat-appstudio/application-api v0.0.0-20230116140450-3ef8e0cd2c9e
+	github.com/redhat-appstudio/application-api v0.0.0-20230303141937-394e8c127f03
 	github.com/redhat-appstudio/managed-gitops/appstudio-controller v0.0.0
 	github.com/redhat-appstudio/managed-gitops/backend v0.0.0
 	github.com/redhat-appstudio/managed-gitops/backend-shared v0.0.0
-	github.com/redhat-appstudio/managed-gitops/cluster-agent v0.0.0
+	github.com/redhat-appstudio/managed-gitops/cluster-agent v0.0.0-00010101000000-000000000000
```
jgwest, line 14: I think this can just be `github.com/redhat-appstudio/managed-gitops/cluster-agent v0.0.0` (I don't actually know if there is a difference, but I'd prefer to stay on the side of caution. I tested that `go mod tidy` doesn't touch it.)

jgwest (changes requested):

jgwest (approved):
Looks great, thanks @jparsai! It's great to now have proper validation on our API resources :+1:

PR #344, opened 2023-02-08
Add Namepaces to appstudio-controller role

jgwest:
#### Description:
- appstudio-controller requires access to the Namespace K8s resource, and is currently failing due to this lack of access


jgwest:
Merging w/o reviews or CI, since this change is YAML/comments only, and it is blocking deployment to staging cluster.

PR #345, opened 2023-02-09
Update PostgreSQL version

chetan-rns:
#### Description:
- Update the PostgreSQL image version to match the version in RDS


jgwest (approved):
Very quick to review :grin:, thanks @chetan-rns!

PR #349, opened 2023-02-09
Add an exported version of the stonesoup dashboard from Stonesoup Staging cluster

jgwest:
#### Description:
- The current stonesoup dashboard JSON isn't working (for non-obvious reasons), so instead trying a JSON blob exported from the Stonesoup staging cluster.


jgwest:
Dashboard YAML only, merging without tests/reviews.

PR #351, opened 2023-02-10
Remove temporary Grafana dashboard files

jgwest:
#### Description:
- I briefly needed a 'stonesoup-new' directory with a different version of the dashboard. 
- I fixed the issue I was seeing with the 'stonesoup' directory, so the 'stonesoup-new' is no longer needed.


jgwest:
JSON change only, merging with tests/reviews.

PR #354, opened 2023-02-13
Add db-migration tests to run our database migration tests against live data that represent our actual usage

Rizwana777:
#### Description:
- db-migration tests for adding and verifying data in database
- Added Github CI for db-migration-tests

https://issues.redhat.com/browse/GITOPSRVCE-313

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -190,12 +190,12 @@ jobs:
           cd $GITHUB_WORKSPACE/utilities/db-migration/migration_test/add_test_values
           go test $GITHUB_WORKSPACE/utilities/db-migration/migration_test/add_test_values -run "TestInitializeValues"
       
-      - name: "Migrate to latest database version"
+      - name: "Migrate to latest version of database"
         run: |
           cd $GITHUB_WORKSPACE
           make db-migrate
 
-      - name: "Run migration tests to verify data still present in database"
+      - name: "Run migration tests to verify that data added is still present/valid in database"
         run: |
           cd $GITHUB_WORKSPACE/utilities/db-migration/migration_test/verify_test_values
           go test $GITHUB_WORKSPACE/utilities/db-migration/migration_test/verify_test_values -run "TestVerifyDBValues"
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -163,6 +163,42 @@ jobs:
           cd $GITHUB_WORKSPACE/appstudio-controller
           go mod tidy
           git diff --exit-code -- .
+  
+  db-migration-tests:
+    name: Run db-migration tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup Golang
+        uses: actions/setup-go@v3.4.0
+        with:
+          go-version-file: './backend/go.mod'
+      
+      - name: "Start PostgreSQL"
+        run: |
+          cd $GITHUB_WORKSPACE
+          ./create-dev-env.sh
+
+      - name: "Migrate database to version 10"
+        run: |
+          cd $GITHUB_WORKSPACE/utilities/db-migration
+          go run main.go migrate_to 10
+
+      - name: "Run migration tests to add data in database"
+        run: |
+          cd $GITHUB_WORKSPACE/utilities/db-migration/migration_test/add_test_values
+          go test $GITHUB_WORKSPACE/utilities/db-migration/migration_test/add_test_values -run "TestInitializeValues"
+      
+      - name: "Migrate to latest database version"
+        run: |
+          cd $GITHUB_WORKSPACE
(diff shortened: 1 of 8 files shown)


.github/workflows/static.yml
```diff
@@ -163,6 +163,42 @@ jobs:
...
+      
+      - name: "Start PostgreSQL"
+        run: |
+          cd $GITHUB_WORKSPACE
+          ./create-dev-env.sh
+
+      - name: "Migrate database to version 10"
+        run: |
+          cd $GITHUB_WORKSPACE/utilities/db-migration
+          go run main.go migrate_to 10
```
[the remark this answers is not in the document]
Rizwana777: yes right, I will update this

utilities/db-migration/migrate/migrate_db.go
```diff
@@ -61,6 +63,16 @@ func Migrate(opType string, migrationPath string) error {
 			return fmt.Errorf("unable to upgrade migration version by 1 level: %v", err)
 		}
 		return nil
+	} else if opType == "migrate_to" {
+		u64, err := strconv.ParseUint(os.Args[2], 10, 32)
+		if err != nil {
+			fmt.Println(err)
```
chetan-rns, line 69: Looks good. One final question: do we want to return the error if ParseUint fails to parse?

Rizwana777: I think yes, we need to capture all the errors right? as ParseUint function returns error, please let me know if my assumption is wrong

chetan-rns: Yes. Please return the error

chetan-rns (approved):
Thanks @Rizwana777 
/lgtm

jgwest (changes requested):

utilities/db-migration/migration_test/verify_test_values/verify_db_values_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/db"
+	"k8s.io/apimachinery/pkg/util/uuid"
+)
+
+var _ = Describe("Test to verify that the data added to database is still present", func() {
+	Context("Verify Test Values in database", func() {
```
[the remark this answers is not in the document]
Rizwana777: Thanks @jgwest for the review, addressed the comments,, PTAL, I have declared the structs as `var` because when I searched about this I got to know that structs cannot be declared as const

utilities/db-migration/migration_test/add_test_values/add_db_values.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/db"
+)
+
+// To avoid code duplication all the structs are declared as variable which can be acessed in our test files
+var (
+	AddTest_PreClusterUser = db.ClusterUser{
+		Clusteruser_id: "test-user-1",
+		User_name:      "test-user-1",
+	}
+	clusterUser = AddTest_PreClusterUser
```
jgwest, line 13: Looks great! A final tweak, can we remove the lowercase variables from the variable list (e.g. `clusterUser`, `clusterCredentials`, etc), and just reference these variables via the `AddTest_*` version of these tests?

My guess is that you are splitting them in two so that in the add tests you can modify the lowercase version of the variable? (Correct me if I'm wrong)

I think you can instead do this, in the add tests:
```go
			By("Create cluster user")
			clusterUser := AddTest_PreClusterUser
			err = dbq.CreateClusterUser(ctx, &clusterUser)
			Expect(err).To(BeNil())
```

This should allow the test to only modify the local variable data, without worrying about the Create function modifying the constant.

utilities/db-migration/migration_test/verify_test_values/verify_db_values_test.go
```diff
@@ -0,0 +1,186 @@
...
+		})
+		It("Test to verify the data added in database is still present", func() {
+			ctx := context.Background()
+			dbq, err := db.NewUnsafePostgresDBQueries(true, true)
+			Expect(err).To(BeNil())
+			defer dbq.CloseDatabase()
+
+			By("Get cluster user")
+			clusterUser := db.ClusterUser{
+				Clusteruser_id: "test-user-1",
```
jgwest, line 26: For the primary keys in this test, I think you can retrieve these values from the constants in `add_db_values.go`, e.g. using `addtestvalues.AddTest_PreClusterUser.Clusteruser_id` here.

Likewise for the other primary keys in this test.

jgwest (approved):
Looks great, this should keep us honest whenever we make new DB schema changes, thanks again @Rizwana777! And thanks @chetan-rns for reviewing.

PR #357, opened 2023-02-17
KubernetesToDBResourceMapping patch should exit with zero error code if DB has not been initialized

jgwest:
#### Description:
- The Stonesoup E2E tests are currently failing on our latest code drop, as the init container code (introduced to fix a bug on staging as part of https://github.com/redhat-appstudio/managed-gitops/pull/287) is returning an error if the database has not been initialized (which is always the case when the E2E tests first run, as part of a PR)
- We can detect if the database has not been initialized by using the exact error code returned by postgresql
- In this case, the init container could should return 0, and allow the backend controller to initialize the database.

### DIFF
--- a/manifests/overlays/appstudio-staging-cluster/backend-deployment-patch.yaml
+++ b/manifests/overlays/appstudio-staging-cluster/backend-deployment-patch.yaml
@@ -6,21 +6,21 @@ metadata:
 spec:
   template:
     spec:
-      initContainers:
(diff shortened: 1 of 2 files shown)


jgwest:
Merging as this is now YAML only (the code is unused when this YAML is commented out), and this is blocking Stonesoup infra-deployments gitops branch e2e-tests.

PR #358, opened 2023-02-17
Tests-e2e GitOpsDeployment fixture function should update spec, not status

jgwest:
#### Description:
- The `UpdateDeploymentWithFunction` fixture function was incorrectly labeled: it was updating the status, rather than the object (spec)
- The test that was calling it was attempting to update the Spec, and likely made an incorrect assumption based on the function name
- This PR adds both status and non-status versions of the function.


jopit (approved):
lgtm

jopit (approved):
lgtm

PR #359, opened 2023-02-20
cluster-agent should refresh before syncing an Application

chetan-rns:
#### Description:
- cluster-agent will refresh the application by adding an annotation to it.
- Once the refresh is done, the application will be synced
- added unit tests/e2e tests

https://issues.redhat.com/browse/GITOPSRVCE-384

### DIFF
--- a/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
+++ b/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
@@ -574,8 +574,9 @@ func refreshApplication(ctx context.Context, k8sClient client.Client, appName, a
 		if appCR.Annotations == nil {
 			appCR.Annotations = map[string]string{}
 		}
-		if _, ok := appCR.Annotations[appv1.AnnotationKeyRefresh]; !ok {
-			appCR.Annotations[appv1.AnnotationKeyRefresh] = "normal"
+		refreshType, ok := appCR.Annotations[appv1.AnnotationKeyRefresh]
+		if !ok || refreshType != string(appv1.RefreshTypeNormal) {
+			appCR.Annotations[appv1.AnnotationKeyRefresh] = string(appv1.RefreshTypeNormal)
 			return k8sClient.Update(ctx, appCR)
 		}
 		return nil
--- a/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
+++ b/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
@@ -566,6 +566,8 @@ func refreshApplication(ctx context.Context, k8sClient client.Client, appName, a
 		},
 	}
 
+	// Update the application with refresh annotation if it is not present already.
+	// RetryOnConflict retries updating the application if there are conflicts.
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		err := k8sClient.Get(ctx, client.ObjectKeyFromObject(appCR), appCR)
 		if err != nil {
--- a/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
+++ b/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
@@ -21,6 +21,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -540,6 +541,10 @@ func processOperation_SyncOperation(ctx context.Context, dbOperation db.Operatio
 
 	// 3) Process the event, based on whether the SyncOperation is requesting an app sync, or a terminate.
 	if dbSyncOperation.DesiredState == db.SyncOperation_DesiredState_Running {
+		// refresh the Application before syncing to make sure that the latest revision is deployed.
+		if err := opConfig.syncFuncs.refreshApp(ctx, opConfig.eventClient, dbApplication.Name, opConfig.argoCDNamespace.Name); err != nil {
+			return shouldRetryTrue, err
+		}
 
 		return runAppSync(ctx, dbOperation, *dbSyncOperation, &dbApplication, opConfig)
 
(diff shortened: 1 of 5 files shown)


tests-e2e/core/gitopsdeployment_syncrun_test.go
```diff
@@ -470,6 +471,55 @@ var _ = Describe("GitOpsDeploymentSyncRun E2E tests", func() {
 				Expect(app.Status.History[i].Revision).Should(Equal(cr.revision))
 			}
 		})
+
+		It("should refresh before syncing the Application", func() {
```
chetan-rns, line 475: To verify whether Argo CD refreshes the app, I've added the refresh annotation here. Although there is no control over when Argo CD removes the annotation and this might be a race condition, we can still verify if Argo CD behaves as expected. Let me know if there is a better way to test this feature.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1196,6 +1198,32 @@ var _ = Describe("Operation Controller", func() {
...
+							return err
+						}
+
+						if _, ok := appCR.Annotations[appv1.AnnotationKeyRefresh]; ok {
+							delete(appCR.Annotations, appv1.AnnotationKeyRefresh)
+							return k8sClient.Update(ctx, &appCR)
+						}
+						return nil
+					})
+					Expect(err).To(BeNil())
```
jgwest, line 1224: Since this function is called within a goroutine, I think you need  `defer GinkgoRecover()` at the top for Ginkgo to catch this.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1196,6 +1198,32 @@ var _ = Describe("Operation Controller", func() {
 				err = k8sClient.Update(ctx, applicationCR)
 				Expect(err).To(BeNil())
 			}
+			refreshHandler := func(appCR appv1.Application, closeRefreshHandler chan struct{}) {
+				for {
```
jgwest, line 1202: In the success, non-conflict case, this loop will run over and over again, as fast as the CPU will allow (since this is a mock there is no I/O bound): instead, you may want to consider add short delay after each iteration.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1196,6 +1198,32 @@ var _ = Describe("Operation Controller", func() {
 				err = k8sClient.Update(ctx, applicationCR)
 				Expect(err).To(BeNil())
 			}
+			refreshHandler := func(appCR appv1.Application, closeRefreshHandler chan struct{}) {
```
jgwest, line 1201: Can you add a quick comment to this function? It's obvious what the purpose is in the context of this PR, but may not be obvious to a casual observer reading through the test after-the-fact.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go
```diff
@@ -553,6 +558,59 @@ func processOperation_SyncOperation(ctx context.Context, dbOperation db.Operatio
...
 
+func refreshApplication(ctx context.Context, k8sClient client.Client, appName, appNS string) error {
+	appCR := &appv1.Application{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      appName,
+			Namespace: appNS,
+		},
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
jgwest, line 569: We can add a quick comment like that explains what this block of code is doing

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1269,11 +1301,23 @@ var _ = Describe("Operation Controller", func() {
...
+				By("verify if the refresh annotation is removed")
+				Eventually(func() bool {
+					err = k8sClient.Get(ctx, client.ObjectKeyFromObject(applicationCR), applicationCR)
+					if err != nil {
+						GinkgoWriter.Println(err)
+						return false
+					}
+					return applicationCR.Annotations == nil
+				}, "10s", "1s").Should(BeTrue())
+
```
jgwest, line 1320: I think we can also add a check to make sure that the refresh annotation was added by the call to `PerformTask`: for example, have `refreshHandler` set a bool when it detects that refresh annotation was added, then Expect on that bool here.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1298,13 +1342,96 @@ var _ = Describe("Operation Controller", func() {
...
+					appSync: func(ctx context.Context, s1, s2, s3 string, c client.Client, cs *utils.CredentialService, b bool) error {
+						return nil
+					},
+					refreshApp: refreshApplication,
+				}
+				retry, err := task.PerformTask(ctx)
+				Expect(err).Should(BeNil())
+				Expect(retry).To(BeFalse())
+
+				By("verify if the refresh annotation is removed")
```
jgwest, line 1421: It seems like this `Eventually` is testing the goroutine the refreshHandler goroutine that is started in `BeforeEach`, rather than testing the product code itself.

I would think we would want to instead test if the refresh annotation was added.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -1298,13 +1342,96 @@ var _ = Describe("Operation Controller", func() {
...
+				appCRClone := applicationCR.DeepCopy()
+				appCRClone.Annotations = make(map[string]string)
+				appCRClone.Annotations = map[string]string{
+					"key-1": "value-1",
+					"key-2": "value-2",
+				}
+				err = k8sClient.Update(ctx, appCRClone)
+				Expect(err).To(BeNil())
+
+				applicationCR.Annotations = make(map[string]string)
```
jgwest, line 1406: Can you briefly summarize what this part of the test is doing? It seems like it's creating a conflict in order cause the `refreshHandler` above to report and resolve a conflict. But it seems like this is the test testing itself, rather than the test testing the product code.

tests-e2e/core/gitopsdeployment_syncrun_test.go
```diff
@@ -470,6 +471,56 @@ var _ = Describe("GitOpsDeploymentSyncRun E2E tests", func() {
...
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(app), app)
+				if err != nil {
+					return err
+				}
+				app.Annotations = make(map[string]string)
+				app.Annotations[appv1.AnnotationKeyRefresh] = string(appv1.RefreshTypeNormal)
+				return k8sClient.Update(ctx, app)
+			})
+			Expect(err).To(BeNil())
```
jgwest, line 492: Do we need this part of the test? I ask because I think creating the `GitOpsDeploymentSyncRun` below would also cause the `AnnotationKeyRefresh` to be added.

tests-e2e/core/gitopsdeployment_syncrun_test.go
jgwest: Yup, agreed, it's very tough to test as part of an E2E test.

jgwest (changes requested):

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
chetan-rns: I've added a delay. Also, the goroutine will be running only during the lifecycle of a test since we close it in `AfterEach`

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
chetan-rns: The conflict is intentionally introduced to check if it will be handled inside PerformTask while adding the annotation.  I have added comments to make it clear

tests-e2e/core/gitopsdeployment_syncrun_test.go
chetan-rns: I agree. It doesn't make sense and I have removed it from the test.

jgwest (approved):
Perfect, thanks @chetan-rns!

PR #362, opened 2023-02-22
STONEBLD-777 - tekton-ci: update bundle

Michkov:
Switching to new bundle since development of old one was stopped before attempt to migrate to kcp. New bundle is removing dependency on shared secret.

#### Description:


jgwest (approved):
LGTM, thanks @Michkov!

PR #365, opened 2023-02-26
Add logic to update GitopsDeploymentRepositoryCredential CR's status conditions

ishitasequeira:
#### Description:


jgwest (changes requested):
Thanks @ishitasequeira ! In addition to the comments posted to the PR, I noticed that the PR is missing unit/E2E tests. The PR should also contain unit tests that verify the new functionality (for example), and E2E tests as well (implemented in tests-e2e), as noted in the JIRA ticket. You can see some of the existing examples of how we unit test status conditions, for details.

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
```diff
@@ -0,0 +1,592 @@
...
+		}
+		cloneOptions.Auth = privateKey
+	} else {
+		cloneOptions.Auth = &http.BasicAuth{
+			Username: authUsername,
+			Password: authPassword,
+		}
+	}
+
+	if _, err := git.PlainClone(root, false, cloneOptions); err != nil {
```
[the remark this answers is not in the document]
ishitasequeira: I was trying to look for this but could not find a way to check for correct repository credentials without cloning the repo as part of go-git. For using `HEAD()` as well, it requires an instance of the repository cloned using `PlainClone()` or `Clone()` functions.

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
```diff
@@ -0,0 +1,592 @@
...
+			// Something went wrong, retry
+			errMessage = fmt.Errorf("error retrieving secret: %v", err)
+		}
+		repositoryCredentialStatusConditon := &metav1.Condition{
+			Type:    managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredentialConditionErrorOccurred,
+			Reason:  err.Error(),
+			Status:  metav1.ConditionTrue,
+			Message: errMessage.Error(),
+		}
+		UpdateGitopsDeploymentRepositoryCredentialStatus(gitopsDeploymentRepositoryCredentialCR, ctx, apiNamespaceClient, secret, repositoryCredentialStatusConditon, l)
```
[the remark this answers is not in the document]
ishitasequeira: Do we want to return the error we received while updating status here? I think we would want to continue with return the error message saying that we could not fetch the Secret, right?

backend/eventloop/repocred_reconciler.go
```diff
@@ -0,0 +1,156 @@
...
+		r.startTimerForNextCycle()
+	}()
+
+}
+
+// /////////////
+// Reconcile logic for API CR To Database Mapping table and utility functions.
+// This will reconcile repository credential entries from ACTDM table and RepoistoryCredential table
+// /////////////
+func reconcileRepositoryCredentials(ctx context.Context, dbQueries db.DatabaseQueries, client client.Client, k8sClientFactory sharedresourceloop.SRLK8sClientFactory, log logr.Logger) {
```
jgwest, line 66: The unit tests appear not to cover `reconcileRepositoryCredentials`, we should add a unit tests for this

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
```diff
@@ -0,0 +1,663 @@
...
+		// Update the GitOpsDeploymentRepositoryCredential CR
+		if err := client.Status().Update(ctx, repositoryCredential); err != nil {
+			log.Error(err, "updating repository credential CR's status condition")
+		}
+	}
+	return nil
+
+}
+
+func checkForValidRepositoryCredential(repositoryCredential *managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredential, ctx context.Context, secret *corev1.Secret) []*metav1.Condition {
```
jgwest, line 478: `checkForValidRepositoryCredential` is currently missing full test coverage for all of the if-else branches: you can check this by running `make test` in the `backend/` folder, and then doing `Go: Apply Cover Profile', and passing the path to the cover.out file at `backend/cover.out`.

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential_test.go
```diff
@@ -0,0 +1,220 @@
...
+					if !conditionExists {
+						fmt.Println("GitOpsDeploymentCondition:", conditionExists, "/ Expected:", expectedRepoCredStatus.Conditions, "/ Actual:", repoCred.Status.Conditions)
+						break
+					}
+				}
+				return conditionExists
+
+			}, BeTrue())
+		}
+
```
jgwest, line 137: We should also test the actual `UpdateGitopsDeploymentRepositoryCredentialStatus` functionality itself, for example, creating a `GitOpsDeploymentRepositoryCredential` with a valid/invalid URL, and making sure the conditions are set properly.

jgwest (changes requested):

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential_test.go
ishitasequeira: I have added this as part of e2e tests.

backend/eventloop/repocred_reconciler_test.go
```diff
@@ -0,0 +1,118 @@
...
+				NamespaceUID:         "test-" + string(uuid.NewUUID()),
+				DBRelationType:       db.APICRToDatabaseMapping_DBRelationType_RepositoryCredential,
+				DBRelationKey:        gitopsRepositoryCredentialsDb.RepositoryCredentialsID,
+			}
+
+			err = dbq.CreateAPICRToDatabaseMapping(ctx, &apiCRToDatabaseMappingDb)
+			Expect(err).To(BeNil())
+		})
+
+		It("It should reconcile status for RepositoryCredentials if the entry is present in RepositoryCredentials DB", func() {
```
jgwest, line 95: `It("It should"` -> `It("should`

backend/eventloop/repocred_reconciler_test.go
```diff
@@ -0,0 +1,118 @@
...
+			reconcileRepositoryCredentials(ctx, dbq, k8sClient, log)
+
+			By("Verify that status is updated for GitopsDeploymentRepositoryCredentialCR.")
+			objectMeta := metav1.ObjectMeta{
+				Name:      apiCRToDatabaseMappingDb.APIResourceName,
+				Namespace: apiCRToDatabaseMappingDb.APIResourceNamespace,
+			}
+			repoCredCR := managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredential{ObjectMeta: objectMeta}
+
+			if isOrphaned := isRowOrphaned(ctx, k8sClient, &apiCRToDatabaseMappingDb, &repoCredCR, log); isOrphaned {
```
jgwest, line 110: I believe this check should be removed. Based on code coverage data, it appears that these lines are causing the `Expect(...)` calls below not to run.

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential_test.go
jgwest: That's true: I think we can just query Git (e.g. GitHub) directly, as part of the test, rather than mocking it. If that causes us problems later (for example, due to rate limiting), we can re-examine again.

backend-shared/apis/managed-gitops/v1alpha1/gitopsdeploymentrepositorycredential_types.go
```diff
@@ -63,3 +75,52 @@ type GitOpsDeploymentRepositoryCredentialList struct {
...
+	RepositoryCredentialReasonInvalidCredentials   = "InvalidCredentials"
+	RepositoryCredentialReasonInValidRepositoryUrl = "InvalidRepositoryUrl"
+	RepositoryCredentialReasonValidRepositoryUrl   = "ValidRepositoryUrl"
+)
+
+// SetConditions updates the GitOpsDeploymentRepositoryCredential status conditions for a subset of evaluated types.
+// If the GitOpsDeploymentRepositoryCredential has a pre-existing condition of a type that is not in the evaluated list,
+// it will be preserved. If the GitOpsDeploymentRepositoryCredential has a pre-existing condition of a type, status, reason that
+// is in the evaluated list, but not in the incoming conditions list, it will be removed.
+func (status *GitOpsDeploymentRepositoryCredentialStatus) SetConditions(conditions []*metav1.Condition) {
```
jgwest, line 93: Likewise, `[]*metav1.Condition)` -> `[]metav1.Condition`

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
```diff
@@ -0,0 +1,653 @@
...
+				errorOccuredCondition = &metav1.Condition{
+					Type:    managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredentialConditionErrorOccurred,
+					Reason:  managedgitopsv1alpha1.RepositoryCredentialReasonInvalidCredentials,
+					Status:  metav1.ConditionTrue,
+					Message: fmt.Sprintf("Repository Credentials provided %s for Repository %s are invalid", secret.Name, repositoryCredential.Spec.Repository),
+				}
+			}
+		} else {
+			// No errors occured
+			errorOccuredCondition = &metav1.Condition{
```
jgwest, line 549: This part of the function appears not to be covered by unit tests (correct me if I am wrong!)

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
```diff
@@ -0,0 +1,653 @@
...
+			validRepoCredCondition = &metav1.Condition{
+				Type:    managedgitopsv1alpha1.GitOpsDeploymentRepositoryCredentialConditionValidRepositoryCredential,
+				Reason:  managedgitopsv1alpha1.RepositoryCredentialReasonCredentialsUpToDate,
+				Status:  metav1.ConditionTrue,
+				Message: fmt.Sprintf("Repository Credentials provided %s for Repository %s are valid", secret.Name, repositoryCredential.Spec.Repository),
+			}
+		}
+	}
+
+	return []*metav1.Condition{errorOccuredCondition, validRepoUrlCondition, validRepoCredCondition}
```
jgwest, line 570: `[]*metav1.Condition` -> `[]metav1.Condition` as above

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential_test.go
```diff
@@ -0,0 +1,283 @@
...
+				},
+			}
+
+			err := UpdateGitopsDeploymentRepositoryCredentialStatus(ctx, gitopsDeploymentRepositoryCredentialCR, k8sClient, &corev1.Secret{}, log.FromContext(ctx))
+			Expect(err).To(BeNil())
+
+			Expect(gitopsDeploymentRepositoryCredentialCR).Should(SatisfyAll(haveErrOccurredConditionSet(expectedRepoCredStatus)))
+		})
+
+		It("shouldn't update an existing condition if it hasn't changed", func() {
```
jgwest, line 177: In order to tell if the condition has changed, we can check using lastTransitionTime, and verify it was not updated for any of the conditions.

backend-shared/apis/managed-gitops/v1alpha1/gitopsdeploymentrepositorycredential_types.go
```diff
@@ -63,3 +75,52 @@ type GitOpsDeploymentRepositoryCredentialList struct {
...
+	repoCredConditions := make([]*metav1.Condition, 0)
+	now := metav1.Now()
+	for i := range conditions {
+		condition := conditions[i]
+		if condition.LastTransitionTime.IsZero() {
+			condition.LastTransitionTime = now
```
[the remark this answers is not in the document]
jgwest: We should also verify that `LastCondifionTime` is set, via unit tests. I didn't see any unit tests that were checking this field.

jgwest (changes requested):

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential_test.go
ishitasequeira: The problem with using Git tests as part of unit tests is that we would need to store username/password as well for testing the success scenario. I have added tests for wrong secrets, but I don't think we can test success scenario as we would need to store real username and password for the test.

backend/eventloop/shared_resource_loop/sharedresourceloop_repositorycredential.go
ishitasequeira: This part is not covered as it needs the Git commands to succeed with real secrets

backend-shared/apis/managed-gitops/v1alpha1/gitopsdeploymentrepositorycredential_types.go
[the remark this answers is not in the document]
ishitasequeira: The default time for `LastConditionTime` is zero. We create new conditions when we call this function thus setting the lastTransitionTime as zero. So, we update the value only if for the new incoming condition. And we set the new condition as status below if there was any change in the condition.

PR #368, opened 2023-03-02
Periodic clean-up for orphaned DB entries.

jparsai:
## Description
This PR is to add a logic which will periodically iterate through `Applications` table and delete entry if it does not have entry in `DeploymentToApplicationMapping`, same is for `ManagedEnvironment`, `RepositoryCredentials`, `SyncOperation` if they dont have entry in `APICRToDatabaseMapping`

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-350

## Type of changes

- Introduced new functions for `ManagedEnvironment`, `RepositoryCredentials`, `SyncOperation` to read data in batches.
- Added logic to go through `Application`, `RepositoryCredential`, `SyncOperation`, `ManagedEnvironment` tables and delete orphaned entries.
- Grouped functions according to their use cases.
- Introduces unit test cases for new logic.
- Renamed few functions
```
deplToAppMappingDbReconcile -> cleanOrphanedEntriesfromTable_DTAM

cleanDeplToAppMappingFromDb -> cleanOrphanedEntriesfromTable_DTAM_DeleteEntry

apiCRToDbMappingDbReconcile -> cleanOrphanedEntriesfromTable_ACTDM

apiCRToDbMappingDbReconcile_ManagedEnvironment -> cleanOrphanedEntriesfromTable_ACTDM_ManagedEnvironment

apiCRToDBMappingDbReconcile_RepositoryCredential -> cleanOrphanedEntriesfromTable_ACTDM_RepositoryCredential

apiCRToDBMappingDBReconcile_GitOpsDeploymentSyncRun -> cleanOrphanedEntriesfromTable_ACTDM_GitOpsDeploymentSyncRun

cleanRowFromDB -> deleteDbEntry
```

### DIFF
--- a/backend-shared/db/managedenvironment.go
+++ b/backend-shared/db/managedenvironment.go
@@ -248,3 +248,15 @@ func (obj *ManagedEnvironment) GetAsLogKeyValues() []interface{} {
 		"managedEnvironmentID", obj.Managedenvironment_id,
 		"managedEnvironmentName", obj.Name}
 }
+
+// Get ManagedEnvironments in a batch. Batch size defined by 'limit' and starting point of batch is defined by 'offSet'.
+// For example if you want ManagedEnvironments starting from 51-150 then set the limit to 100 and offset to 50.
+func (dbq *PostgreSQLDatabaseQueries) GetManagedEnvironmentBatch(ctx context.Context, managedEnvironments *[]ManagedEnvironment, limit, offSet int) error {
+	return dbq.dbConnection.
+		Model(managedEnvironments).
+		Order("seq_id ASC").
+		Limit(limit).   // Batch size
+		Offset(offSet). // offset+1 is starting point of batch
+		Context(ctx).
+		Select()
+}
--- a/backend/eventloop/db_reconciler.go
+++ b/backend/eventloop/db_reconciler.go
@@ -6,13 +6,16 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
 	db "github.com/redhat-appstudio/managed-gitops/backend-shared/db"
 	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
+	"github.com/redhat-appstudio/managed-gitops/backend-shared/util/fauxargocd"
 	"github.com/redhat-appstudio/managed-gitops/backend-shared/util/operations"
 	sharedresourceloop "github.com/redhat-appstudio/managed-gitops/backend/eventloop/shared_resource_loop"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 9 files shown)


backend/eventloop/db_reconciler.go
```diff
@@ -444,6 +469,360 @@ func cleanRowFromDB(ctx context.Context, dbQueries db.DatabaseQueries, id string
...
+	crIdMap := make(map[string][]string)
+
+	// Continuously iterate and fetch batches until all entries of table processed.
+	for {
+		if offSet != 0 {
+			time.Sleep(sleepIntervalsOfBatches)
+		}
+
+		// If resource type is Application then get list of Application IDs from DTAM table.
+		if tableType == "Application" {
```
chetan-rns, line 767: Can we use the `dbTableName` type?
```suggestion
		if tableType == dbType_Application {
```

backend/eventloop/db_reconciler.go
```diff
@@ -444,6 +469,360 @@ func cleanRowFromDB(ctx context.Context, dbQueries db.DatabaseQueries, id string
...
+			}
+		}
+
+		// Skip processed entries in next iteration
+		offSet += appRowBatchSize
+	}
+}
+
+// getListOfCRIdsFromTable loops through DTAMs in database and returns list of resource IDs for each CR type (i.e. RepositoryCredential, ManagedEnvironment, SyncOperation) .
+func getListOfCRIdsFromTable(ctx context.Context, dbQueries db.DatabaseQueries, tableType string, log logr.Logger) map[string][]string {
```
chetan-rns, line 753: I understand we are prefetching all the CR IDs to verify whether there is a corresponding entry in the DTAM/ACTDM table. But they are retrieved in batches with a sleep interval and I wonder if this would make it slower than directly checking if an individual entry exists in the table using something like `GetDeploymentToApplicationMappingByApplicationId`. WDYT?

backend-shared/db/managedenvironment.go
```diff
@@ -248,3 +248,20 @@ func (obj *ManagedEnvironment) GetAsLogKeyValues() []interface{} {
 		"managedEnvironmentID", obj.Managedenvironment_id,
 		"managedEnvironmentName", obj.Name}
 }
+
+// Get GetManagedEnvironmentBatch in a batch. Batch size defined by 'limit' and starting point of batch is defined by 'offSet'.
+// For example if you want applications starting from 51-150 then set the limit to 100 and offset to 50.
```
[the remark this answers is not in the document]
jparsai: Ah, did silly mistake in copy paste.

backend/eventloop/db_reconciler.go
```diff
@@ -6,36 +6,40 @@ import (
...
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	appRowBatchSize            = 50               // Number of rows needs to be fetched in each batch.
+	appRowBatchSize            = 100              // Number of rows needs to be fetched in each batch.
 	databaseReconcilerInterval = 30 * time.Minute // Interval in Minutes to reconcile Database.
 	sleepIntervalsOfBatches    = 1 * time.Second  // Interval in Millisecond between each batch.
+	waitTimeforRowDelete       = 1                // Number of hours to wait before deleting DB row
```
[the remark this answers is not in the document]
jparsai: Yes, I mentioned unit in comment, but this is also Okay, will mention unit here.

backend/eventloop/db_reconciler.go
jparsai: Yes, it can be done, but in that case there will be many DB calls, that is what we want to avoid and it is deleting entries which are older than 1 Hour, it means there is enough time for components to update DB entries for CRs and there should not be any chance of race conditions.

backend/eventloop/db_reconciler.go
jparsai: Yes, initially I used this only, but we don't have any `dbTableName` for the else block, which will execute for `RepositoryCredential`, `ManagedEnvironment`, `SyncOperation` tables. I wanted to use a better name but couldn't come up with one, thought about `dbType_Multiple dbTableName = "Multiple"` but didn't look good to me.
https://github.com/redhat-appstudio/managed-gitops/blob/b04b7dac2c1c5545f4ca9d1ff7578576a2dc20ae/backend/eventloop/db_reconciler.go#L787

backend/eventloop/db_reconciler.go
```diff
@@ -444,6 +469,360 @@ func cleanRowFromDB(ctx context.Context, dbQueries db.DatabaseQueries, id string
...
+	}
+}
+
+// getListOfCRIdsFromTable loops through DTAMs in database and returns list of resource IDs for each CR type (i.e. RepositoryCredential, ManagedEnvironment, SyncOperation) .
+func getListOfCRIdsFromTable(ctx context.Context, dbQueries db.DatabaseQueries, tableType string, log logr.Logger) map[string][]string {
+
+	offSet := 0
+
+	// Create Map to store resource IDs according to type, Ex: {"RepositoryCredential" : [], "ManagedEnvironment" : [], "SyncOperation" : []}
+	crIdMap := make(map[string][]string)
```
chetan-rns, line 758: We are converting the `dbTableName` type to `string` to use this map. Can we change the key type to `dbTableName` so we don't have to convert the type?

backend/eventloop/db_reconciler.go
jparsai: Actually there are few more clean jobs I have to implement, so I was thinking of using same function and in that case there would be more `if..else` blocks will be added, but lets change it when new logic is added for now I removed 2nd if condition.

chetan-rns (approved):
Thanks, @jparsai 

/lgtm

jgwest (changes requested):
Thanks @jparsai for the awesome, very detailed, very thorough job of refactoring the existing periodic cleanup logic/tests to reduce code duplication, and for the new tests. Really nice job :tada:! Just some minor opinionated tweaks, then good to go!

jgwest (approved):
LGTM, thanks again @jparsai, and thanks for reviewing @chetan-rns

PR #371, opened 2023-03-02
Add Argo CD application status to gitops dashboard

jopit:
#### Description:
- Added panels to display the health and sync status of Argo CD applications managed by gitops service.


jgwest (approved):
Looks great, thanks @jopit! And thanks for your patience in the time it took for review

PR #372, opened 2023-03-02
Add sync/health/commit ID to SnapshotEnvironmentBinding (#199)

keithchong:
#### Description:
- No new e2e tests were added to test this, but rather, existing e2e tests were modified to take into account the additional fields.
- Please double-check if all the files that are needed to be modified are made when pulling in a new application-api version

See Jira

See https://issues.redhat.com/browse/GITOPSRVCE-199

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
@@ -163,7 +163,7 @@ func (r *SnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, re
 		}
 	}
 
-	statusField := []appstudioshared.BindingStatusGitOpsDeployment{}
+	var statusField []appstudioshared.BindingStatusGitOpsDeployment
 	var allErrors error
 
 	// For each deployment, check if it exists, and if it has the expected content.
@@ -183,11 +183,22 @@ func (r *SnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, re
 				allErrors = fmt.Errorf("%s.\n%s, error: %w", allErrors.Error(), errorMessage, err)
 			}
 		} else {
-			// No error: add to status
-			statusField = append(statusField, appstudioshared.BindingStatusGitOpsDeployment{
+			// If no error, provide status
+			deployment := apibackend.GitOpsDeployment{}
+			newStatusEntry := appstudioshared.BindingStatusGitOpsDeployment{
 				ComponentName:    componentName,
 				GitOpsDeployment: expectedGitOpsDeployment.Name,
-			})
+			}
+			// gosec check doesn't allow taking the address of a loop variable. Here, reassign the loop variable.
+			expectedGitOpsDeployment := expectedGitOpsDeployment
+			if err := rClient.Get(ctx, client.ObjectKeyFromObject(&expectedGitOpsDeployment), &deployment); err == nil {
+				newStatusEntry.GitOpsDeploymentSyncStatus = string(deployment.Status.Sync.Status)
+				newStatusEntry.GitOpsDeploymentHealthStatus = string(deployment.Status.Health.Status)
+				newStatusEntry.GitOpsDeploymentCommitID = deployment.Status.Sync.Revision
+			} else {
+				log.Error(err, "unable to get the deployment for "+componentName)
+			}
+			statusField = append(statusField, newStatusEntry)
 		}
 	}
 
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
@@ -758,6 +758,81 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler Tests", func() {
 			Expect(binding.Status.ComponentDeploymentConditions[0].Message).To(Equal("1 of 2 components deployed"))
 		})
 
+		// Similar to previous test except to test the SnapshotEnvironmentBinding's statuses of its GitOpsDeployments
+		It("sets the status fields of the GitOpsDeployments of a SnapshotEnvironmentBinding", func() {
+			By("Creating SnapshotEnvironmentBinding CR in cluster.")
+			err := bindingReconciler.Create(ctx, binding)
+			Expect(err).To(BeNil())
+
+			By("Checking status field before calling Reconciler")
+			binding = &appstudiosharedv1.SnapshotEnvironmentBinding{}
+			err = bindingReconciler.Get(ctx, request.NamespacedName, binding)
+			Expect(err).To(BeNil())
+			Expect(len(binding.Status.ComponentDeploymentConditions)).To(Equal(0))
+
+			By("Triggering Reconciler to create the GitOpsDeployments")
+			_, err = bindingReconciler.Reconcile(ctx, request)
+			Expect(err).To(BeNil())
+			binding = &appstudiosharedv1.SnapshotEnvironmentBinding{}
+			err = bindingReconciler.Get(ctx, request.NamespacedName, binding)
+			Expect(err).To(BeNil())
+
+			By("Updating the first GitOpsDeployment to simulate a successful sync")
+			deployment := &apibackend.GitOpsDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      binding.Status.GitOpsDeployments[0].GitOpsDeployment,
+					Namespace: binding.Namespace,
+				},
+			}
+			err = bindingReconciler.Get(ctx, client.ObjectKeyFromObject(deployment), deployment)
+			Expect(err).To(BeNil())
+			deployment1Revision := "1.0.0"
+			deployment.Status.Sync.Status = apibackend.SyncStatusCodeSynced
+			deployment.Status.Health.Status = apibackend.HeathStatusCodeHealthy
+			deployment.Status.Sync.Revision = deployment1Revision
+			err = bindingReconciler.Status().Update(ctx, deployment)
+			Expect(err).To(BeNil())
+
+			By("Updating the second GitOpsDeployment to simulate a failed sync")
+			deployment = &apibackend.GitOpsDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      binding.Status.GitOpsDeployments[1].GitOpsDeployment,
+					Namespace: binding.Namespace,
+				},
+			}
+			err = bindingReconciler.Get(ctx, client.ObjectKeyFromObject(deployment), deployment)
+			Expect(err).To(BeNil())
+			deployment2Revision := "2.0.0"
+			deployment.Status.Sync.Status = apibackend.SyncStatusCodeOutOfSync
+			deployment.Status.Health.Status = apibackend.HeathStatusCodeDegraded
+			deployment.Status.Sync.Revision = deployment2Revision
+			err = bindingReconciler.Status().Update(ctx, deployment)
+			Expect(err).To(BeNil())
+
+			By("Triggering Reconciler to update the status field")
+			_, err = bindingReconciler.Reconcile(ctx, request)
+			Expect(err).To(BeNil())
+
+			By("Checking status field after calling Reconciler")
+			binding = &appstudiosharedv1.SnapshotEnvironmentBinding{}
+			err = bindingReconciler.Get(ctx, request.NamespacedName, binding)
+			Expect(err).To(BeNil())
+			Expect(len(binding.Status.GitOpsDeployments)).To(Equal(2))
+			// Order of components not predictable, so just check the status for the specific name
+			for _, deployment := range binding.Status.GitOpsDeployments {
+				// For test purposes, we can compare exactly the Commit IDs, although in the e2e tests, it is not deterministic
+				if deployment.ComponentName == "component-a" {
+					Expect(deployment.GitOpsDeploymentSyncStatus).To(Equal(string(apibackend.SyncStatusCodeSynced)))
+					Expect(deployment.GitOpsDeploymentHealthStatus).To(Equal(string(apibackend.HeathStatusCodeHealthy)))
+					Expect(deployment.GitOpsDeploymentCommitID).To(Equal(deployment1Revision))
+				} else {
+					Expect(deployment.GitOpsDeploymentSyncStatus).To(Equal(string(apibackend.SyncStatusCodeOutOfSync)))
+					Expect(deployment.GitOpsDeploymentHealthStatus).To(Equal(string(apibackend.HeathStatusCodeDegraded)))
+					Expect(deployment.GitOpsDeploymentCommitID).To(Equal(deployment2Revision))
+				}
+			}
+		})
+
 		It("sets the binding's ComponentDeploymentConditions status field when all components deployed successfully.", func() {
 			By("Creating SnapshotEnvironmentBinding CR in cluster.")
 			err := bindingReconciler.Create(ctx, binding)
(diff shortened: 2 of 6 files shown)


keithchong:
Here is an example e2e test failure output when the expected status is not the same as the actual. The CommitID comparison test is not matched exactly.  The test only checks the length.

```
    HaveStatusGitOpsDeploymentsWithProperties: no match for {component-a appa-staging-binding-new-demo-app-staging-component-a   } 
     - Expected number of deployments:  1
                 GitOpsDeployment:  appa-staging-binding-new-demo-app-staging-component-a 
                 GitOpsDeploymentSyncStatus:   Synced
                 GitOpsDeploymentHealthStatus:   Healthy
                 GitOpsDeploymentCommitID:  CurrentlyIDIsUnknownInTestcase
    does not match with the following deployments (not in any specific order):
     - Actual number of deployments:  1
                 GitOpsDeployment:  appa-staging-binding-new-demo-app-staging-component-a 
                 GitOpsDeploymentSyncStatus:  OutOfSync 
                 GitOpsDeploymentHealthStatus:  Progressing 
                 GitOpsDeploymentCommitID:  9e7867d6274bb59a93956e29b1dce0ce14de8a8d

```

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
```diff
@@ -183,11 +183,21 @@ func (r *SnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, re
...
-			// No error: add to status
-			statusField = append(statusField, appstudioshared.BindingStatusGitOpsDeployment{
+			// If no error, provide status
+			deployment := apibackend.GitOpsDeployment{}
+			newStatusEntry := appstudioshared.BindingStatusGitOpsDeployment{
 				ComponentName:    componentName,
 				GitOpsDeployment: expectedGitOpsDeployment.Name,
-			})
+			}
+			expectedGitOpsDeployment := expectedGitOpsDeployment
```
jparsai, line 192: Here you are assigning to same variable.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
```diff
@@ -758,6 +758,81 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler Tests", func() {
...
 
+		// Similar to previous test except to test the SnapshotEnvironmentBinding's statuses of its GitOpsDeployments
+		It("sets the status fields of the GitOpsDeployments of a SnapshotEnvironmentBinding", func() {
+			By("Creating SnapshotEnvironmentBinding CR in cluster.")
+			err := bindingReconciler.Create(ctx, binding)
+			Expect(err).To(BeNil())
+
+			By("Checking status field before calling Reconciler")
+			binding = &appstudiosharedv1.SnapshotEnvironmentBinding{}
+			err = bindingReconciler.Client.Get(ctx, request.NamespacedName, binding)
```
jparsai, line 769: Nit, it can also be `bindingReconciler.Get`

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
keithchong: Hi @jparsai , originally, I didn't have this and the code worked. However, the gosec check failed with the error: implicit memory aliasing in for loop.  It's complaining that I'm using the address of an iteration variable. There are several ways to get around this, and this is one way.  

Another way could be instead of 192:
			expectedDep := expectedDeployments[componentName]
			if err := rClient.Get(ctx, client.ObjectKeyFromObject(&expectedDep), &deployment); err == nil {


I will add a comment at line 191 to explain this when I rebase this PR.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
keithchong: Done.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
keithchong: Added the comment.

keithchong:
The go.mod and go.sum changes in my original branch are no longer needed since another PR that was just merged now includes the application-api version with my changes. Also, this [kustomization.yaml ](https://github.com/redhat-appstudio/managed-gitops/blob/main/manifests/base/crd/overlays/local-dev/kustomization.yaml) has already been updated too. ID: 394e8c127f031990838397fcb2d1e10110b2e7ae 

The only other change left I believe is the Makefile change which is included in this PR.

keithchong:
The unit test failed in the PR check, but when I ran it locally, it passed.  After running it several times, I could finally reproduce the error...so it is random.  The latest commit updates the unit test to take into account the random order of the binding's deployments after reconciliation.

jparsai:
Thanks @keithchong 

/lgtm

keithchong:
Thanks @jparsai .  FYI, here was the error.  The INFO message showed that the statuses and the commit ID were assigned to the wrong component.  eg. component-a had OutOfSync and commit ID of 2.0.0 and component-b had Synced and commit ID of 1.0.0.  It is opposite to what was set in the test case.


```
STEP: Updating the first GitOpsDeployment to simulate a successful sync 03/08/23 18:01:34.036
[61](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:62)
    STEP: Updating the second GitOpsDeployment to simulate a failed sync 03/08/23 18:01:34.036
[62](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:63)
    STEP: Triggering Reconciler to update the status field 03/08/23 18:01:34.036

    1.678298494036475e+09	INFO	API Resource changed: Modified	{"name": "appa-staging-binding", "namespace": "my-user", "audit": "true", "namespace": "my-user", "name": "appa-staging-binding", "object": "{\"kind\":\"SnapshotEnvironmentBinding\",\"apiVersion\":\"appstudio.redhat.com/v1alpha1\",\"metadata\":{\"name\":\"appa-staging-binding\",\"namespace\":\"my-user\",\"resourceVersion\":\"3\",\"creationTimestamp\":null,\"labels\":{\"appstudio.application\":\"new-demo-app\",\"appstudio.environment\":\"staging\"}},\"spec\":{\"application\":\"new-demo-app\",\"environment\":\"staging\",\"snapshot\":\"my-snapshot\",\"components\":[{\"name\":\"component-a\",\"configuration\":{\"replicas\":3,\"env\":[{\"name\":\"My_STG_ENV\",\"value\":\"1000\"}]}},{\"name\":\"component-b\",\"configuration\":{\"replicas\":3,\"env\":[{\"name\":\"My_STG_ENV\",\"value\":\"1000\"}]}}]},\"status\":{\"gitopsDeployments\":[{\"componentName\":\"component-a\",\"gitopsDeployment\":\"appa-staging-binding-new-demo-app-staging-component-a\",\"syncStatus\":\"OutOfSync\",\"health\":\"Degraded\",\"commitID\":\"2.0.0\"},{\"componentName\":\"component-b\",\"gitopsDeployment\":\"appa-staging-binding-new-demo-app-staging-component-b\",\"syncStatus\":\"Synced\",\"health\":\"Healthy\",\"commitID\":\"1.0.0\"}],\"components\":[{\"name\":\"component-a\",\"gitopsRepository\":{\"url\":\"[https://github.com/redhat-appstudio/managed-gitops\](https://github.com/redhat-appstudio/managed-gitops/)",\"branch\":\"main\",\"path\":\"resources/test-data/sample-gitops-repository/components/componentA/overlays/staging\",\"generatedResources\":null,\"commitID\":\"\"}},{\"name\":\"component-b\",\"gitopsRepository\":{\"url\":\"[https://github.com/redhat-appstudio/managed-gitops\](https://github.com/redhat-appstudio/managed-gitops/)",\"branch\":\"main\",\"path\":\"resources/test-data/sample-gitops-repository/components/componentB/overlays/staging\",\"generatedResources\":null,\"commitID\":\"\"}}],\"componentDeploymentConditions\":[{\"type\":\"AllComponentsDeployed\",\"status\":\"False\",\"lastTransitionTime\":\"2023-03-08T18:01:34Z\",\"reason\":\"CommitsUnsynced\",\"message\":\"1 of 2 components deployed\"}]}}"}
[64](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:65)
    1.678298494036493e+09	DEBUG	Snapshot Environment Binding Reconcile() complete.	{"name": "appa-staging-binding", "namespace": "my-user"}
[65](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:66)
    STEP: Checking status field after calling Reconciler 03/08/23 18:01:34.036
[66](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:67)
  << End Captured GinkgoWriter Output
[67](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:68)

[68](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:69)
  Expected
[69](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:70)
      <string>: OutOfSync
[70](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:71)
  to equal
[71](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:72)
      <string>: Synced
[72](https://github.com/redhat-appstudio/managed-gitops/actions/runs/4367057552/jobs/7637876310#step:8:73)
  In [It] at: /home/runner/work/managed-gitops/managed-gitops/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go:825
```

tests-e2e/fixture/binding/fixture.go
```diff
@@ -105,6 +105,81 @@ func HaveStatusGitOpsDeployments(gitOpsDeployments []appstudiosharedv1.BindingSt
...
+		}
+
+		for _, aDeployment := range a { // Expected
+			match := false
+			for _, bDeployment := range b { // Actual
+				// Compare exact strings but only check the length of the Commit ID since it is not deterministic.
+				if aDeployment.ComponentName == bDeployment.ComponentName && aDeployment.GitOpsDeployment == bDeployment.GitOpsDeployment &&
+					aDeployment.GitOpsDeploymentHealthStatus == bDeployment.GitOpsDeploymentHealthStatus &&
+					aDeployment.GitOpsDeploymentSyncStatus == bDeployment.GitOpsDeploymentSyncStatus &&
+					len(bDeployment.GitOpsDeploymentCommitID) > 0 && len(bDeployment.GitOpsDeploymentCommitID) <= 40 {
```
jgwest, line 123: :+1:

tests-e2e/fixture/binding/fixture.go
```diff
@@ -105,6 +105,81 @@ func HaveStatusGitOpsDeployments(gitOpsDeployments []appstudiosharedv1.BindingSt
 	}, BeTrue())
 }
 
+func HaveGitOpsDeploymentsWithStatusProperties(gitOpsDeployments []appstudiosharedv1.BindingStatusGitOpsDeployment) matcher.GomegaMatcher {
+
+	// compare compares two slices, returning true if the fields other than the CommitID are equal, regardless of the order of elements in the slices
```
jgwest, line 110: FYI, in some cases we may be able to use `ContainElements` function from Gomega (e.g. `Expect(a).ToContainElements(b)`). This function will compare the list of elements, and ignore the order,  instead of writing a `compare` function. BUT because the commit id is non-deterministic, what we have here works for me.

tests-e2e/core/snapshotenvironmentbinding_test.go
```diff
@@ -770,8 +820,20 @@ func buildSnapshotEnvironmentBindingStatus(components []appstudiosharedv1.Bindin
 				URL: url, Branch: branch, Path: path[i], GeneratedResources: []string{}, CommitID: commitID,
 			},
 		})
+		if len(deploymentNames) > 0 {
```
jgwest, line 823: Feel free to poke holes in my logic here!

I think this addition to `buildAndUpdateBindingAndDeploymentStatus` is incorrect: the purpose of the SnapshotEnvironmentBinding E2E tests is to make sure that the `snapshotenvironmentbinding_controller.go` is updating the `.status` field, but instead here we are updating the status field ourselves, in this function.

The purpose of this function is to simulate the HAS controller (another stonesoup component), which is not running when the GitOps Service E2E tests are running, thus we need to fake it. Some of the fields of the `.status` of the `SnapshotEnvironmentBinding` are set by HAS, and since HAS isn't present here, we need to simulate what the HAS controller would do.
- The fields that the HAS controller would normally update are listed as `This status is updated by the Application Service controller.`

Instead, it looks like we've added code here to update the non-HAS fields status in the test, which means the test is testing itself, rather than testing the product code.

(Or, said another way, I believe the existing E2E tests should still pass without this change to this function.)

jgwest (changes requested):
Looks great @keithchong! I think I caught an issue with the update of `.status` in the SnapshotEnvironmentBinding, feel free to poke holes in my logic. After that, just some minor tweaks then good to go.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
```diff
@@ -758,6 +758,96 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler Tests", func() {
...
+			Expect(len(binding.Status.GitOpsDeployments)).To(Equal(2))
+			// We are not sure of the order
+			var indexOfComponentA, indexOfComponentB int
+			if binding.Status.GitOpsDeployments[0].ComponentName == "component-a" {
+				indexOfComponentA = 0
+				indexOfComponentB = 1
+			} else { // There are only two components, so the indices are switched
+				indexOfComponentA = 1
+				indexOfComponentB = 0
+			}
```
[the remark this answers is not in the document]
keithchong: Updated. See lines 794 to 807.

tests-e2e/fixture/binding/fixture.go
```diff
@@ -105,6 +105,81 @@ func HaveStatusGitOpsDeployments(gitOpsDeployments []appstudiosharedv1.BindingSt
 	}, BeTrue())
 }
 
+func HaveGitOpsDeploymentsWithStatusProperties(gitOpsDeployments []appstudiosharedv1.BindingStatusGitOpsDeployment) matcher.GomegaMatcher {
```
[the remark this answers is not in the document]
keithchong: Added.

tests-e2e/fixture/binding/fixture.go
keithchong: I've looked at the ContainElements function and will keep that in mind.

"ContainElements succeeds if actual contains the passed in elements. The ordering of the elements does not matter. By default ContainElements() uses Equal() to match the elements, however custom matchers can be passed in instead."

tests-e2e/fixture/binding/fixture.go
keithchong: I hope the commit hash lengths don't increase in the future?  I noticed they are 40 characters in length.

tests-e2e/core/snapshotenvironmentbinding_test.go
keithchong: I updated the tests.  I believe you are correct. I was thinking about how the expected values could be initialized.  I initially hard-coded the strings where they are now (after the update), but I saw places in the tests where the expected values are dependent upon the successful completion of previous steps. Take for example:

```
expectedGitOpsDeployments := 
    []appstudiosharedv1.BindingStatusGitOpsDeployment{{
	ComponentName: binding.Spec.Components[0].Name, GitOpsDeployment: gitOpsDeploymentName,
    }}
```

For the ComponentName, should we just hard-code `component-a` there?  In theory, if somewhere in the code made that name not be the original intended name, then that part of the test will still pass...I think?  So this led me to think we need to do the same for the status....it's just initializing the expected value.  The actual values of the binding object will be changed by the product code (the health will change from Progressing to Healthy, the commit ID will be overwritten, etc. I saw the test fail because the app was stuck on Progressing.)

jgwest (approved):
Great work, LGTM, thanks @keithchong! And thanks @jparsai for reviewing

PR #374, opened 2023-03-03
Add Argo CD Grafana dashboard for the GitOps Service instance

jopit:
#### Description:
- Add Argo CD Grafana dashboard for the GitOps Service instance


jgwest (approved):
Love to see all these great metrics we can now use to monitor the gitops-service-argocd Argo CD instance... thanks @jopit!

PR #376, opened 2023-03-06
GITOPSRVCE-449: update known bad values on stonesoup clusters, via init-container

jgwest:
#### Description:
- See [GITOPSRVCE-449](https://issues.redhat.com//browse/GITOPSRVCE-449) for details of the problem, and the fix.
- This fix detects if the database contains the problematic field values, and if so, updates them, before the GitOps Service backend container starts, via an init container.
- If the problematic field values are not detected, this is a no-op.


jparsai:
This PR will be reverted once DB is fixed, right?

/lgtm

chetan-rns (approved):
Thanks @jgwest 

/lgtm

PR #378, opened 2023-03-08
Fix kustomize error in appstudio-staging-cluster overlay

jgwest:
#### Description:
- The appstudio-staging-cluster overlay is currently failing to build when generated via kustomize
- Solution is to prevent dual definition of prometheus manifests


jgwest:
Merging w/o reviews as this is YAML-only change.

PR #381, opened 2023-03-15
chore: Update wait.Poll to wait.PollImmediate

jgwest:
#### Description:
- Update references from `wait.Poll` to `wait.PollImmediate`. In all cases, they should have been `wait.PollImmediate` to begin with.


jgwest:
Failed due to:
> E0315 06:16:40.839428    7559 portforward.go:233] lost connection to pod 

I believe this is from `kubectl port-foward`. I will open a bug for this.

/retest

jgwest:
Thanks @jopit!

PR #383, opened 2023-03-15
chore: remove deprecated getK8sClientForGitopsEngineInstance function

jgwest:
#### Description:
- The `getK8sClientForGitOpsEngineInstance` field has long since been deprecated by the newer `k8sClientFactory` field.
- The `k8sClientFactory` interface is a superset of the `getK8sClientForGitOpsEngineInstance`  functionality
- This deprecated field likely should have been removed as part of my original ManagedEnvironment PR, but better late than never :grinning:


chetan-rns (approved):
Thanks @jgwest 

/lgtm

PR #385, opened 2023-03-16
Add metrics for number of operations CR

Rizwana777:
#### Description:
- Added metrics for number of operation CR on a cluster
- Unit tests

https://issues.redhat.com/browse/GITOPSRVCE-430


jgwest (changes requested):
Hey @Rizwana777 , PR looks great, but I made a mistake in the original story description: the implementation for this should go into cluster-agent, rather than to backend. 
- Why? Architecturally, the Operation CRs are not always guaranteed to be on the same cluster as the backend component. But, they are guaranteed to be on the same cluster as cluster-agent.
- The Operation CRs _are_ all on the same cluster at the moment, but in the future they may not be, so we can take the time now to make sure we are scanning from the right splace.

My guess is it should be super quick to move the code you wrote into cluster-agent component, sorry for the inconvience! Otherwise, everything looks good to go.

Rizwana777:
> Hey @Rizwana777 , PR looks great, but I made a mistake in the original story description: the implementation for this should go into cluster-agent, rather than to backend.
> 
> * Why? Architecturally, the Operation CRs are not always guaranteed to be on the same cluster as the backend component. But, they are guaranteed to be on the same cluster as cluster-agent.
> * The Operation CRs _are_ all on the same cluster at the moment, but in the future they may not be, so we can take the time now to make sure we are scanning from the right splace.
> 
> My guess is it should be super quick to move the code you wrote into cluster-agent component, sorry for the inconvience! Otherwise, everything looks good to go.

No problem, I will move the code to cluster-agent and push the changes

Rizwana777:
@jgwest Moved the code from backend to cluster-agent , PTAL

jgwest (approved):
Looks great, thanks again @Rizwana777!

PR #386, opened 2023-03-16
Re-enable deletion of ManagedEnvironments, as long as no K8sToDBResourceMapping points to it

jgwest:
#### Description:
- I've updated the ManagedEnvironment cleanup logic: it should only allow deletion of a ManagedEnvironment if there is no corresponding K8sToDBResourceMapping that ALSO points to the ManagedEnvironment
- New unit test to verify the new behaviour
- Add a 'skipDelay' parameter to functions in db_reconciler_test.go, in order to reduce the time it takes to run these functions when run as part of unit tests. This significantly reduces the length of time it takes to run the backend tests.
- Rename appRowBatchSize => rowBatchSize

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+*.yaml
+*.txt
+jgw/
 .stfolder
 .idea/
 .vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-*.yaml
-*.txt
-jgw/
 .stfolder
 .idea/
 .vscode/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,6 @@
+*.yaml
+*.txt
+jgw/
```
jparsai, line 3: I think this is committed by mistake.

jparsai (commented):
LGTM, just have one comment.

.gitignore
jgwest: Whoops, thanks @jparsai!

jparsai:
Thanks @jgwest 
/lgtm

jgwest:
Thanks @jparsai!

PR #387, opened 2023-03-16
Don't include DB credentials in the connection err

chetan-rns:
#### Description:
- Don't include DB credentials in the errors because they will be logged.


jgwest (approved):
Probably my fault :sweat_smile: , great catch! LGTM, thanks @chetan-rns!

PR #394, opened 2023-03-22
Fix zap log level Deployment YAML

jgwest:
#### Description:
- The zap log level parameter is incorrectly specified in the Deployment YAML


jgwest:
Merging w/o reviews/tests, as this is breaking AppStudio cluster.

PR #395, opened 2023-03-23
Remove unnecessary logging of SnapshotEnvBinding .status changes

jgwest:
#### Description:
- In appstudio-controller on live cluster,  we're seeing SnapshotEnvironmentBinding controller logs about ~1 per second, related to updates to SnapshotEnvironmentBinding resource.
- However, in most cases, these logs are only updates to the `.status` field of the SnapshotEnvironmentBinding, for which there is little value in logging (e.g. for auditing purposes).
- We can thus remove that log statement for this case.


jopit (approved):
LGTM

jgwest:
Thanks @jopit!

PR #397, opened 2023-03-27
Refactor backend GitOpsDeployManagedEnvironment status update code

jgwest:
#### Description:
- This PR has no logic changes, just refactoring of existing code as follows:
- Rather than calling `updateManagedEnvironmentConnectionStatus` at a bunch of places in the code, instead we should just return a struct, `connectionInitializedCondition`, from functions, and set the status in a single central location `internalProcessMessage_ReconcileSharedManagedEnv`.
- Utility functions for creating various conditions, to reduce code duplication
- Moving the conditions into backend-shared, so they can be consumed by other components.

### DIFF
--- a/backend-shared/apis/managed-gitops/v1alpha1/gitopsdeploymentmanagedenvironment_types.go
+++ b/backend-shared/apis/managed-gitops/v1alpha1/gitopsdeploymentmanagedenvironment_types.go
@@ -58,6 +58,23 @@ type GitOpsDeploymentManagedEnvironment struct {
 	Status GitOpsDeploymentManagedEnvironmentStatus `json:"status,omitempty"`
 }
 
+type ManagedEnvironmentConditionReason string
+
+const (
+	ConditionReasonSucceeded                        ManagedEnvironmentConditionReason = "Succeeded"
+	ConditionReasonKubeError                        ManagedEnvironmentConditionReason = "KubernetesError"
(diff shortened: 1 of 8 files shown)


jgwest:
```
Summarizing 4 Failures:
  [FAIL] Managed Environment Status E2E tests Create a ManagedEnvironment [It] should have a connection status condition of False when the host for the targeted cluster does not exist
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/managed_environment_status_test.go:41
  [FAIL] Managed Environment Status E2E tests Create a ManagedEnvironment [It] should have a connection status condition of False when the credentials are missing a required field
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/managed_environment_status_test.go:71
  [FAIL] Managed Environment Status E2E tests Create a ManagedEnvironment [It] should have a connection status condition of False when the credentials contain bad data
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/managed_environment_status_test.go:101
  [FAIL] Managed Environment Status E2E tests Create a ManagedEnvironment [It] should have a connection status condition of False when there is no current contex in the kubeconfig
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/managed_environment_status_test.go:131 
```

chetan-rns (approved):
Thanks @jgwest 

/lgtm

jgwest:
Thanks @chetan-rns!

PR #412, opened 2023-04-05
Fix failing of unit tests in db_metrics_test.go file

Rizwana777:
#### Description:


jgwest (approved):
LGTM, thanks @Rizwana777!

PR #414, opened 2023-04-12
Run E2E by installing service in cluster.

jparsai:
## Description
This PR is a prerequisite for https://github.com/openshift/release/pull/34335. There are few changes needed in GitOps Service in order to run E2E by installing service in cluster instead of CLI as it does now.

## Type of changes

- Use a separate `Dockerfile` for creating image from OpenShift CI, it will allow us to use different configs for CI and development.
- Added `DISABLE_APPSTUDIO_WEBHOOK` environment variables to enable webhooks in CI.
- Added a script which will be executed by CI to trigger tests.
https://github.com/jparsai/release/blob/GitOps-e2e/ci-operator/config/redhat-appstudio/managed-gitops/redhat-appstudio-managed-gitops-openShiftCI.yaml#L44
 
## Ticket 
https://issues.redhat.com/browse/GITOPSRVCE-220


scripts/openshiftci-presubmit-e2e-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Do not show token in CI log
+set +x
+
+# show commands
+set -x
+
+# Set necessary env variables
+export PATH="$PATH:$(pwd)"
+KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
```
chetan-rns, line 11: The `KUBECONFIG` variable is not used in the script. Do we need to keep it?

scripts/openshiftci-presubmit-e2e-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+# Install Argo CD
+make install-argocd-openshift
+
+# Deploy GitOps Service components
+# Use Image built by CI, image name is provided as env variable 'CI_IMAGE' by CI
+echo "Using Image $CI_IMAGE"
+make install-all-k8s IMG=$CI_IMAGE
+
+# Port forwarding is needed by E2E to connect with DB
+kubectl port-forward --namespace gitops svc/gitops-postgresql-staging 5432:5432 &
```
chetan-rns, line 22: `make install-all-k8s` will also port-forward after installing everything. Do we still need to explicitly port-forward?

scripts/openshiftci-presubmit-e2e-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+KUBE_PID=$!
+
+# DB Password is required by E2E to connect with DB
+DB_PASS=$(kubectl get -n gitops secret gitops-postgresql-staging -o jsonpath="{.data.postgresql-password}" | base64 --decode)
+make test-e2e DB_PASS=$DB_PASS
```
chetan-rns, line 27: Nit: We can also export the DB_PASS `variable` without passing it as an argument.

Dockerfile.ci
chetan-rns, line 1: If this Dockerfile is specific to ci should we move it inside the `.ci` folder?

appstudio-controller/config/manager/manager.yaml
```diff
@@ -40,6 +40,9 @@ spec:
           - containerPort: 9443
             name: webhook-server
             protocol: TCP
+        env:
+        - name: DISABLE_APPSTUDIO_WEBHOOK
```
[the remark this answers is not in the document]
jparsai: No, actually it is disabled, PTAL at here
https://github.com/redhat-appstudio/managed-gitops/blob/main/backend/Makefile#L64

scripts/openshiftci-presubmit-e2e-tests.sh
jparsai: Yes, we dont need it. I put it there when I was not able to make DB connection from E2E test, but later port forwarding fixed the problem and it is not needed anymore.

scripts/openshiftci-presubmit-e2e-tests.sh
jparsai: We can but in that case it will print the password in logs, to avoid that we have to remove `set -x` but then we wont be able to see commands executed in logs which makes it easy to separate the logs for debugging.

image

Dockerfile.ci
jparsai: I tried it, but then we have to make changes to it, since it has to copy folders from root directory and if we move it to a child folder then its not easy to copy from parent folder. it would require few changes in build command to pass parent folder as an argument, AFAIK and docker build command is executed by OpneShift CI and I am not sure if I can pass any arguments to it. Keeping docker file in rot folder was the simple and quick solution. 

https://stackoverflow.com/questions/24537340/docker-adding-a-file-from-a-parent-directory

Makefile
```diff
@@ -5,6 +5,7 @@ BASE_IMAGE ?= gitops-service
 USERNAME ?= redhat-appstudio
 IMG ?= quay.io/${USERNAME}/${BASE_IMAGE}:${TAG}
 APPLICATION_API_COMMIT ?= 394e8c127f031990838397fcb2d1e10110b2e7ae
+DB_PASS ?= gitops
```
jgwest, line 8: Hey Jayendra, is this needed? If possible, I'd prefer not to hardcode DB_PASS to any particular value. (We are using a hardcoded DB_PASS for the local case, but in that case the postgresql server shouldn't be public facing)

appstudio-controller/config/manager/manager.yaml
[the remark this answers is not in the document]
jgwest: I just double checked, the logic is to enable webhooks unless DISABLE_APPSTUDIO_WEBHOOK=true. So, I believe this means that for the `manager.yaml` case, this env var isn't needed. (But the env var IS still needed in the backend/Makefile case)

scripts/openshiftci-presubmit-e2e-tests.sh
jgwest: You can disable, then re-enable, logging, like:
```
set +x
# ( run a command you don't want to output )
set -x
```

backend/config/manager/manager.yaml
```diff
@@ -45,6 +45,8 @@ spec:
             secretKeyRef:
               key: postgresql-password
               name: gitops-postgresql-staging
+        - name: DISABLE_APPSTUDIO_WEBHOOK
+          value: "false"
```
jgwest, line 49: Likewise, think this isn't needed for this case.

jgwest (changes requested):

appstudio-controller/config/manager/manager.yaml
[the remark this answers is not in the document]
jparsai: This is removed.

scripts/openshiftci-presubmit-e2e-tests.sh
jparsai: This is done.

Makefile
jparsai: I removed it, since we are not passing DB_PASS as an argument anymore.

jgwest (approved):
LGTM, thanks @jparsai! And thanks @chetan-rns for reviewing!

PR #415, opened 2023-04-12
Reduce memory usage by reusing the client from the workspace eventloop

chetan-rns:
#### Description:
- Instead of creating a new client whenever the `startNewStatusUpdateTimer()` is invoked, we reuse an existing client passed from the workspace event loop.  Thereby we can avoid multiple client objects escaping to the heap.
- Use a predicate in the secrets controllers to filter secrets of type: managed-environment.
- Test to verify memory usage.

https://issues.redhat.com/browse/GITOPSRVCE-508

### DIFF
--- a/backend/controllers/managed-gitops/secret_controller.go
+++ b/backend/controllers/managed-gitops/secret_controller.go
@@ -27,7 +27,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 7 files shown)


chetan-rns:
### How to verify:

Adding the steps that I followed to compare the memory consumption before and after introducing load.

1. Deploy GitOps Service from the main branch and note down the memory consumed by the backend controller for comparison.
2. Increase the resource requests and limits before introducing the load to prevent OOM kills.

```yaml
containers:
- resources:
  	limits:
     		cpu: 2000m
           memory: 4800Mi
 	requests:
     		cpu: 1000m
           memory: 2400Mi
```

3. Run the test that will create resources to simulate the production environment 

```shell
go test -v -p=1 -timeout=20m -race -count=1 -coverprofile=coverage.out ./utilities/load-test/gitops-service/...
```
4. Observe the increase in memory with the load 
5. Build and push the image with the updates in this PR. Update the image tag of the backend controller and wait for a new rollout of the pod
6. The memory consumption will go down once the new pod starts running.

jgwest (approved):
Looks and works great, thanks @chetan-rns! I've pushed a go.mod update to load-test, waiting for PR tests to complete, then I will merge.

PR #416, opened 2023-04-12
Add application/component/environment labels to GitOpsDeployment resources

jopit:
#### Description:
- Added labels to GitOpsDeployment resources to identify the associated application, component and environment

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
@@ -412,6 +412,11 @@ func generateExpectedGitOpsDeployment(component appstudioshared.BindingComponent
 		}
 	}
 
+	// Append labels to identify the Application, Component and Environment associated with this GitOpsDeployment
+	res.ObjectMeta.Labels[appstudioLabelKey+"/application"] = binding.Spec.Application
+	res.ObjectMeta.Labels[appstudioLabelKey+"/component"] = component.Name
+	res.ObjectMeta.Labels[appstudioLabelKey+"/environment"] = binding.Spec.Environment
+
 	// Ensures that this method only adds 'appstudio.openshift.io' labels
 	// - Note: If you remove this line, you need to search for other uses of 'removeNonAppStudioLabelsFromMap' in the
 	// code, as you may break the logic here.
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
@@ -435,7 +435,7 @@ func limitLength(str string, logger logr.Logger) string {
 	if len(str) <= 63 {
 		return str
 	}
-	logger.Error(nil, "label value should be at most 63 characters", "value", str)
+	logger.Error(nil, "SEVERE: label value in SnapshotEnvironmentBinding should be at most 63 characters", "value", str)
 	return str[:63]
 }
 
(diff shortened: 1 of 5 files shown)


appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
```diff
@@ -412,6 +413,14 @@ func generateExpectedGitOpsDeployment(component appstudioshared.BindingComponent
 		}
 	}
 
+	// Append labels to identify the Application, Component and Environment associated with this GitOpsDeployment.
+	// Label values are limited to a maximum of 63 characters, while resource names by default can be up to 253.
+	// StoneSoup artificially limits these names to 63 characters, but to be safe we truncate the values if they
+	// are too long
+	res.ObjectMeta.Labels[appstudioLabelKey+"/application"] = limitLength(binding.Spec.Application, logger)
```
keithchong, line 420: I suppose this is the right label key, and not the one that ends with .com as mentioned in the Jira?

tests-e2e/core/snapshotenvironmentbinding_test.go
```diff
@@ -575,6 +575,117 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler E2E tests", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      gitOpsDeploymentName,
+					Namespace: binding.Namespace,
+				},
+			}
+
+			err = k8s.Get(&gitopsDeployment, k8sClient)
+			Expect(err).To(BeNil())
+			Eventually(gitopsDeployment, "2m", "10s").Should(gitopsDeplFixture.HaveLabel("appstudio.openshift.io/application", binding.Spec.Application))
+			Eventually(gitopsDeployment, "2m", "10s").Should(gitopsDeplFixture.HaveLabel("appstudio.openshift.io/component", binding.Spec.Components[0].Name))
```
keithchong, line 619: Looks like the controller will add the labels per component (it will loop over the list of components), so probably the test for one component is sufficient, like in this case.  Do you think we need to consider a multi-component app (eg. for componentB in this example) or do you think this is redundant?

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
jopit: Yes, after a slack discussion with Jonathan, we decided it made more sense to use `appstudio.openshift.io` as the label.  I'll put a comment in the jira.

keithchong: Ok.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
```diff
@@ -422,6 +431,14 @@ func generateExpectedGitOpsDeployment(component appstudioshared.BindingComponent
 	return res, nil
 }
 
+func limitLength(str string, logger logr.Logger) string {
+	if len(str) <= 63 {
+		return str
+	}
+	logger.Error(nil, "label value should be at most 63 characters", "value", str)
```
jgwest: SEVERE?

jgwest:
Rebased to fix go.mod conflict

jgwest (approved):
Great stuff, nice work threading the needle between the existing label addition code, and this new story :smile:. Thanks @jopit! Waiting for tests to pass after rebasing, then good to go.

jgwest:
Merging even though there is an RHTAP error: it appears to be an intermittent issue which has been reported by others, and which is not specific/related to any changes made on this PR.

PR #418, opened 2023-04-13
Configured Environment Variable to enable customisation of database

raghavi101:
#### Description:

- Added env var to facilitate customisation of database name which was previously hardcoded to "postgres"

https://issues.redhat.com/browse/GITOPSRVCE-338

### DIFF
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -112,7 +112,6 @@ if [ "$1" = "kube" ]; then
     echo " * Port forwarding is still active for some reason. Investigate further ..."
   fi
 export POSTGRESQL_DATABASE="postgres"
-
   # Exit now, do not continue with the rest of the bash script
   echo
   echo " ------------------------------------------------"
--- a/create-dev-env.sh
+++ b/create-dev-env.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
 # Checks if a binary is present on the local system
 exit_if_binary_not_installed() {
   for binary in "$@"; do
@@ -112,6 +111,7 @@ if [ "$1" = "kube" ]; then
   if lsof -i:5432 | grep LISTEN; then
     echo " * Port forwarding is still active for some reason. Investigate further ..."
   fi
+export POSTGRESQL_DATABASE="postgres"
 
   # Exit now, do not continue with the rest of the bash script
   echo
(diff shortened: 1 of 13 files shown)


create-dev-env.sh
```diff
@@ -269,7 +267,7 @@ else
     -p 5432:5432 \
     --network gitops-net \
     -d \
-    postgres:13 \
+    $POSTGRESQL_DATABASE:13 \
```
jgwest, line 270: `postgres:13 \` is the container image name/tag, not the database name, so it's fine to leave it as `postgres:13`

create-dev-env.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-
 # Checks if a binary is present on the local system
+export POSTGRESQL_DATABASE="postgres"
```
jgwest, line 3: If the `POSTGRESQL_DATABASE` environment variable is already set by the user calling the script, we should use that value, otherwise, if it's blank, then we should default to `postgres`. See [here](https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash) for how that looks in bash script.

psql.sh
```diff
@@ -3,12 +3,13 @@
 # Tip: More info on psql - https://tomcam.github.io/postgres/
 
 POSTGRES_CONTAINER="managed-gitops-postgres"
+export POSTGRESQL_DATABASE="postgres"
```
jgwest, line 6: As above, if the environment variable is already set by the user calling script, we should use the existing value, otherwise we should default to `postgres`

utilities/check-db-schema.sh
```diff
@@ -2,7 +2,7 @@
 
 # Stop exiting and return non-zero error code if an error occurs
 set -ue
-
+export POSTGRESQL_DATABASE="postgres"
```
jgwest, line 5: As above, if the environment variable is already set by the user calling script, we should use the existing value, otherwise we should default to `postgres`

utilities/verify-db-migration.sh
```diff
@@ -37,19 +37,20 @@ echo "MIGRATION_SQL_FILE: $MIGRATION_SQL_FILE"
 
 exit_if_binary_not_installed "pg_dump"
 
+export POSTGRESQL_DATABASE="postgres"
```
jgwest, line 40: As above, if the environment variable is already set by the user calling script, we should use the existing value, otherwise we should default to `postgres`

docs/debug.md
```diff
@@ -39,7 +39,7 @@ $ kubectl port-forward --namespace gitops-postgresql svc/gitops-postgresql-stagi
 Then run a `psql` command, to verify your access:
 
 ```shell
-psql postgresql://postgres:$PASSWORD@127.0.0.1:5432/postgres -c "select now()"
+psql postgresql://postgres:$PASSWORD@127.0.0.1:5432/$POSTGRESQL_DATABASE -c "select now()"
```
jgwest, line 42: Before this line, it couldn't hurt to say that POSTGRESQL_DATABASE defaults to `postgres`, so that users new to the project know what value to put in this field.

jgwest (changes requested):
LGTM, thanks @raghavi101, just some minor tweaks then good to go!

raghavi101:
Hi @jgwest thanks for your review comments! I've updated with the latest changes PTAL

jgwest:
Rebased to latest to remove confict.

jgwest (approved):
LGTM, thanks @raghavi101!

PR #420, opened 2023-04-13
Update backward compatibility test to use 'go test' instead of ginkgo 

trdoyle81:
#### Description:
Update the 'check-backward-compatibility' target to use 'go test' instead of 'ginkgo focus' as Ginkgo is not installed on OpenshiftCI

https://issues.redhat.com/browse/GITOPSRVCE-453


jgwest (approved):
LGTM, thanks @trdoyle81!

PR #430, opened 2023-04-21
Add namespaces and clusterResources fields to the app studio Environment API

jopit:
#### Description:
- Added namespaces and clusterResources fields to the app studio Environment API
- Updated the appstudio controller to copy these fields from the app studio Environment when a GitOpsDeploymentManagedEnvironment is created

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
@@ -57,7 +57,7 @@ var _ = Describe("Environment controller tests", func() {
 
 		})
 
-		createEnvironmentTest := func(allowInsecureSkipTLSVerifyParam bool) {
+		createEnvironmentTest := func(allowInsecureSkipTLSVerifyParam, clusterResources bool, namespaces []string) {
 			var err error
 
 			secret := corev1.Secret{
@@ -90,11 +90,8 @@ var _ = Describe("Environment controller tests", func() {
 							APIURL:                     "https://my-api-url",
 							ClusterCredentialsSecret:   secret.Name,
 							AllowInsecureSkipTLSVerify: allowInsecureSkipTLSVerifyParam,
-							ClusterResources:           true,
-							Namespaces: []string{
-								"namespace-1",
-								"namespace-2",
-							},
+							ClusterResources:           clusterResources,
+							Namespaces:                 namespaces,
 						},
 					},
 				},
@@ -136,14 +133,21 @@ var _ = Describe("Environment controller tests", func() {
 		}
 
 		It("should create a GitOpsDeploymentManagedEnvironment, if the Environment is created where AllowInsecureSkipTLSVerify field is true", func() {
-			createEnvironmentTest(true)
+			createEnvironmentTest(true, false, nil)
 		})
 
 		It("should create a GitOpsDeploymentManagedEnvironment, if the Environment is created where AllowInsecureSkipTLSVerify field is false", func() {
-			createEnvironmentTest(false)
+			createEnvironmentTest(false, false, nil)
 		})
 
-		updateEnvTest := func(allowInsecureSkipTLSVerifyParam bool) {
+		It("should create a GitOpsDeploymentManagedEnvironment, if the Environment is created where ClusterResources is true and namespaces are specified", func() {
+			createEnvironmentTest(false, true, []string{
+				"namespace-1",
+				"namespace-2",
+			})
+		})
+
+		updateEnvTest := func(allowInsecureSkipTLSVerifyParam, initialClusterResources, updatedClusterResources bool, initialNamespaces, updatedNamespaces []string) {
 			var err error
 
 			By("creating first managed environment Secret")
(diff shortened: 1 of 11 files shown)


appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
```diff
@@ -90,6 +90,11 @@ var _ = Describe("Environment controller tests", func() {
 							APIURL:                     "https://my-api-url",
 							ClusterCredentialsSecret:   secret.Name,
 							AllowInsecureSkipTLSVerify: allowInsecureSkipTLSVerifyParam,
+							ClusterResources:           true,
+							Namespaces: []string{
+								"namespace-1",
+								"namespace-2",
+							},
```
jgwest, line 97: It could be good to also have a test similar to what we had before, where we the environment that we created did not have these values set:
```go
							ClusterResources:           true,
							Namespaces: []string{
								"namespace-1",
								"namespace-2",
							},
```
(e.g. clusterresources is false, and namespace is empty, indicating that Argo CD has full access to the cluster.)

With the change it looks like (correct me if I'm wrong) we are now only testing the clusterResources = true and len(namespaces) > 0 case.

It should be possible to update the `createEnvironmentTest` function to support both (for example, a parameter that controls which case is being tested, and different expect values based on that parameter)

appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
```diff
@@ -183,6 +192,11 @@ var _ = Describe("Environment controller tests", func() {
 							APIURL:                     "https://my-api-url",
 							ClusterCredentialsSecret:   secret2.Name,
 							AllowInsecureSkipTLSVerify: allowInsecureSkipTLSVerifyParam,
+							ClusterResources:           true,
+							Namespaces: []string{
+								"namespace-1",
+								"namespace-2",
+							},
```
jgwest, line 199: Same as above, it would be good to have a update test that continues to verify the original case: cluster resources = false, and empty namespaces.

tests-e2e/fixture/managedenvironment/fixture.go
```diff
@@ -91,3 +91,57 @@ func HaveCredentials(expectedEnvSpec managedgitopsv1alpha1.GitOpsDeploymentManag
 		return res
 	}, BeTrue())
 }
+
+// HaveClusterResources checks if the ClusterResources field of Environment is equal to clusterResouces.
+func HaveClusterResources(clusterResouces bool) matcher.GomegaMatcher {
```
jgwest, line 96: clusterResouces => clusterResources

tests-e2e/fixture/managedenvironment/fixture.go
```diff
@@ -91,3 +91,57 @@ func HaveCredentials(expectedEnvSpec managedgitopsv1alpha1.GitOpsDeploymentManag
 		return res
 	}, BeTrue())
 }
+
+// HaveClusterResources checks if the ClusterResources field of Environment is equal to clusterResouces.
```
jgwest, line 95: Environment => GitOpsDeploymentManagedEnvironment, I believe

tests-e2e/fixture/managedenvironment/fixture.go
```diff
@@ -91,3 +91,57 @@ func HaveCredentials(expectedEnvSpec managedgitopsv1alpha1.GitOpsDeploymentManag
...
+		res := clusterResouces == menv.Spec.ClusterResources
+
+		fmt.Println("HaveClusterResources:", res, "/ Expected:", clusterResouces, "/ Actual:", menv.Spec.ClusterResources)
+
+		return res
+
+	}, BeTrue())
+}
+
+// HaveNamespaces checks if the HaveNamespaces field of Environment is equal to namespaces.
```
jgwest, line 122: Environment => GitOpsDeploymentManagedEnvironment, I believe

tests-e2e/core/managed_environment_test.go
```diff
@@ -809,6 +809,86 @@ var _ = Describe("Environment E2E tests", func() {
...
+
+			err = k8s.Update(&environment, k8sClient)
+			Expect(err).To(BeNil())
+
+			Eventually(managedEnvCR, "2m", "1s").Should(
+				SatisfyAll(
+					managedenvironment.HaveClusterResources(environment.Spec.UnstableConfigurationFields.ClusterResources),
+					managedenvironment.HaveNamespaces(environment.Spec.UnstableConfigurationFields.Namespaces),
+				),
+			)
```
jgwest, line 888: Can we add update one more quick check that removes the namespaces (and clusterResources is false), and then verifies that they are removed from the ManagedEnv?

jgwest (changes requested):
Looks great, thanks @jopit! I have some (entirely opinionated) minor test-related requests, then good to go.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
jopit: I did add a test for this case at around line 681, but I'll update `createEnvironmentTest` anyway.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller_test.go
jopit: Agreed, I'll make the suggested change.

tests-e2e/core/managed_environment_test.go
jopit: Will do.

tests-e2e/fixture/managedenvironment/fixture.go
jopit: I'll make the change

tests-e2e/fixture/managedenvironment/fixture.go
jopit: I'll update this

jopit:
Hi @jgwest I've made all the changes based on your review.

jgwest (approved):
Perfect, thanks @jopit! And thanks @jparsai for reviewing!

PR #432, opened 2023-04-25
log refinement to ADR6.

jparsai:
#### Description:
This PR is to update logs to ADR6 format.

#### JIRA Ticket 
https://issues.redhat.com/browse/GITOPSRVCE-549

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
@@ -62,7 +62,9 @@ type EnvironmentReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *EnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	ctx = sharedutil.AddKCPClusterToContext(ctx, req.ClusterName)
-	log := log.FromContext(ctx).WithValues("request", req)
+	log := log.FromContext(ctx).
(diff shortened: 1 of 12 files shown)


appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -125,11 +127,14 @@ func (r *EnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-
-			log.Info("Creating GitOpsDeploymentManagedEnvironment", "managedEnv", desiredManagedEnv.Name)
 			if err := rClient.Create(ctx, desiredManagedEnv); err != nil {
 				return ctrl.Result{}, fmt.Errorf("unable to create new GitOpsDeploymentManagedEnvironment: %v", err)
 			}
+
+			log.Info("Created GitOpsDeploymentManagedEnvironment", "managedEnv", desiredManagedEnv.Name,
+				sharedutil.LogAction, sharedutil.LogAction_Add,
+				sharedutil.LogAudit, true)
+
```
jgwest, line 137: IMHO, if we are already calling `sharedutil.LogAPIResourceChangeEvent`, we probably can skip adding `LogAction` and `LogAudit` to this log statement above the call to `LogAPIResourceChangeEvent`. `LogAPIResourceChangeEvent` will already output this data. (Same with similar cases below)

jgwest (changes requested):
LGTM, just one (opinionated) suggestion re: whether we need to add the `audit` and `action` logs for cases where we are already reporting those via `LogAPIResourceChangeEvent`

PR #433, opened 2023-04-25
Set Argo CD reconciliation timeout env variable

chetan-rns:
#### Description:
- Set the refresh rate by configuring the `ARGOCD_RECONCILIATION_TIMEOUT` environment variable in the application controller and repo server

#### How to test?

1. Create a sample GitOpsDeployment and check if the corresponding Argo CD app is synced.
2. Check the log timestamp of the application controller pod in the `gitops-service-argocd` namespace and verify if Argo CD refreshes the app every 60s.

https://issues.redhat.com/browse/GITOPSRVCE-425

### DIFF
--- a/cluster-agent/utils/argocd_create.go
+++ b/cluster-agent/utils/argocd_create.go
@@ -37,6 +37,9 @@ const (
 	DefaultAppProject = "default"
 
 	ArgoCDFinalizerName = "argoproj.io/finalizer"
+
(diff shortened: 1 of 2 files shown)


jgwest (approved):
Looks great, thanks @chetan-rns! I've verified that the Argo CD pods in `gitops-service-argocd` have the new env var.

PR #434, opened 2023-04-26
Fix gradual memory leak in the backend controller

chetan-rns:
#### Description:
* Use a common context for all events in the application event runner. Creating a new context for every event will cause it to stay on the heap until it is canceled.
* Increase maximum backoff retry duration so the controller doesn't reconcile too much on CRs with errors.

##### Before:

The memory increased by almost 200MiB in the course of  2 hours.

Screenshot from 2023-04-24 13-58-36

##### After:

We can observe that memory has been almost the same in the 2 hours window with the same load.

Screenshot from 2023-04-25 12-26-37

https://issues.redhat.com/browse/GITOPSRVCE-557


jgwest (approved):
Looks great, thanks @chetan-rns! Our service continues to get leaner and more efficient :grin:

PR #435, opened 2023-04-26
Create route-less copy of test data component-based-gitops-repository folder

jopit:
#### Description:
- In order to fix an intermittent promotionrun e2e failure, the applications created by the tests need to exist in it's own namespace.  However, the current e2e test application creates a route, and so the two instances of the application try to create a route with the same hostname, which isn't allowed.  The simplest fix is to create a copy of the test application without the route. This PR is to create that copy.


jgwest (approved):
Perfect, thanks @jopit!

PR #436, opened 2023-04-26
Run PromotionRun E2E test apps in separate namespaces

jopit:
#### Description:
- Run PromotionRun E2E test apps in separate namespaces so that the two apps don't share resources, which leads to syncing issues with Argo CD.
- This requires a new version of the e2e apps, since otherwise the two instances of the application will try to create a route with the same name, which is not allowed by OpenShift. The route-less version of the apps was committed via https://github.com/redhat-appstudio/managed-gitops/pull/435


jgwest (approved):
LGTM, thanks @jopit!

PR #445, opened 2023-05-02
Fix comparison race condition in SnapshotEnvironmentBinding E2E tests

jgwest:
#### Description:
- This PR fixes this race condition in SEB tests:
```
Summarizing 1 Failure:
  [FAIL] SnapshotEnvironmentBinding Reconciler E2E tests Testing SnapshotEnvironmentBinding Reconciler. [It] Should recreate GitOpsDeployment, if Binding still exists but GitOpsDeployment is deleted.
  /home/jgw/managed-gitops/managed-gitops/tests-e2e/core/snapshotenvironmentbinding_test.go:370
```

The race condition was due to the fixture function comparing the sync status/health/commit id, which the test was not designed to expect.


jopit (approved):
lgtm

jgwest:
Thanks @jopit!

PR #447, opened 2023-05-04
fix: remove finalizer for a deleted GitOpsDepl when the controllers are restarted

chetan-rns:
#### Description:
- The backend controller should not forget to remove the finalizer from a deleted GitOpsDepl when the controllers are restarted. Currently, we skip handling the deletion if the DB resources are already removed even though the GitOpsDeployment with a finalizer could still exist.

https://issues.redhat.com/browse/GITOPSRVCE-455


jgwest (approved):
LGTM, thanks @chetan-rns!

PR #449, opened 2023-05-05
Update appstudio-controller container memory limit

jgwest:
#### Description:
- Does what it says on the tin :grin:


jgwest:
Change to infrastructure YAML only, merging without reviews.

PR #450, opened 2023-05-05
Delete GitOpsDeployment when Component removed from SEB

jopit:
#### Description:
- Updated the SnapshotEnvironmentBinding reconciler to delete any GitOpsDeployments that are not associated with a Component.
- Added unit and e2e tests for the deletion of deployments in response to removal of components
- Added constant strings for the keys used for the labels applied to the deployment to identify the associated application, component and environment

The most efficient way to find the deployments to delete uses the `component` label on the deployment as the key to a map of expected deployments.  Label values are limited in length to 63 characters (in the production, the names of components, applications and environments will be limited to 63 characters by a webhook.) The existing code that sets the labels on the deployment truncates the value if the limit is exceeded.  There are some unit and e2e tests that set component/application names to longer than 63 characters in order to test the algorithm that produces the name for a GitOpsDeployment.  For the the e2e tests, if the component name is truncated, the new deletion code causes the associated deployment to continually be deleted then recreated.

The right solution is for the reconciler to error out if any of the application, component or environment names are greater than 63 characters. This requires the following additional changes:

- Changed the reconciler so that instead of truncating label values that are too long, it will error out instead.
- Changed the test cases that test the logic for generating shortened names for deployments to keep the names of the application, component and environment to at most 63 characters.

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
@@ -164,7 +164,7 @@ func (r *SnapshotEnvironmentBindingReconciler) Reconcile(ctx context.Context, re
 		}
 
 		// Delete all existing deployments associated with this binding
-		err := deleteUnmatchedDeployments(ctx, rClient, binding, nil, log)
+		err := deleteUnmatchedDeployments(ctx, *binding, nil, rClient, log)
 		if err != nil {
 			return ctrl.Result{}, err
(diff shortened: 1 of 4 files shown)


jgwest (approved):
Looks great, thanks @jopit ! And thanks @jparsai for reviewing!

FYI I rebased and pushed some minor (opinionated) tweaks: normally I would add these as review comments so we coudl discuss, but in the interest of time (and because @jopit is on PTO today), I've included them in the commit as is. @jopit  - feel free to take a look and let me know if you spot any issues.

PR #453, opened 2023-05-09
Improve error message when ServiceAccount doesn't have needed roles

jopit:
#### Description:
- Added new message for when the check for cluster access returns status of `forbidden`.
- Added new unit tests for this situation.


Rizwana777:
Looks good to me

jgwest (approved):
Looks perfect, nice mocking to ensure we're handling it correctly :star_struck:, thanks @jopit! And thanks @Rizwana777 for reviewing!

PR #456, opened 2023-05-10
Add ability to control the frequency of self-healing

jopit:
#### Description:
- Added ability to use an environment variable to set the the interval in minutes for both database and namespace reconciliation.  If the variable is set to `0`, reconciliation will be disabled.  If the variable is not set, the current value of `30` minutes will be used
- Updated the Makefiles for the backend and the cluster-agent to set the variable to `0` to disable reconciliation when running the e2e and chaos tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,10 +62,10 @@ deploy-k8s-env: pre-deploy-crds postgresql-secret-on-k8s ## Deploy all controlle
 undeploy-k8s-env: pre-undeploy-crds ## Remove all controller/DB workloads from K8s.
 	$(KUSTOMIZE) build $(MAKEFILE_ROOT)/manifests/overlays/k8s-env | kubectl delete -f - 
 
-deploy-k8s-env-e2e: kustomize postgresql-secret-on-k8s ## Deploy all controller/DB workloads to K8s for e2e tests, use e.g. IMG=quay.io/pgeorgia/gitops-service:latest to specify a specific image
+deploy-k8s-env-e2e: pre-deploy-crds postgresql-secret-on-k8s ## Deploy all controller/DB workloads to K8s for e2e tests, use e.g. IMG=quay.io/pgeorgia/gitops-service:latest to specify a specific image
 	$(KUSTOMIZE) build $(MAKEFILE_ROOT)/manifests/overlays/k8s-env-e2e | COMMON_IMAGE=${IMG} envsubst | kubectl apply -f -
 
-undeploy-k8s-env-e2e: kustomize ## Remove all controller/DB e2e test workloads from K8s.
+undeploy-k8s-env-e2e: pre-undeploy-crds ## Remove all controller/DB e2e test workloads from K8s.
 	$(KUSTOMIZE) build $(MAKEFILE_ROOT)/manifests/overlays/k8s-env-e2e | kubectl delete -f -
 
 
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,11 @@ deploy-k8s-env: pre-deploy-crds postgresql-secret-on-k8s ## Deploy all controlle
 undeploy-k8s-env: pre-undeploy-crds ## Remove all controller/DB workloads from K8s.
 	$(KUSTOMIZE) build $(MAKEFILE_ROOT)/manifests/overlays/k8s-env | kubectl delete -f - 
 
+deploy-k8s-env-e2e: kustomize postgresql-secret-on-k8s ## Deploy all controller/DB workloads to K8s for e2e tests, use e.g. IMG=quay.io/pgeorgia/gitops-service:latest to specify a specific image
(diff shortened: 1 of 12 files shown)


jgwest (changes requested):
Code in the PR looks great!

In reviewing the PR, I realize I hadn't considered this in the original story:

Since  we now run the E2E tests on the cluster itself, using 'make install-all-k8s' (https://github.com/redhat-appstudio/managed-gitops/blob/293cd249dc48b9eea5e4e253e50c4af4bc43fecc/scripts/openshiftci-presubmit-e2e-tests.sh#LL30C1-L30C1), this means we will need a new kustomize overlay for the case where we want on cluster but with the self-healing disabled (an on-cluster 'e2e test' mode)

In general (feel free/please check my logic!), I think self-heling should be enabled/disabled in the following cases.

*Self-healing should be enabled when:*
A) developers are running on their local machine, via 'make start'
B) when developers are running an image on K8s itself, via 'make deploy-k8s-env' or 'install-all-k8s'
C) on staging/prod, of course

*Self-healing should be disabled when:*
D) Running e2e tests via 'make start-e2e'
E) When E2E tests are running on cluster, as part of a PR

In the PR as it currently exists, _AFAICT_ the following are not currently true: A, E

I think this can be fixed with:
- For A (or D), a new Procfile? One for the self-healing enabled case, and one for the non self-healing enabled case. Then update the makefile targets for these cases.
- For E: New kustomize overlay when running as E2E test, something like a new overlay under 'manifests/overlays/k8s-env-e2e', with a patch (similar to how we add env vars in manifests/overlays/stonesoup-member-clsuter), then update/create new makefile target to use that

WDYT?

jopit:
Made the changes as suggested by @jgwest. I added a new Procfile and makefile targets to disable self healing, so now `make start` will enable self healing and `make start-e2e` will disable it.  I also added a new kustomize overlay to disable self healing, and updated the script run for PRs to use the new overlay.

manifests/overlays/k8s-env-e2e/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+- ../../base/gitops-namespace
+- ../../../appstudio-controller/config/default
+- ../../../backend/config/default
+- ../../../cluster-agent/config/default
+- ../../base/postgresql-staging
+# - ../../base/gitops-service-argocd
+
+patchesStrategicMerge:
+- backend-deployment-patch.yaml
+- cluster-agent-deployment-patch.yaml
```
jgwest, line 15: In order to reduce some of the duplication (which is what kustomize is good at!) between this overlay and the `k8s-env` overlay, I wonder if it would be possible to further reduce, by doing something like:

```yaml
resources:
- ../k8s-env # pull the same resources as the parent....

patchesStrategicMerge: # but patch the env as before

patchesStrategicMerge:
- backend-deployment-patch.yaml
- cluster-agent-deployment-patch.yaml
```

I _think_ this will work? WDYT?

backend-shared/util/util.go
```diff
@@ -156,3 +160,17 @@ func GetRESTConfig() (*rest.Config, error) {
 	res.Burst = 250
 	return res, nil
 }
+
```
jgwest, line 163: Sorry I didn't think of this during first review, but I think we can write a quick little unit test for this, just so we have some coverage for this function. (It can likely be just testing the empty env var case, and then the non-empty env var case. I'm pretty sure `os.SetEnv` can be used in a test, so no need to mock this)

Makefile
```diff
@@ -62,6 +62,11 @@ deploy-k8s-env: pre-deploy-crds postgresql-secret-on-k8s ## Deploy all controlle
 undeploy-k8s-env: pre-undeploy-crds ## Remove all controller/DB workloads from K8s.
 	$(KUSTOMIZE) build $(MAKEFILE_ROOT)/manifests/overlays/k8s-env | kubectl delete -f - 
 
+deploy-k8s-env-e2e: kustomize postgresql-secret-on-k8s ## Deploy all controller/DB workloads to K8s for e2e tests, use e.g. IMG=quay.io/pgeorgia/gitops-service:latest to specify a specific image
```
jgwest, line 65: Just need to add `pre-deploy-crds` or `pre-undeploy-crds` to these 2 Makefile targets, as we are doing in the previous target in the Makefile (these were very recently added by another PR)

jgwest (changes requested):
Looks great, thanks @jopit! Some minor tweaks requested then good to go.

manifests/overlays/k8s-env-e2e/kustomization.yaml
jopit: Done

backend-shared/util/util.go
jopit: Done

Makefile
jopit: Done

jgwest (approved):
Looks great, thanks again @jopit!

PR #459, opened 2023-05-11
Add time series database metrics and grafana dashboard (#511)

keithchong:
#### Description:
See https://issues.redhat.com/browse/GITOPSRVCE-511 for details

To launch from your own development environment, just run:
make setup-e2e-openshift
manifests/scripts/metrics-dashboard/run.sh

### DIFF
--- a/manifests/scripts/metrics-dashboard/postgresql/grafana-postgresql-exporter-dashboard-cm.yaml
+++ b/manifests/scripts/metrics-dashboard/postgresql/grafana-postgresql-exporter-dashboard-cm.yaml
@@ -1,0 +1,1790 @@
+apiVersion: v1
+data:
+  postgresql-exporter-table-dashboard.json: |
+    {
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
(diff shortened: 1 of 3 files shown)


jopit (changes requested):
The legends shown in the screenshots attached to the Jira aren't very informative (at least to me) -- is it possible to make these more informative?

keithchong:
> The legends shown in the screenshots attached to the Jira aren't very informative (at least to me) -- is it possible to make these more informative?

RIght now the Legend is "Auto".  Those are the labels that are shown.  For example, this is Auto, and take note the first value, the container.

Screen Shot 2023-06-02 at 1 33 23 PM

But it can be customized. There is an option to edit the panel and change the legend.  In this case, I only want to show the container name, like so:

Screen Shot 2023-06-02 at 1 34 21 PM

It seems ok to leave the legend as it is for now.

jgwest (approved):
LGTM, will follow up on new epic for productizing postgresql exporter, thanks @keithchong and thanks @jopit for reviewing!

PR #460, opened 2023-05-12
Prevent excessive status updates on  .status conditions of SnapshotEnvironmentBinding/Environment resources

jgwest:
#### Description:
- Add additional checks to prevent calls to `k8sClient.Status().Update` in cases where the `.status.bindingConditions` field was not actually modified.
- Remove duplication from the condition update logic of the SnapshotEnvironmentBinding/Environment CRs, since both CRs are doing the exact same thing (updating []metav1.Condition)
    - (It would not surprise me if there existed a utility function in k8s api packages that did this, but one thing at a time)
- Add new unit tests for above
- Added short log statements to indicate when the .status.bindingConditions field is being updated, to help us detect this during dev
- Fix status-update bug in our proxy client (proxy client is used in various unit tests)

#### Link to JIRA Story (if applicable): [GITOPSRVCE-584](https://issues.redhat.com/browse/GITOPSRVCE-584)

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
@@ -177,46 +177,31 @@ const (
 	EnvironmentReasonErrorOccurred                   = "ErrorOccurred"
 )
 
-// Update Status.Condition field of Environment
+// Update .status.conditions field of Environment
 func updateStatusConditionOfEnvironment(ctx context.Context, client client.Client, message string,
 	environment *appstudioshared.Environment, conditionType string,
 	status metav1.ConditionStatus, reason string, log logr.Logger) error {
-	// Check if condition with same type is already set, if Yes then check if content is same,
-	// If content is not same update LastTransitionTime
-	index := -1
-	for i, Condition := range environment.Status.Conditions {
-		if Condition.Type == conditionType {
-			index = i
-			break
+
+	newCondition := metav1.Condition{
+		Type:    conditionType,
+		Message: message,
+		Status:  status,
+		Reason:  reason,
+	}
+
+	changed, newConditions := insertOrUpdateConditionsInSlice(newCondition, environment.Status.Conditions)
+
+	if changed {
+		environment.Status.Conditions = newConditions
+
+		if err := client.Status().Update(ctx, environment); err != nil {
+			log.Error(err, "unable to update environment status condition.")
+			return err
 		}
 	}
 
-	now := metav1.Now()
-
-	if index == -1 {
-		environment.Status.Conditions = append(environment.Status.Conditions,
-			metav1.Condition{
-				Type:               conditionType,
-				Message:            message,
-				LastTransitionTime: now,
-				Status:             status,
-				Reason:             reason,
-			})
-	} else if environment.Status.Conditions[index].Message != message ||
-		environment.Status.Conditions[index].Reason != reason ||
-		environment.Status.Conditions[index].Status != status {
-		environment.Status.Conditions[index].Reason = reason
-		environment.Status.Conditions[index].Message = message
-		environment.Status.Conditions[index].Status = status
-		environment.Status.Conditions[index].LastTransitionTime = now
-	}
-
-	if err := client.Status().Update(ctx, environment); err != nil {
-		log.Error(err, "unable to update environment status condition.")
-		return err
-	}
-
 	return nil
+
 }
 
 // Update status condition when error is resolved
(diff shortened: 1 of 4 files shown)


appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -177,46 +177,31 @@ const (
...
-			})
-	} else if environment.Status.Conditions[index].Message != message ||
-		environment.Status.Conditions[index].Reason != reason ||
-		environment.Status.Conditions[index].Status != status {
-		environment.Status.Conditions[index].Reason = reason
-		environment.Status.Conditions[index].Message = message
-		environment.Status.Conditions[index].Status = status
-		environment.Status.Conditions[index].LastTransitionTime = now
-	}
+	if changed {
```
keithchong, line 194: Just a clarification question:  So before, we were always making the client.Status().Update call regardless if the bindingCondition changed or not. (Hence the high put request rate).  At this point, we will only call it if it only changed....right?

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
```diff
@@ -1177,6 +1179,186 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler Tests", func() {
...
+				InnerClient: fakeK8sClient,
+				Informer:    &eventReceiver,
+			}
+
+		})
+
+		DescribeTable("verify updateBindingConditionOfSEB works as expected",
+			func(preCondition []metav1.Condition, newCondition metav1.Condition, expectedResult []metav1.Condition, expectUpdateStatusCalled bool) {
+
+				By("creating an initial SEB with the pre- state of bindingConditions")
```
keithchong, line 1222: pre-state?

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
```diff
@@ -1177,6 +1179,186 @@ var _ = Describe("SnapshotEnvironmentBinding Reconciler Tests", func() {
...
+					},
+				}
+
+				err := k8sClient.Create(ctx, &env)
+				Expect(err).To(BeNil())
+				env.Status.BindingConditions = preCondition
+				err = k8sClient.Update(ctx, &env)
+				Expect(err).To(BeNil())
+
+				By("reseting the list of events after we updated the binding conditions")
```
keithchong, line 1236: resetting

backend-shared/util/proxyclient.go
```diff
@@ -222,8 +222,13 @@ func (pc *ProxyClient) DeleteAllOf(ctx context.Context, obj client.Object, opts
 
 // StatusWriter knows how to update status subresource of a Kubernetes object.
 func (pc *ProxyClient) Status() client.StatusWriter {
-	res := pc.InnerClient.Status()
-	return res
+
+	wrapper := ProxyClientStatusWrapper{
```
keithchong, line 226: Could you elaborate on why a wrapper is needed?  I'm not quite sure what the issue is from the PR description.

keithchong (commented):
Hi @jgwest , just a couple of minor comments, and questions for my understanding of the problem.  Otherwise LGTM.  The test looks ok, I ran it and it passed.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
jgwest: Yup, exactly.

appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller_test.go
jgwest: Ugh, the english language is the worst :stuck_out_tongue:

backend-shared/util/proxyclient.go
jgwest: For unit tests, this is used to keep track of how many/which calls are made to the fake K8s client. In the case of this test, we use to verify that the status field is not updated.

(But we could likely instead use a mocking framework for this same test, rather than using proxy client, if we wanted.)

jgwest:
Thanks @keithchong! In the interests of time, and since the requested changes are non-functional, I'm going to merge the PR as is, and roll your requested changes into a commit I am pushing to https://github.com/redhat-appstudio/managed-gitops/pull/450 (which is also in the interests of time)

PR #464, opened 2023-05-15
Delete GitOpsDeploymentManagedEnvironment when the corresponding Environment resource is deleted

Rizwana777:
#### Description:
- Deleting GitOpsDeploymentManagedEnvironment in environment_controller.go when Environment Resource doesn't exists
- Updated unit test as above

https://issues.redhat.com/browse/GITOPSRVCE-562

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
@@ -93,28 +93,50 @@ func (r *EnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			Namespace: req.Namespace,
 		},
 	}
-	gitOpsDeplManagedEnv := managedgitopsv1alpha1.GitOpsDeploymentManagedEnvironment{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "managed-environment-" + environment.Name,
-			Namespace: environment.Namespace,
-		},
-	}
+
 	if err := rClient.Get(ctx, client.ObjectKeyFromObject(environment), environment); err != nil {
-		if apierr.IsNotFound(err) {
-			log.Info("Environment resource no longer exists")
-			// A) The Environment resource could not be found: the owner reference on the GitOpsDeploymentManagedEnvironment
-			// should ensure that it is cleaned up, so no more work is required.
-			// As the environment resource no longer exists, the corresponding GitOpsDeploymentManagedEnvironment should be deleted.
-			if err = rClient.Get(ctx, client.ObjectKeyFromObject(&gitOpsDeplManagedEnv), &gitOpsDeplManagedEnv); err == nil {
-				if err = rClient.Delete(ctx, &gitOpsDeplManagedEnv); err != nil {
-					log.Info("Unable to delete GitOpsDeploymentManagedEnvironment")
-				}
-				log.Info("The GitOpsDeploymentManagedEnvironment corresponding to the Environment resource has been deleted.")
+
+		if !apierr.IsNotFound(err) {
+			// If a generic error occurred, return it
+			log.Error(err, "unable to retrieve Environment resource")
+			return ctrl.Result{}, fmt.Errorf("unable to retrieve Environment resource: %v", err)
+		}
+
+		// If the Environment resource no longer exists...
+
+		gitOpsDeplManagedEnv := generateEmptyManagedEnvironment(environment.Name, environment.Namespace)
+
+		// A) The Environment resource could not be found: As the environment resource no longer exists, the
+		// corresponding GitOpsDeploymentManagedEnvironment should be deleted.
+		if err := rClient.Get(ctx, client.ObjectKeyFromObject(&gitOpsDeplManagedEnv), &gitOpsDeplManagedEnv); err != nil {
+
+			if apierr.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -83,12 +83,24 @@ func (r *EnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	if err := rClient.Get(ctx, client.ObjectKeyFromObject(environment), environment); err != nil {
-
 		if apierr.IsNotFound(err) {
 			log.Info("Environment resource no longer exists")
 			// A) The Environment resource could not be found: the owner reference on the GitOpsDeploymentManagedEnvironment
 			// should ensure that it is cleaned up, so no more work is required.
+			// As the environment resource no longer exists, the corresponding GitOpsDeploymentManagedEnvironment should be deleted.
+			if err = rClient.Get(ctx, client.ObjectKeyFromObject(&gitOpsDeplManagedEnv), &gitOpsDeplManagedEnv); err == nil {
+				if err = rClient.Delete(ctx, &gitOpsDeplManagedEnv); err != nil {
+					log.Info("Unable to delete GitOpsDeploymentManagedEnvironment")
```
keithchong, line 100: Just something minor, but it LGTM. 

Do you think it will be useful to mention the name of the environment and/or the GitOpsDeploymentMangedEnvironment that we are trying to delete?

keithchong (approved):
Hi @Rizwana777 , just one minor comment. I'll pre-approve.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
Rizwana777: I think it will be useful if something fails and we can know why it is failing like for eg, `Unable to delete GitOpsDeploymentManagedEnvironment`

jgwest (changes requested):
Thanks @Rizwana777! I've pushed a commit where I've added some additional logic, plus a unit test, to handle the case where a GitOpsDeployment is still hanging around long after the parent Environment has been deleted (using `findObjectsForGitOpsDeploymentManagedEnvironment` should catch this case). You will need to `git pull` this commit into your local branch. Feel free to let me know if I broke your logic!

One new request: can you add an E2E tests that:
- creates a GitOpsDeploymentManagedEnvironment with an ownerref to an Environment that doesn't exist
    - In this case, the new logic in the Environment controller should detect that, and delete the GitOpsDeploymentManagedEnvironment
    - The E2E test should verify that the GitOpsDeploymentManagedEnvironment is deleted in this case.

Finally, a couple points I noticed while reviewing:
- Errors should be logged with `log.Error(err, ...)`, rather than `log.Info()`
- IMHO we should prefer to use `if err :=` (rather than `if err =`) where possible

Rizwana777:
> Thanks @Rizwana777! I've pushed a commit where I've added some additional logic, plus a unit test, to handle the case where a GitOpsDeployment is still hanging around long after the parent Environment has been deleted (using `findObjectsForGitOpsDeploymentManagedEnvironment` should catch this case). You will need to `git pull` this commit into your local branch. Feel free to let me know if I broke your logic!
> 
> One new request: can you add an E2E tests that:
> 
> * creates a GitOpsDeploymentManagedEnvironment with an ownerref to an Environment that doesn't exist
>   
>   * In this case, the new logic in the Environment controller should detect that, and delete the GitOpsDeploymentManagedEnvironment
>   * The E2E test should verify that the GitOpsDeploymentManagedEnvironment is deleted in this case.
> 
> Finally, a couple points I noticed while reviewing:
> 
> * Errors should be logged with `log.Error(err, ...)`, rather than `log.Info()`
> * IMHO we should prefer to use `if err :=` (rather than `if err =`) where possible

Sure @jgwest I will add the e2e test

tests-e2e/core/managed_environment_test.go
```diff
@@ -1101,6 +1103,35 @@ var _ = Describe("Environment E2E tests", func() {
...
+			Expect(err).To(Succeed())
+
+			err = k8s.Create(&managedEnv, k8sClient)
+			Expect(err).To(BeNil())
+
+			Eventually(func() bool {
+				err = k8sClient.Get(ctx, client.ObjectKeyFromObject(&managedEnv), &managedEnv)
+				Expect(apierr.IsNotFound(err)).To(BeTrue())
+				return Expect(err).ToNot(BeNil())
+			}, 1*time.Minute)
```
jgwest, line 1132: Even easier than doing this, you can just do:
```
			Eventually(managedEnv, "60s", "1s").ShouldNot(k8s.ExistByName(k8sClient))
```

(And in general, I'm not sure if `Expect` statements work (are correct) inside `Eventually`, my guess is they don't, but I haven't actually experimented to find out :thinking: )

jgwest (changes requested):

tests-e2e/core/managed_environment_test.go
Rizwana777: @jgwest I will update this and `Expect` works as expected in `Eventually`  I have tested in some of the test cases and it worked  , but not sure whether it will work for all types

jgwest (approved):
LGTM, thanks @Rizwana777 and thanks @keithchong for reviewing!

PR #465, opened 2023-05-15
Log and return if the ApplicationID of the SyncOperation is not found

chetan-rns:
#### Description:
- The Application row could have been deleted while processing SyncRun so log and return if the ApplicationID of SyncOperation is empty.

https://issues.redhat.com/browse/GITOPSRVCE-543

### DIFF
--- a/backend/eventloop/application_event_loop/application_event_runner_syncruns.go
+++ b/backend/eventloop/application_event_loop/application_event_runner_syncruns.go
@@ -359,7 +359,7 @@ func (a *applicationEventLoopRunner_Action) handleDeletedGitOpsDeplSyncRunEvent(
 	log.Info("Received GitOpsDeploymentSyncRun event for a GitOpsDeploymentSyncRun resource that no longer exists")
 
 	if syncOperation.Application_id == "" {
-		log.Info("Application row not found for SyncOperation", "syncOperationID", syncOperation.SyncOperation_id, "applicationID", syncOperation.Application_id)
+		log.Info("Application row not found for SyncOperation. This is normally because the Application has already been deleted.", "syncOperationID", syncOperation.SyncOperation_id, "applicationID", syncOperation.Application_id)
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


backend/eventloop/application_event_loop/application_event_runner_syncruns.go
```diff
@@ -358,6 +358,11 @@ func (a *applicationEventLoopRunner_Action) handleDeletedGitOpsDeplSyncRunEvent(
 	log := a.log
 	log.Info("Received GitOpsDeploymentSyncRun event for a GitOpsDeploymentSyncRun resource that no longer exists")
 
+	if syncOperation.Application_id == "" {
+		log.Info("Application row not found for SyncOperation", "syncOperationID", syncOperation.SyncOperation_id, "applicationID", syncOperation.Application_id)
```
jgwest, line 362: Something like this might help us determine this is not an error, next time we see this message in the logs:
"Application row not found for SyncOperation: this is normally because the Application has already been deleted."

backend/eventloop/db_reconciler.go
```diff
@@ -385,6 +385,11 @@ func cleanOrphanedEntriesfromTable_ACTDM_GitOpsDeploymentSyncRun(ctx context.Con
 		return
 	}
 
+	if syncOperationDb.Application_id == "" {
```
jgwest, line 388: Rather than exiting here, we likely want to just skip the parts of this function that require this field to be non-empty (correct me if I'm wrong!)

For example, when Application_id == "", we probably still want to do most of the rest of the function:
- Update DesiredState of SyncOperation
- Create an Operation pointing to the SyncOperation
- Delete the SyncOperation

I'm actually slightly surprised the E2E tests passed with this change :thinking:, I presume because we aren't testing SyncOperation where application_id empty...

jgwest (changes requested):

backend/eventloop/db_reconciler.go
chetan-rns: Yeah. Since it's not possible to create operations without accessing the GitOpsEngineInstance, we can directly delete the resources without creating an operation.

PR #466, opened 2023-05-15
Remove unused ApiExport references

chetan-rns:
#### Description:
- Since we no longer support KCP we can remove the ApiExport flag and its references from the codebase.
- Revert back to creating a standard manager directly in main.go.


jgwest (approved):

PR #468, opened 2023-05-16
Temporarily disable infra-deployments repo update

jgwest:
#### Description:
- Pushes to the `main` branch should not update the infra-deployments repo: instead we should only rely on pushes to the `release-summit-2023` branch to update the infra-deployments repo
- This will be re-enabled once summit is over (and the summit PipelineRun and branch will be removed).


jgwest:
Merging w/o reviews or tests as this is a change to CI yaml only.

PR #469, opened 2023-05-17
Convert SpaceRequest secret into a ManagedEnvironment secret

chetan-rns:
#### Description:
- The Environment controller should convert a SpaceRequest secret into a ManagedEvironment secret.

https://issues.redhat.com/browse/GITOPSRVCE-586

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
@@ -29,6 +29,7 @@ import (
 	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -403,6 +404,16 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
 			Labels: map[string]string{
 				managedEnvironmentSecretLabel: env.Name,
 			},
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         env.APIVersion,
+					Kind:               env.Kind,
+					Name:               env.Name,
+					UID:                env.UID,
+					BlockOwnerDeletion: pointer.Bool(true),
+					Controller:         pointer.Bool(true),
+				},
+			},
 		},
 		Type: sharedutil.ManagedEnvironmentSecretType,
 	}
@@ -425,6 +436,8 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
 				}
 			}
 
+			logutil.LogAPIResourceChangeEvent(managedEnvSecret.Namespace, managedEnvSecret.Name, managedEnvSecret, logutil.ResourceDeleted, log)
+
 			return nil, true, fmt.Errorf("the secret '%s' referenced by the Environment resource was not found: %v", secret.Name, err)
 		}
 
@@ -438,7 +451,6 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
 		return nil, true, err
 	}
 
-	// We should only reconcile secrets created by the SpaceRequest controller.
 	managedEnv := generateEmptyManagedEnvironment(env.Name, env.Namespace)
 
 	// We only want to reconcile managed environment secrets for secrets coming from SpaceRequest.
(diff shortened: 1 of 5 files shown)


appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -386,6 +395,18 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
...
 		},
 	}
+
+	managedEnvSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generateManagedEnvSecretName(env.Name),
+			Namespace: secret.Namespace,
+			Labels: map[string]string{
+				managedEnvironmentSecretLabel: env.Name,
+			},
```
jgwest, line 405: Thinking out loud, could we add an ownerref from the  Secret back to the Environment? This might aid cleanup, and I don't think there are any cases where the lifecycle of the ManagedEnvironment Secret needs to outlive the Environment itself.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -397,6 +418,13 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
 				return nil, true, fmt.Errorf("unable to update environment status condition. %v", err)
 			}
 
+			// Delete the managed Environment secret if the orginal secret is not found.
+			if err := k8sClient.Delete(ctx, &managedEnvSecret); err != nil {
+				if !apierr.IsNotFound(err) {
+					return nil, true, fmt.Errorf("unable to delete the secret for managed Environment: %s", env.Name)
+				}
+			}
+
```
jgwest, line 427: We should add a call to `LogAPIResourceChangeEvent` for this case (or any cases where we are creating/modifying/deleting a resource in an API namespace)

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -410,13 +438,42 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
...
+			if !apierr.IsNotFound(err) {
+				return nil, false, fmt.Errorf("failed to fetch the secret %s for managed Environment %s: %v", managedEnvSecret.Name, managedEnv.Name, err)
+			}
+
+			// Create a new managed environment secret if it is not found
+			managedEnvSecret.Data = secret.Data
+			if err := k8sClient.Create(ctx, &managedEnvSecret); err != nil {
+				return nil, false, fmt.Errorf("failed to create a secret for managed Environment %s: %v", managedEnv.Name, err)
+			}
+			log.Info("Created a new secret for managed Environment", "ManagedEnvironment", managedEnv.Name, "Environment", env.Name)
```
jgwest, line 457: As above, can add a call to LogAPIResourceChangeEvent for this block.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -410,13 +438,42 @@ func generateDesiredResource(ctx context.Context, env appstudioshared.Environmen
...
+		} else {
+			// The managed Environment secret is found. Compare it with the original secret and update if required.
+			if !reflect.DeepEqual(secret.Data, managedEnvSecret.Data) {
+				managedEnvSecret.Data = secret.Data
+				if err := k8sClient.Update(ctx, &managedEnvSecret); err != nil {
+					return nil, false, fmt.Errorf("failed to update the secret for managed Environment %s: %v", managedEnv.Name, err)
+				}
+				log.Info("Updated the managed Environment secret", "ManagedEnvironment", managedEnv.Name)
```
jgwest, line 465: As above, can add a call to LogAPIResourceChangeEvent for this block.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
```diff
@@ -541,3 +607,90 @@ func (r *EnvironmentReconciler) findObjectsForDeploymentTarget(dt client.Object)
...
+						Name:      envName,
+						Namespace: secret.GetNamespace(),
+					},
+				},
+			}
+		}
+		return []reconcile.Request{}
+	}
+
+	// If the secret is created by the SpaceRequest controller, find the corresponding Environment.
```
jgwest, line 642: Looking at what we're doing here, I'm slightly worried about the amount of work our Environment controller has to do.

AFAICT this code will get called for any Secret that is created/modified on the cluster, rather than just Secrets that are related to our (or space request) API (and RHTAP Prod has ~2500 Secrets).  And for each Secret event we receive, the code we run looks pretty heavy: 2 (namespace-scoped) lists and 1 or more Gets (though hopefully the controller-runtime cache helps reduce this).

In order to significantly reduce this issue, I think we can filter out Secrets that have any of these types:
```
kubernetes.io/dockerconfigjson
kubernetes.io/service-account-token
kubernetes.io/dockercfg
kubernetes.io/tls
```

That reduces the number of Secrets that our code will process event for from ~2500 to (currently) about ~350.

WDYT?

jgwest (changes requested):
Great work, thanks @chetan-rns! Just some very minor changes requested, then good to go.

appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
chetan-rns: Good idea. I've added a check to filter the secrets of other types.

jgwest (approved):
Looks great, thanks again @chetan-rns!

PR #470, opened 2023-05-17
Fix the placement of the error parameter

Rizwana777:
#### Description:
- In backend , there are some places where error was placed as first parameter, fixed the error placement


jgwest (approved):
LGTM too, thanks @Rizwana777! And thanks @jparsai for reviewing!

PR #477, opened 2023-05-31
Add E2E test for Argo CD query-scoped GitOpsDeployments

jgwest:
#### Description:
- New E2E test that verifies that using `?managedEnvironment=(managed environment id)` at the end of a Argo CD cluster Secret URL will cause Argo CD to treat the cluster as a separate cluster.
- Some refactoring of shared functions into fixture (the code for the functions themselves is unchanged)


jparsai:
LGTM @jgwest 
/lgtm

jgwest:
Thanks for reviewing @jparsai!

PR #478, opened 2023-06-01
Remove DTClass from appstudio-e2e tests overlay

jgwest:
#### Description:
- The `DeploymentTargetClass` is not needed to be defined when running the appstudio-e2e tests, as part of the infra-deployments repository.


jgwest:
Merging as YAML change only.

PR #479, opened 2023-06-01
Switch to cluster-scoped Argo CD (#560)

keithchong:
#### Description:

See JIra for details of changes

https://issues.redhat.com/browse/GITOPSRVCE-560

### DIFF
--- a/manifests/scripts/openshift-argo-deploy/openshift-gitops-subscription.yaml
+++ b/manifests/scripts/openshift-argo-deploy/openshift-gitops-subscription.yaml
@@ -13,3 +13,5 @@ spec:
     env:
       - name: DISABLE_DEFAULT_ARGOCD_INSTANCE
         value: "true"
+      - name: ARGOCD_CLUSTER_CONFIG_NAMESPACES
+        value: "gitops-service-argocd"
--- a/tests-e2e/argocd/argocd_instance_test.go
+++ b/tests-e2e/argocd/argocd_instance_test.go
@@ -74,9 +74,6 @@ var _ = Describe("Standalone ArgoCD instance E2E tests", func() {
 			destinationNamespace := &corev1.Namespace{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: fixture.NewArgoCDInstanceDestNamespace,
-					Labels: map[string]string{
-						"argocd.argoproj.io/managed-by": argocdNamespace,
-					},
 				},
 			}
 			err = k8sClient.Create(ctx, destinationNamespace)
--- a/tests-e2e/core/gitopsdeployment_test.go
+++ b/tests-e2e/core/gitopsdeployment_test.go
@@ -820,9 +820,6 @@ var _ = Describe("GitOpsDeployment E2E tests", func() {
 				namespace := &corev1.Namespace{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: fmt.Sprintf("user-%d", i),
-						Labels: map[string]string{
-							"argocd.argoproj.io/managed-by": "gitops-service-argocd",
-						},
 					},
 				}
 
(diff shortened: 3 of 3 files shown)


jgwest:
```

Summarizing 1 Failure:
  [FAIL] GitOpsDeploymentSyncRun E2E tests Create, update and delete GitOpsDeploymentSyncRun [It] deleting the GitOpsDeploymentSyncRun should terminate a running Sync operation
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_syncrun_test.go:265 
  
  ```

/retest

**Review by jgwest — CHANGES_REQUESTED** [review]
Hey Keith, for this story we also need to remove the labels in this block (around 427 in `tests-e2e/fixture/fixture.go`):
```go
	// Create the namespace again
	_, err = kubeClientSet.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
		Name: namespaceParam,
		// \/ \/ \/
		// Labels: map[string]string{
		// 	"argocd.argoproj.io/managed-by": argoCDNamespaceParam,    # <=== these lines needed to be commented out
		// },
		// /\ /\ /\
	}}, metav1.CreateOptions{})
	if err != nil {
		return fmt.Errorf("unable to create namespace '%s': %v", namespaceParam, err)
	}
```

I've pushed a commit which does this, plus adds some error statements. However, it seems like the E2E tests are failing once this label is removed. The test appear to be failing to missing Argo CD permissions (which I wouldn't expect with cluster-scoped Argo CD :thinking: ).

manifests/base/gitops-service-argocd/base/argo-cd-clusterrole.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: appstudio-openshift-gitops-argocd-application-controller
+rules:
```
jgwest, line 5: Something that occurred to me while reviewing the story: Rather than adding * * * cluster role, it likely makes more sense to scope Argo CD to only deploying a specific set of resources.

I think we can copy the list of resources from here:
https://github.com/redhat-appstudio/managed-gitops/blob/753e58188a55a7627f056bb8c2117ebdc7af69e9/manifests/base/gitops-service-argocd/base/argo-cd.yaml#L10

The idea is that this is a security defense-in-depth, to make sure that Argo CD only has write accesses to the reasons that we want it to. The list defined in the file above is the current list of resources that Argo CD can deploy.

tests-e2e/fixture/fixture.go
```diff
@@ -425,71 +425,71 @@ func EnsureDestinationNamespaceExists(namespaceParam string, argoCDNamespacePara
...
-		if count >= 2 {
-			// We expect at least 2 roles
-			return true, nil
-		}
-
-		return false, nil
-	}); err != nil {
-		return fmt.Errorf("argo CD never setup rolebindings for namespace '%s': %v", namespaceParam, err)
-	}
+		return fmt.Errorf("unable to create namespace '%s': %v", namespaceParam, err)
```
jgwest, line 433: Rather than commenting these out (or deleting them), lets move them into their own function. We don't need to call that function, but we can at least have the function lying around if we need this functionality again.

jgwest (changes requested):
Looks great, just some minor tweaks then good to go!

manifests/base/gitops-service-argocd/base/argo-cd-clusterrole.yaml
keithchong: i tried this but the tests failed.  I added the CR/CRBs for the Argo CD server as well, taken from the roles and bindings created by the gitops operator.  The tests passed after this.  WDYT?

jgwest (approved):
LGTM, thanks again @keithchong!

PR #483, opened 2023-06-06
Temporarily remove environment webhook

jgwest:
#### Description:
- Temporarily remove the Webhook manifest, to avoid issue reported on RHTAPBUGS


jgwest:
Merging w/o reviews and tests as this is already breaking RHTAP clusters.

PR #487, opened 2023-06-14
Create an E2E test to verify AppProject behaviour which we rely on

jopit:
#### Description:
- Added four new e2e tests to ensure that Argo CD behaves as expected when the repo url specified in an Application matches or doesn't match the url specified in the AppProject to which the Application belongs.

### DIFF
--- a/tests-e2e/argocd/argocd_appproject_test.go
+++ b/tests-e2e/argocd/argocd_appproject_test.go
@@ -91,7 +91,6 @@ var _ = Describe("ArgoCD AppProject E2E tests", func() {
 			}
 
 			secret = corev1.Secret{
-				TypeMeta: metav1.TypeMeta{},
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "test-app-project-secret",
 					Namespace: app.Namespace,
--- a/tests-e2e/argocd/argocd_appproject_test.go
+++ b/tests-e2e/argocd/argocd_appproject_test.go
@@ -30,6 +30,18 @@ var _ = Describe("ArgoCD AppProject E2E tests", func() {
 			secret     corev1.Secret
 		)
 
+		cleanup := func() {
+			// Ensure there's no lingering app project or secret
+			err := k8sClient.Delete(ctx, &appProject)
+			if err != nil {
+				Expect(errors.IsNotFound(err)).To(BeTrue())
+			}
+			err = k8sClient.Delete(ctx, &secret)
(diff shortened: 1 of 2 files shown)


tests-e2e/argocd/argocd_appproject_test.go
```diff
@@ -0,0 +1,196 @@
...
+						Namespace: fixture.GitOpsServiceE2ENamespace,
+					},
+					SyncPolicy: &appv1.SyncPolicy{
+						Automated: &appv1.SyncPolicyAutomated{},
+					},
+				},
+			}
+
+			secret = corev1.Secret{
+				TypeMeta: metav1.TypeMeta{},
```
Rizwana777, line 94: I think this `TypeMeta` can be removed

Rizwana777:
@jopit left a minor comment and everything LGTM, thank you

tests-e2e/argocd/argocd_appproject_test.go
```diff
@@ -0,0 +1,195 @@
...
+					appFixture.HaveSyncStatusCode(appv1.SyncStatusCodeSynced),
+					appFixture.HaveHealthStatusCode(health.HealthStatusHealthy),
+				))
+		})
+
+		// This test requires the following environment variable to be set.
+		//		GITHUB_SSH_KEY: The SSH key of the private GitHub repository
+		// If the variable is not set, the test will be skipped.
+		It("Should succeed if the app project and the application both use a git url", func() {
+			const giturl = "git@github.com:managed-gitops-test-data/private-repo-test.git"
```
jgwest, line 134: I'm 90% sure you already answered this in standup a little while ago, but it's always good to double check:
- I presume it's not possible to use 'git@github.com:redhat-appstudio/managed-gitops.git' here? (IIRC you said it's not)
- If my presumption is correct, could you add a quick little comment to the test that mentions that (that way the next person that comes to this test will know why we're doing it this way).

jgwest (changes requested):
Looks perfect, thanks @jopit! And thanks @Rizwana777 for reviewing! One quick question (which I think has already been answered) then good to go.

tests-e2e/argocd/argocd_appproject_test.go
jopit: I've added a comment to explain.

jgwest (approved):
LGTM, thanks again @jopit!

PR #489, opened 2023-06-15
Add new database table ApplicationOwner

Rizwana777:
#### Description:
- Added new database table - ApplicationOwner
- Unit tests

https://issues.redhat.com/browse/GITOPSRVCE-602


backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -321,6 +321,19 @@ func (a applicationEventLoopRunner_Action) handleNewGitOpsDeplEvent(ctx context.
 	}
 	a.log.Info("Created new Application in DB: "+application.Application_id, application.GetAsLogKeyValues()...)
 
+	// Create ApplicationOwner row in DB
+	applicationOwner := db.ApplicationOwner{
+		ApplicationOwnerApplicationID: application.Application_id,
+		ApplicationOwnerUserID:        clusterUser.Clusteruser_id,
+	}
+
+	if err := dbQueries.CreateApplicationOwner(ctx, &applicationOwner); err != nil {
```
jopit, line 330: Based on the discussion in standup, it's possible that the application owner row already exists, so the code should check for that and skip the creation in that case.

Rizwana777:
@jopit thank you for reviewing PR, updated changes , PTAL

backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
```diff
@@ -956,5 +966,66 @@ var _ = Describe("ApplicationEventLoop Handle deployment modified Test", func()
 			Expect(res).To(Equal(deploymentModifiedResult_NoChange),
 				"since the Namespace is being deleted, the request should not be acted upon")
 		})
+
+		It("Should verify whether ApplicationOwner row has been created, retrieved and deleted.", func() {
+			By("Create new deployment.")
```
jgwest, line 971: One more case I think we can test here:
1) Create a GitOpsDeployment/Application/ApplicationOwner like normal, like with the other tests in this file.
2) Delete the ApplicationOwner table, pointing to that Application.
3) Call 'applicationEventRunner_handleDeploymentModified()' and verify that ApplicationOwner is recreated

This is testing the update case: if we get a request for an Application, but the Application doesn't have an ApplicationOwner (for whatever reason), then we should make sure we create it. The code already does this, so it's just a matter of making sure we have a test case for this as well.

backend/eventloop/application_event_loop/application_event_runner_deployments_test.go
```diff
@@ -926,5 +936,66 @@ var _ = Describe("ApplicationEventLoop Handle deployment modified Test", func()
...
+			By("Verify whether ApplicationOwner has been deleted")
+			err = k8sClient.Delete(ctx, gitopsDepl)
+			Expect(err).To(BeNil())
+
+			_, _, _, message, userDevErr = appEventLoopRunnerAction.applicationEventRunner_handleDeploymentModified(ctx, dbQueries)
+			Expect(userDevErr).To(BeNil())
+			Expect(message).To(Equal(deploymentModifiedResult_Deleted))
+
+			err = dbQueries.GetApplicationOwnerByPrimaryKey(context.Background(), &applicationOwner)
+			Expect(err).ToNot(BeNil())
```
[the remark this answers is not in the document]
jgwest: I think this one is still open to do.

jgwest (changes requested):
Looks great, thanks @Rizwana777 , and thanks @jopit  for reviewing!

Here are a couple items that came to mind while reviewing:
		
### 1) 

As with other with cases where we are adding new database tables to the DB, we should add a migration_test to `utilities/db-migration/migration_test`

### 2) 

This is one that should have been in the original story description, but I only remembered it as being necessary while reviewing another PR in the last sprint:

In 'cleanOrphanedEntriesfromTable_Application' in db_reconciler.go of 'backend' component, we are scanning through the database for any Application rows that are 'orphaned', that is, they are not referenced by a GitOpsDeployment (and thus the Application row should be deleted.)

When we find an Application row, to delete, we call `deleteDbEntry` on the Application. However, now that we have implemented ApplicationOwner, ApplicationOwner will have a foreign key to Application; this foreign key from ApplicationOwner to Application will prevent the Application from being deleted.

Thus, in 'cleanOrphanedEntriesfromTable_Application', we should find and delete the ApplicationOwner that points to the Application, before we delete the Application itself.

- We should also a unit test to `db_reconciler_test.go` that verifies this case: when an orphaned Application is deleted, we should make sure the ApplicationOwner is deleted as well.

Rizwana777:
@jgwest thanks for the review, updated changes, PTAL

Rizwana777:
@jgwest thanks for the review, updated changes, PTAL

jgwest (approved):
Looks perfect, thanks again @Rizwana777!

PR #490, opened 2023-06-26
Unset DT claimRef when the reclaim policy is Retain

chetan-rns:
#### Description:
- When the DTC is deleted with a reclaim policy of Retain, we don't remove the matching DT. So the DT might end up pointing to a DTC that doesn't exist. This PR unsets the DT claimRef whenever a DTC is deleted with Retain policy.

Ref: https://redhat-internal.slack.com/archives/C02C3SE8QS2/p1686889611653789?thread_ts=1686730438.404799&cid=C02C3SE8QS2

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/deploymenttargetbinder_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/deploymenttargetbinder_controller.go
@@ -120,7 +120,6 @@ func (r *DeploymentTargetClaimReconciler) Reconcile(ctx context.Context, req ctr
 						return ctrl.Result{}, err
 					}
 					log.Info("DeploymentTarget is marked to Deleted", "DeploymentTarget", dt.Name)
-					return ctrl.Result{}, nil
 
 				} else if dtcls.Spec.ReclaimPolicy == applicationv1alpha1.ReclaimPolicy_Retain {
 					log.Info("ReclaimPolicy is ReclaimPolicy_Retain")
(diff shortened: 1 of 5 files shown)


appstudio-controller/controllers/appstudio.redhat.com/deploymenttargetbinder_controller.go
```diff
@@ -129,15 +128,19 @@ func (r *DeploymentTargetClaimReconciler) Reconcile(ctx context.Context, req ctr
...
-							return ctrl.Result{}, fmt.Errorf("failed to update DeploymentTarget %s in namespace %s to Released status", dt.Name, dt.Namespace)
-						}
-						return ctrl.Result{}, nil
+					}
+
+					dt.Spec.ClaimRef = ""
+					if err := r.Client.Update(ctx, dt); err != nil {
+						return ctrl.Result{}, fmt.Errorf("failed to update the claimRef: %v", err)
+					}
+
```
jgwest, line 137: On success, it couldn't hurt to log that we've updated the claimRef field of the DT. 
We should also call `logutil.LogAPIResourceChangeEvent`, as this is the mechanism for logging whenever we've updated another resource in the Namespace.

jgwest (changes requested):
Looks great! 

One question: would you object to also adding the same logic to the `DeploymentTarget` reconciler?
- E.g. on the Reconcile of a DT(): 
    - if (that DT is bound to a DTC AND the DTC no longer exists AND the class is Reclaim) then set the claimRef to "".

Why do this? Well it's very unlikely, but I can imagine a race condition where:
-  the DTC reconcile fails to update claimRef of DT 
    - (technically the finalizer on the DTC should prevent this, that's why I'm saying it's very unlikely)
- next time the DT is reconciled, we don't have logic to detect that the claimRef is invalid, and thus that field is never cleared.

I _think_ it should be simple to add this logic (to `deploymenttargetreclaimer_controller.go`), but if you think it's not worth doing, or it looks like it's more complex than expected, don't worry about it.

jgwest (approved):
Looks perfect, thanks @chetan-rns!

PR #491, opened 2023-06-27
code refactoring for better performance.

jparsai:
## Description:
This PR is to do some code refactoring in clean up job for better performance.

## JIRA Ticket:
https://issues.redhat.com/browse/GITOPSRVCE-609

### DIFF
--- a/backend/eventloop/db_reconciler.go
+++ b/backend/eventloop/db_reconciler.go
@@ -542,7 +542,7 @@ func cleanOrphanedEntriesfromTable(ctx context.Context, dbQueries db.DatabaseQue
 }
 
 // cleanOrphanedEntriesfromTable_RepositoryCredential loops through RepositoryCredentials in database and verifies they are still valid (Having entry in ACTDM). If not, the resources are deleted.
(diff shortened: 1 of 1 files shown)


backend/eventloop/db_reconciler.go
```diff
@@ -1102,11 +1107,11 @@ func getListOfCRIdsFromTable(ctx context.Context, dbQueries db.DatabaseQueries,
 			// Save list of IDs in map according to resource type.
 			for _, deplToAppMapping := range tempList {
 				if deplToAppMapping.DBRelationType == db.APICRToDatabaseMapping_DBRelationType_RepositoryCredential {
-					crIdMap[dbType_RespositoryCredential] = append(crIdMap[dbType_RespositoryCredential], deplToAppMapping.DBRelationKey)
+					crIdMap[dbType_RespositoryCredential][deplToAppMapping.DBRelationKey] = false
```
keithchong, line 1110: This is a nit, but when you reference the map, you use `_, ok := ....` so you don't check the value.  So, to maintain consistency, then should these be assigned `true` ?  That way you can check the existence by way of the key/id, or by the value?

keithchong (approved):
LGTM

jgwest (approved):
LGTM too, thanks @jparsai, and thanks @keithchong for reviewing!

PR #493, opened 2023-06-30
Remove datasources from Grafana dashboards

jgwest:
#### Description:
- Remove datasource fields from Grafana dashboard, as Grafana will just use the main data source if one exists (this was recommended by the RHTAP monitoring team)


jgwest:
Merging without tests/reviews, YAML changes only.

PR #494, opened 2023-06-30
Remove generatorOptions as it prevents dashboard refresh in RHTAP

jgwest:
#### Description:
- Remove `disableNameSuffixHash`, as recommended as a solution to the lack of dashboard refresh that was mentioned on https://github.com/redhat-appstudio/application-service/pull/318/files


jgwest:
Dashboard YAML change only, merging w/o tests.

PR #500, opened 2023-07-12
Explicitly report cert authority invalid error

jopit:
#### Description:
- When the connection to a cluster fails due to the cluster certificate being signed by an unknown authority, explicitly report this in the error status condition


jgwest (approved):
LGTM too, thanks @jopit, and thanks @jparsai for reviewing!

PR #504, opened 2023-07-17
Remove unused code and update codecov.yml

jgwest:
#### Description:
- Remove unused route code (this dates from early dates PoC days of GitOps Service)
- Add additional test-only code to `.codecov.yml`


jgwest:
Thanks @Rizwana777!

PR #505, opened 2023-07-17
Don't log the contents of non-pointer Secret objects

jgwest:
#### Description:
- At present, it looks like we're only detecting pointers to Secrets as Secrets in the `LogAPIResourceChangeEvent` code. We should instead check for both pointers and struct objects.

### DIFF
--- a/backend-shared/util/log/log.go
+++ b/backend-shared/util/log/log.go
@@ -36,8 +36,9 @@ func LogAPIResourceChangeEvent(resourceNamespace string, resourceName string, re
 		log.Error(nil, "resource passed to LogAPIResourceChangeEvent was nil")
 		return
 	}
-	_, isSecret := (resource).(*corev1.Secret)
-	if isSecret {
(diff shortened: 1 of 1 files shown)


keithchong (approved):
LGTM.

Curious, I only see the function being called from here

https://github.com/redhat-appstudio/managed-gitops/blob/b28828aef9d285e088b32bbf5823af123d8f3f96/backend-shared/util/service_account.go#L176

where we pass the pointer to the secret.

jgwest:
Hey @keithchong, thanks for your LGTM, it's also called from other places: in this case, the reason I saw this bug was due to a call to this function (logging a Secret) in the `cluster-agent` component.

keithchong (approved):
Ok, maybe its GoLand, but I did a find ref and I don't see it.  I do see other references that pass in, for example, a cluster role.

jgwest:
Thanks @keithchong!

PR #515, opened 2023-07-18
DeploymentTarget reconciler should add the DT finalizer in all cases

jgwest:
#### Description:
- Updated DeploymentTarget Reconciler so that it will ensure the finalizer is ALWAYS added to the DeploymentTarget.
- Updated unit/e2e tests to verify the finalizer is set
- Wrote a quick unit test for findDeploymentTargetsForSpaceRequests


jgwest:
Thanks for reviewing @chetan-rns!

PR #586, opened 2023-08-01
Clean up K8s resources orphaned by Argo CD

chetan-rns:
#### Description:
- Add a reconciliation loop that periodically inspects if a resource is orphaned in a cluster
- If it is orphaned, the resource will be deleted
- Update clusterole to give additional permissions(get,list,delete) for backend
- Add unit tests

https://issues.redhat.com/browse/GITOPSRVCE-653

### DIFF
--- a/backend/eventloop/cluster_reconciler.go
+++ b/backend/eventloop/cluster_reconciler.go
@@ -3,15 +3,14 @@ package eventloop
 import (
 	"context"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
 	managedgitopsv1alpha1 "github.com/redhat-appstudio/managed-gitops/backend-shared/apis/managed-gitops/v1alpha1"
 	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
+	argocdutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util/argocd"
 	logutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util/log"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -21,11 +20,8 @@ import (
 )
 
 const (
-	// Interval in Minutes to reconcile ClusterReconciler.
-	orphanedResourcesCleanUpInterval = 30 * time.Minute
-
-	// Label added to the resources managed by Argo CD.
-	argocdResourceLabel = "app.kubernetes.io/instance"
+	// Interval in Hours to reconcile ClusterReconciler.
+	orphanedResourcesCleanUpInterval = 2 * time.Hour
 )
 
 type ClusterReconciler struct {
@@ -75,7 +71,7 @@ func (c *ClusterReconciler) Start() {
 // 2. It doesn't have a corresponding GitOpsDeployment resource.
 func (c *ClusterReconciler) cleanOrphanedResources(ctx context.Context, log logr.Logger) {
 	// Use a label selector to filter resources managed by Argo CD
-	labelSelector, err := labels.Parse(argocdResourceLabel)
+	labelSelector, err := labels.Parse(argocdutil.ArgocdResourceLabel)
 	if err != nil {
 		log.Error(err, "failed to create a label selector for listing resources managed by Argo CD")
 		return
@@ -92,7 +88,7 @@ func (c *ClusterReconciler) cleanOrphanedResources(ctx context.Context, log logr
 	}
 
 	for i, obj := range apiObjects {
-		appName := getArgoCDApplicationName(obj.GetLabels())
+		appName := argocdutil.GetArgoCDApplicationName(obj.GetLabels())
 		if appName != "" && obj.GetDeletionTimestamp() == nil && !strings.HasPrefix(obj.GetNamespace(), "openshift") {
 			// Check if a GitOpsDeployment exists with the UID specified in the Application name.
 			gitopsDeplList := &managedgitopsv1alpha1.GitOpsDeploymentList{}
@@ -104,7 +100,7 @@ func (c *ClusterReconciler) cleanOrphanedResources(ctx context.Context, log logr
 				continue
 			}
 
-			expectedUID := extractUIDFromApplicationName(appName)
+			expectedUID := argocdutil.ExtractUIDFromApplicationName(appName)
 			found := false
 			for _, gitopsDepl := range gitopsDeplList.Items {
 				if gitopsDepl.UID == types.UID(expectedUID) {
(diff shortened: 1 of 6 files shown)


jgwest:
@chetan-rns 

I think the way I wrote this Bug originally is kind of a (confusing, sorry!) mash between two separate problems, A and B:

A) We should delete orphaned resources that are already in the process of being deleted (they have a deletion timestamp, and also a finalizer)
- for this case, our new code would therefore just remove the finalizer.
- BUT, this DOES requires us to remove the finalizer, which standup discussions revealed we probably don't want to do (at least, at this time).

B) We should delete orphaned resources that are not in the process of being deleted (they do NOT have a deletion timestamp, and they might or might not have a finalizer)
- for this case, we should delete the resource (and not touch the finalizer, if it is set)

The bug text is mainly focused on A), it seems, but on subsequent discussions in standup, it seems like we should mainly handle problem B) for this bug/PR.

Why? Essentially, the problem you identified in standup is 100% correct: 
- If a kubernetes object does NOT have a finalizer, and it DOES have a deletion timestamp, then K8s will GC (delete) it very quickly. 
- In this case, there is nothing for us to do, it will be cleaned by K8s already.
- It is thus very difficult to test this case, and since K8s will necessarily delete that object anyways, I'm not sure we need to test it.

And since we decided in standup that we likely don't want to remove finalizers from resources we didn't create (e.g. we don't want to accidentally delete something important that is being protected by a finalizer), that likely means we can't really implement A).

Thus, we likely want to focus on B): finding K8s resources that were deployed by Argo CD, but that DON'T have deletion timestamp set.

Does this make sense in the context of the work you've done on the story so far?

chetan-rns:
@jgwest Thanks for the detailed explanation. It makes sense.  So far the work in the PR takes care of option B. Additionally, I have updated the PR with the below changes:
1. Skip resources with a deletion timestamp
2. Use a label selector in the List API instead of filtering the resources explicitly.  This reduced the time taken for bulk list from 30s to 3s.

backend/eventloop/cluster_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+
+				if err := c.client.List(ctx, objList, opts...); err != nil {
+					log.V(logutil.LogLevel_Debug).Error(err, "failed to list resources", "resource", apiResource.Kind)
+					return
+				}
+
+				mu.Lock()
+				resources = append(resources, objList.Items...)
+				mu.Unlock()
+			}(apiResources, apiResource)
```
jgwest, line 167: I'm a big fan of using goroutines/concurrency to speed up slow I/O requests :+1:, BUT :grin:,  I think in this very specific case, using a goroutine per resource might be a bit too aggressive re: issuing too many requests to the K8s API server at once.

One possible option would be to change this logic to only allow X number of concurrent requests at once. (2? 3? 5?). This would be one approach to explore, if you're interested. Another option might be investigating the throttling options of the client object.

But personally, I'd be fine with just doing it the slow, steady way: using a single thread, and just steadily going through the list.

Why? In general, orphaned Argo CD resources should be a pretty rare occurrence: the vast majority of the time Argo CD will successfully clean up all the resources of the Application, I would expect.  So I think it's perfectly fine if it takes this cleanup job a little while to find those orphaned resouces in the rare case that they do appear.

WDYT?

backend/eventloop/cluster_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// Interval in Minutes to reconcile ClusterReconciler.
+	orphanedResourcesCleanUpInterval = 30 * time.Minute
```
jgwest, line 25: I'm just pulling numbers out of my butt here :grin:, but what about every 2 hours? Scanning through all the resources of all the namespaces on the cluster is going to be a fairly intensive process, so we probably don't want to do it TOO often.

backend/eventloop/cluster_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+
+			// The resource was managed by Argo CD i.e. it has label "app.kubernetes.io/instance"
+			// But it doesn't have a corresponding GitOpsDeployment so it can be deleted.
+			if !found {
+				if err := c.client.Delete(ctx, &apiObjects[i]); err != nil {
+					log.Error(err, "failed to delete object in the orphaned reconciler", "name", obj.GetName(), "namespace", obj.GetNamespace(), "gvk", obj.GroupVersionKind())
+					continue
+				}
+
+				log.Info("Deleted an orphaned resoure that is not managed by Argo CD anymore", "Name", obj.GetName(), "Namespace", obj.GetNamespace())
```
jgwest, line 124: `resoure` => `resource`

backend/eventloop/cluster_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+				log.Error(err, "failed to list GitOpsDeployments", "namespace", obj.GetNamespace())
+				continue
+			}
+
+			expectedUID := extractUIDFromApplicationName(appName)
+			found := false
+			for _, gitopsDepl := range gitopsDeplList.Items {
+				if gitopsDepl.UID == types.UID(expectedUID) {
+					found = true
+					break
```
jgwest, line 112: This block appears not to be covered by the unit tests (correct me if I'm wrong!). It couldn't hurt to make sure that a resources is NOT deleted when there is a valid GitOpsDeployment existing for it: there IS a test for this scenario in the PR ("should not delete a resource that is still managed by GitOpsDeployment"), but it would be good to double check why that test isn't covering this block.

backend/eventloop/cluster_reconciler_test.go
```diff
@@ -0,0 +1,349 @@
...
+			reconciler.cleanOrphanedResources(ctx, logger)
+
+			err = k8sClient.Get(ctx, client.ObjectKeyFromObject(namespacedObj), namespacedObj)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+	})
+})
+
+func createFakeCluster() *httptest.Server {
```
jgwest, line 303: :+1:

jgwest (changes requested):
Looks great, thanks @chetan-rns! Just some minor request then good to go.

backend/eventloop/cluster_reconciler.go
chetan-rns: The test `should not delete a resource that is still managed by GitOpsDeploymen` will cover this case. It has a GitOpsDeployment that matches the ID from the Application so it is not deleted and the test checks that the resource still exists.

jgwest: This is what I'm seeing in 'cleanOrphanedResources', after running the tests and applying the cover.out profile in VS Code:

image

Based on this, it seems like `gitopsDeplList` never has a value (and thus also the UID check isn't tested).

chetan-rns: Fixed the issue. The test was not creating the resource properly. The code block is now covered 

Screenshot from 2023-08-10 11-42-29

jgwest:
Failing with known CI flake, which needs to be investigated: https://issues.redhat.com/browse/GITOPSRVCE-711

/retest

jgwest (approved):

PR #587, opened 2023-08-01
Set codecov to block PRs if they decrease coverage

jopit:
#### Description:
- Update .codecov.yml to block PRs from merging if they regress coverage by more than 0.2%


jgwest (approved):
Short and sweet, thanks @jopit!

PR #588, opened 2023-08-02
Use WithCallDepth in LogAPIResourceChangeEvent

jopit:
#### Description:
- Changed the logger created within LogAPIResourceChangeEvent to use the location of the caller instead of the location within LogAPIResourceChangeEvent itself.


jgwest (approved):
LGTM, thanks @jopit!

PR #594, opened 2023-08-06
Fix race condition in AppProjectRepository reconciliation and remove RepositoryCredential field

jgwest:
#### Description:
- The logic for determining what `AppProjectRepository` rows should exist in the database is now entirely within the `reconcileAppProjectRepositories` function of the shared resource loop
- This avoids a potential race condition where multiple GitOpsDeployments/RepositoryCredentials were reconciling AppProjects at the same time.
    - To be clear: this race condition was due to a flaw in *my* proposal in the original design document, and not an issue with the implementation itself by the GitOps Service EE team.
- The existing AppProjectRepository reconcile functions now call the shared resource loop function

### DIFF
--- a/backend-shared/db/app_project_repository.go
+++ b/backend-shared/db/app_project_repository.go
@@ -133,23 +133,23 @@ func (dbq *PostgreSQLDatabaseQueries) UpdateAppProjectRepository(ctx context.Con
 
 }
 
-func (dbq *PostgreSQLDatabaseQueries) DeleteAppProjectRepositoryByRepoCredId(ctx context.Context, obj *AppProjectRepository) (int, error) {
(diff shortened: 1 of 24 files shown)


jgwest:
```

Summarizing 1 Failure:
  [FAIL] GitOpsDeployment E2E tests Create, Update and Delete a GitOpsDeployment  [It] should ensure that when 2 GitOpsDeployments are created and point to the same repo url, and one is deleted, the AppProjectRepository for the other still exists
  /go/src/github.com/redhat-appstudio/managed-gitops/tests-e2e/core/gitopsdeployment_test.go:272
Ran 57 of 60 Specs in 2286.788 seconds
FAIL! -- 56 Passed | 1 Failed | 0 Pending | 3 Skipped
```

jgwest:
Known CI flake
/retest

Rizwana777:
@jgwest LGTM, just a minor typos

PR #601, opened 2023-08-11
Move backend-shared test utils code into its own file, update codecov.yml

jgwest:
#### Description:
- `backend-shared/db/utils.go` contained a large block of code that is only used by unit/E2E tets.
- I have moved that code into its own file, so that I could exclude that file in codecov.yml.
- I also added one more file to codecov.yml, and fixed a YAML formatting error.


jgwest:
Thanks @jparsai!

PR #603, opened 2023-08-11
Don't use variable to specify FROM image in Dockerfiles

jopit:
#### Description:
- Explicitly specify the image to use in FROM statement in the Dockerfiles. Dependabot doesn't seem to handle specifying the image via an ARG.


jgwest (approved):
LGTM, thanks @jopit!

PR #604, opened 2023-08-14
Remove secret_controller and test; Use EnqueueRequestsFromMapFunc (#670)

keithchong:
#### Description:
- The reasons for reconciliation of the managed environment appears to be changed in the Jira description, in that it no longer checks if the namespace matches.  If this is no longer required, we simply remove the if block (I kept it in)
- The construction of the Namespace in the original code seems incorrect, in that it uses namespace.name instead of namespace.namespace.  I did not change this behaviour, but if you think it should remain at namespace.name, then we should leave it.

- See https://issues.redhat.com/browse/GITOPSRVCE-560


backend/controllers/managed-gitops/gitopsdeploymentmanagedenvironment_controller_test.go
```diff
@@ -29,9 +31,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 var _ = Describe("GitOpsDeploymentManagedEnvironment Controller Test", func() {
+	//ctx := context.Background()
```
jgwest, line 38: Can remove this if no longer needed

jgwest (changes requested):
Looks great, thanks @keithchong! A minor tweak then good to go.

backend/controllers/managed-gitops/gitopsdeploymentmanagedenvironment_controller_test.go
keithchong: Done

backend/controllers/managed-gitops/gitopsdeploymentmanagedenvironment_controller_test.go
```diff
@@ -67,25 +73,142 @@ var _ = Describe("GitOpsDeploymentManagedEnvironment Controller Test", func() {
...
+
+					createManagedEnvTargetingSecret("testManagedEnv", secret, *namespace, k8sClient)
+
+					Expect(reconciler.findSecretsForManagedEnvironment(&secret)).To(BeEmpty())
+				})
+			})
+
+			When("Five GitOpsDeploymentManagedEnvironment reference a secret in the same namespace", func() {
+
+				It("should return two managed environments", func() {
```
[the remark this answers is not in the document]
keithchong: Done

jgwest (approved):
LGTM, thanks @keithchong! And thanks for your patience in getting this approved :+1:

PR #605, opened 2023-08-14
Allow self-signed certificates in DT provisioner, on dev/test envs

jgwest:
#### Description:
- When running on dev/test K8s environment, where the K8s API URL is potentially self-signed, we need to tell the DeploymentTarget provisioner to allow self-signed certifcates.


jgwest:
Merging w/o reviews: YAML change only.

PR #612, opened 2023-08-17
Add unit tests for LogAPIResourceChangeEvent function

Rizwana777:
https://issues.redhat.com/browse/GITOPSRVCE-689


backend-shared/util/log/log_test.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+			mockLog.EXPECT().WithValues("audit", "true").Return(mockLog)
+			mockLog.EXPECT().Enabled(gomock.Any()).Return(true)
+			mockLog.EXPECT().Info(0, fmt.Sprintf("API Resource changed for secret resource: %s, name: %s, namespace: %s", ResourceModified, resourceName, resourceNamespace))
+
+			LogAPIResourceChangeEvent(resourceNamespace, resourceName, secret, ResourceModified, logger)
+		})
+	})
+
+	Context("when resource is not a Secret", func() {
```
jopit, line 59: Since this is testing that an error is logged when the resource can't be marshalled, I think the `Context` and the `It` should describe that specifically, something like
```
Context("when the resource can't be marshalled") {
    It("should log a marshalling error") {
```

jopit (changes requested):

backend-shared/util/log/log_test.go
Rizwana777: yes, updated the changes, PTAL

jopit (approved):
LGTM

jgwest (approved):
Looks perfect, thanks @Rizwana777! I have one extra semi-related request, but I will post it to the JIRA.

PR #613, opened 2023-08-17
Fix duplicate dashboard ID

jgwest:
#### Description:
- id=34 is used by two separate panels, which causes bug when trying to edit the dashboard.


jgwest:
JSON change only, merging w/o reviews/tests.

PR #614, opened 2023-08-17
Unit tests for waitForOperationToComplete, isOperationComplete (#688)

keithchong:
#### Description:
- This PR does not add the unit tests for service_account.go since it is separate
- The first test tests directly the two functions waitForOperationToComplete and isOperationComplete 
- CreateOperation is not tested thoroughly as well, in particular, there is no test that sets true for waitForOperation. 
   - This PR makes some slight tweaks in product code to accommodate this.
   - The second tests tests the 'old' CreateOperation via the new doCreateOperation.  I added two operations but only one is probably sufficient.  Let me know if you want to remove one of them

https://issues.redhat.com/browse/GITOPSRVCE-688

### DIFF
--- a/backend-shared/util/operations/types_test.go
+++ b/backend-shared/util/operations/types_test.go
@@ -2,7 +2,6 @@ package operations
 
 import (
 	"context"
-	"time"
 
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
@@ -17,8 +16,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var _ = Describe("Testing CreateOperation function.", func() {
-	Context("Testing CreateOperation function.", func() {
+var _ = Describe("Testing CreateOperation function", func() {
+	Context("Testing CreateOperation function", func() {
 		var ctx context.Context
 		var dbq db.AllDatabaseQueries
 		var dbOperationFirst *db.Operation
@@ -57,7 +56,7 @@ var _ = Describe("Testing CreateOperation function.", func() {
 				InnerClient: k8sClientOuter,
 			}
 
-			dbq, err = db.NewUnsafePostgresDBQueries(true, true)
+			dbq, err = db.NewUnsafePostgresDBQueries(false, true)
 			Expect(err).ToNot(HaveOccurred())
 
 			_, managedEnvironment, _, gitopsEngineInstance, _, err := db.CreateSampleData(dbq)
@@ -110,12 +109,9 @@ var _ = Describe("Testing CreateOperation function.", func() {
 		var k8sClient client.Client
 		var logger logr.Logger
 
-		delayUpdateOfState := func(operation *db.Operation, k8sClient client.Client, gitopsEngineInstance db.GitopsEngineInstance) {
+		simulateClusterAgent := func(operation *db.Operation, k8sClient client.Client, gitopsEngineInstance db.GitopsEngineInstance) {
 			defer GinkgoRecover()
-			// Allow waitForOperationToComplete to iterate/wait a few times before we set the operation state to Completed
-			currentTime := time.Now()
-			for time.Now().Before(currentTime.Add(2 * time.Second)) {
-			}
+
 			Eventually(func() bool {
 				operationget := db.Operation{
 					Operation_id: operation.Operation_id,
@@ -170,7 +166,7 @@ var _ = Describe("Testing CreateOperation function.", func() {
 				WithObjects(workspace, argocdNamespace, kubesystemNamespace).
 				Build()
 
-			dbq, err = db.NewUnsafePostgresDBQueries(true, true)
+			dbq, err = db.NewUnsafePostgresDBQueries(false, true)
 			Expect(err).ToNot(HaveOccurred())
 		})
 
@@ -196,7 +192,7 @@ var _ = Describe("Testing CreateOperation function.", func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			go func() {
-				delayUpdateOfState(sampleOperation, k8sClient, *gitopsEngineInstance)
+				simulateClusterAgent(sampleOperation, k8sClient, *gitopsEngineInstance)
 			}()
 
 			// Now directly call waitForOperationToComplete which calls isOperationComplete to test these functions
@@ -262,7 +258,7 @@ var _ = Describe("Testing CreateOperation function.", func() {
 			}
 
 			go func() {
-				delayUpdateOfState(sampleApplicationTypeOperation, k8sClient, *gitopsEngineInstance)
+				simulateClusterAgent(sampleApplicationTypeOperation, k8sClient, *gitopsEngineInstance)
 			}()
 
 			// Directly call CreateOperation with waitForOperation set to true, to create the new Operation
@@ -272,7 +268,7 @@ var _ = Describe("Testing CreateOperation function.", func() {
 			Expect(dbOperationFirst).NotTo(BeNil())
 
 			go func() {
-				delayUpdateOfState(sampleGitopsEngineTypeOperation, k8sClient, *gitopsEngineInstance)
+				simulateClusterAgent(sampleGitopsEngineTypeOperation, k8sClient, *gitopsEngineInstance)
 			}()
 
 			k8sOperationSecond, dbOperationSecond, err := doCreateOperation(ctx, true, *sampleGitopsEngineTypeOperation, "test-user", gitopsEngineInstance.Namespace_name, dbq, k8sClient, logger, "test-engine-operation")
--- a/backend-shared/util/operations/types_test.go
+++ b/backend-shared/util/operations/types_test.go
@@ -112,12 +112,11 @@ var _ = Describe("Testing CreateOperation function.", func() {
 
 		delayUpdateOfState := func(operation *db.Operation, k8sClient client.Client, gitopsEngineInstance db.GitopsEngineInstance) {
 			defer GinkgoRecover()
+			// Allow waitForOperationToComplete to iterate/wait a few times before we set the operation state to Completed
+			currentTime := time.Now()
+			for time.Now().Before(currentTime.Add(2 * time.Second)) {
+			}
 			Eventually(func() bool {
-				// Allow waitForOperationToComplete to iterate/wait a few times
-				currentTime := time.Now()
-				for time.Now().Before(currentTime.Add(2 * time.Second)) {
-				}
-
 				operationget := db.Operation{
 					Operation_id: operation.Operation_id,
 				}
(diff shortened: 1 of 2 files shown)


backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,188 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+				err = dbq.GetOperationById(ctx, &operationget)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(operationget.State).To(BeIdenticalTo(db.OperationState_Completed))
+
+				return true
+			}, "5s", "10ms").Should(BeTrue())
+		}
+
+		AfterEach(func() {
+			defer dbq.CloseDatabase()
```
Rizwana777, line 143: Can you please remove defer here as it is in AfterEach block?

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,188 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+			Eventually(func() bool {
+				// Allow waitForOperationToComplete to iterate/wait
+				time.Sleep(2 * time.Second)
+
+				operationget := db.Operation{
+					Operation_id: operation.Operation_id,
+				}
+
+				err := dbq.GetOperationById(ctx, &operationget)
+				// The operation might not be created yet
```
Rizwana777, line 121: I think we can add Expect statement here `Expect(err).To(HaveOccurred())`, I am not sure, please let me know if its not required

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,188 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+				WithScheme(scheme).
+				WithObjects(workspace, argocdNamespace, kubesystemNamespace).
+				Build()
+
+			k8sClient := &sharedutil.ProxyClient{
+				InnerClient: k8sClientOuter,
+			}
+
+			dbq, err = db.NewUnsafePostgresDBQueries(true, true)
+			Expect(err).ToNot(HaveOccurred())
```
Rizwana777, line 226: can move to BeforeEach block (203 - 226)

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,188 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
```
Rizwana777, line 170: I think we can move the line number 147 to 170 to BeforeEach block to avoid duplication of code and can be used in the next It("") block

Rizwana777 (changes requested):

Rizwana777:
@keithchong Left few comments, other than that everything LGTM , thank you

backend-shared/util/operations/types_test.go
keithchong: Done

backend-shared/util/operations/types_test.go
keithchong: I didn't want the test to fail if it isn't created yet, so it will eventually be true.

backend-shared/util/operations/types_test.go
keithchong: Done.  Good catch on duplicate code :-)

backend-shared/util/operations/types_test.go
keithchong: Done too.

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,167 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+			err = waitForOperationToComplete(ctx, sampleOperation, dbq, logger)
+			Expect(err).ToNot(HaveOccurred())
+
+			rowsAffected, err := dbq.DeleteOperationById(ctx, sampleOperation.Operation_id)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(rowsAffected).Should(Equal(1))
+
+		})
+
+		It("It should test CreateOperation with waitForOperation set to true to indirectly test the waitForOperationToComplete function.", func() {
```
jgwest, line 206: `It("It` -> `It("`

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,167 @@ var _ = Describe("Testing CreateOperation function.", func() {
 			Expect(dbOperationFirst.SeqID).To(Equal(dbOperationSecond.SeqID))
 		})
 	})
+
+	Context("Testing waitForOperationToComplete and IsOperationComplete function.", func() {
+		var ctx context.Context
+		var dbq db.AllDatabaseQueries
+		var k8sClient *sharedutil.ProxyClient
```
jgwest, line 110: This can just be `client.Client`, as we're not using the ProxyClient functionality in this test (likewise with removing ProxyClient references below)

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,167 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+		var logger logr.Logger
+
+		delayUpdateOfState := func(operation *db.Operation, k8sClient client.Client, gitopsEngineInstance db.GitopsEngineInstance) {
+			Eventually(func() bool {
+				// Allow waitForOperationToComplete to iterate/wait
+				time.Sleep(2 * time.Second)
```
jgwest, line 116: Would it be possible to remove the `time.Sleep`? I ask  because Sleep statements in test make me very wary (as a source for race conditions)...  I checked the logic of the test and I THINK this `time.Sleep` isn't actually doing anything, so it should be easy to remove.

jgwest (changes requested):
Looks great, thanks @keithchong and thanks @Rizwana777 for reviewing! Some small tweaks then good to go!

Re: service_account.go, I was confused, but I understand this now: I'm fine with leaving service_account.go out of this PR/story. 
- I figured out why it wasn't running in some cases: it requires a valid cluster to be defined in kubeconfig, otherwise it gets skipped (I had forgotten we had added that logic).

And one final request:
It looks like the following block in `IsOperationComplete` is not covered:
```
		// Either the operation couldn't be found (which shouldn't happen here), or some other issue, so return it
		return false, err
```
It should be a quick fix to just call IsOperationComplete with a fake `db.Operation` that doesn't actually exist in the database.

backend-shared/util/operations/types_test.go
keithchong: The purpose of the delay is to give time for the for loop in `waitForOperationToComplete` in types.go to iterate a few times. Removing the delay does seems to work ok, and the for loop gets executed once.

Try adding a Println just inside the for loop and try adjusting the delay to 2-5 seconds.  Your println statement will get executed more than once.  Thus, IsOperationComplete also gets called multiple times.

Instead of sleep, I've tried the following, and it simulates the delay too.

```
				currentTime := time.Now()
				for time.Now().Before(currentTime.Add(2 * time.Second)) {
				}
```

If you prefer, we can just remove this.

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,167 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+				// Update the operation state to Completed
+				operationget.State = db.OperationState_Completed
+				err = dbq.UpdateOperation(ctx, &operationget)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Get the operation again and check state
+				err = dbq.GetOperationById(ctx, &operationget)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(operationget.State).To(BeIdenticalTo(db.OperationState_Completed))
+
```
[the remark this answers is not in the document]
keithchong: With the existing code that runs/passes successfully, I simply changed a `ToNot` to a `To`, and reran it again, and a failure was reported.

backend-shared/util/operations/types_test.go
```diff
@@ -102,6 +103,167 @@ var _ = Describe("Testing CreateOperation function.", func() {
...
+				Resource_type:           "GitopsEngineInstance",
+				Operation_owner_user_id: "test-user",
+			}
+
+			// Create the Operation which replaces the CreateOperation step in types' CreateOperation function
+			err = dbq.CreateOperation(ctx, sampleOperation, sampleOperation.Operation_owner_user_id)
+			Expect(err).ToNot(HaveOccurred())
+
+			go func() {
+				delayUpdateOfState(sampleOperation, k8sClient, *gitopsEngineInstance)
```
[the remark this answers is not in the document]
keithchong: I will put this before Eventually.  I think Chetan also put it there.

backend-shared/util/operations/types_test.go
keithchong: Done

backend-shared/util/operations/types_test.go
keithchong: Done

keithchong:
> And one final request: It looks like the following block in `IsOperationComplete` is not covered:
> 
> ```
> 		// Either the operation couldn't be found (which shouldn't happen here), or some other issue, so return it
> 		return false, err
> ```
> 
> It should be a quick fix to just call IsOperationComplete with a fake `db.Operation` that doesn't actually exist in the database.

Done. Added a new testcase.  See latest commit.

backend-shared/util/operations/types_test.go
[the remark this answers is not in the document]
jgwest: @keithchong Yah, I'm sure I've made the mistake of forgetting to add that call a well.

backend-shared/util/operations/types_test.go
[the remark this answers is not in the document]
jgwest: @keithchong Ah, I guess one more question: if one of the `Expect` calls does fail, does it permanently fail the test (I would assume so), or does it keep trying via the loop? 

My guess is that we don't want  any of those to be Expect calls to be permanent fails (e.g. a failed Expect will immediately fail the full test), we want the loop to keep trying over and over until it succeeds (or times out and thus fails).

backend-shared/util/operations/types_test.go
jgwest: I believe you can get rid of the delay as long as you convert the Expects to return bools, as above.

keithchong: See my other comment. https://github.com/redhat-appstudio/managed-gitops/pull/614#discussion_r1313213676

jgwest:
I've rebased the PR, and added some additional code/test tweaks. The PR has been open for long enough that I would say lets just go ahead with testing within the original code coverage parameters  of the story, just so we can get this one merged and off our plates :sweat_smile:. I've updated the PR to that effect.

jgwest (approved):
LGTM, thanks @keithchong!

PR #618, opened 2023-08-21
Add unit test for generateDefaultClientForServerAddress function

chetan-rns:
#### Description:
- Add unit test for generateDefaultClientForServerAddress function

https://issues.redhat.com/browse/GITOPSRVCE-698

### DIFF
--- a/cluster-agent/utils/argocd_login_command.go
+++ b/cluster-agent/utils/argocd_login_command.go
@@ -14,7 +14,12 @@ import (
 
 // This file is loosely based on the 'argocd login' CLI command (https://github.com/argoproj/argo-cd/blob/0a46d37fc6af9fe0aa963bdd845e3d799aa0320d/cmd/argocd/commands/login.go#L60)
 
-func generateDefaultClientForServerAddress(server string, optionalAuthToken string, skipTLSTest bool) (argocdclient.Client, error) {
+type tlsConfig struct {
+	skipTLSTest bool
+	testTLS     func(addr string, dialtime time.Duration) (*grpc_util.TLSTestResult, error)
+}
+
+func generateDefaultClientForServerAddress(server string, optionalAuthToken string, tlsCfg tlsConfig) (argocdclient.Client, error) {
 
 	globalClientOpts := argocdclient.ClientOptions{
 		ConfigPath:           "",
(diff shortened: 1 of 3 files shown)


cluster-agent/utils/argocd_login_command.go
```diff
@@ -14,7 +14,12 @@ import (
 
 // This file is loosely based on the 'argocd login' CLI command (https://github.com/argoproj/argo-cd/blob/0a46d37fc6af9fe0aa963bdd845e3d799aa0320d/cmd/argocd/commands/login.go#L60)
 
-func generateDefaultClientForServerAddress(server string, optionalAuthToken string, skipTLSTest bool) (argocdclient.Client, error) {
+type tlsConfig struct {
+	skipTLSTest bool
+	testTLS     func(addr string, dialtime time.Duration) (*grpc_util.TLSTestResult, error)
+}
+
+func generateDefaultClientForServerAddress(server string, optionalAuthToken string, tlsCfg tlsConfig) (argocdclient.Client, error) {
```
chetan-rns, line 22: I have passed the `TestTLS` function as a dependency so that we can mock it in the unit tests. It's much easier than creating a fake gRPC server with TLS configurations. Let me know if this works and I'm open to suggestions if there is an easier way to test this.

cluster-agent/utils/argocd_login_command_test.go
```diff
@@ -29,4 +34,118 @@ var _ = Describe("ArgoCD Login Command", func() {
...
+				skipTLSTest: skipTLS,
+				testTLS:     mockTestTLS,
+			})
+
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(expectedErr))
+			Expect(client).To(BeNil())
+		})
+
+		It("should not return an error when insecure is true and the server certificate had errors", func() {
```
Rizwana777, line 123: I think it should ->  should not return an error when insecure `field` is true, which will be more readable, WDYT?

Rizwana777:
@chetan-rns LGTM, just added smaller comment

jgwest (approved):
LGTM, thanks @chetan-rns, and thanks @Rizwana777 for reviewing!

PR #621, opened 2023-08-22
Add unit tests for getDBOperationForEvent function in cluter-agent

Rizwana777:
#### Description:
- Did little Modification in CountOperationDBRowsByState function return value to avoid errors while creating mocks using mockgen

https://issues.redhat.com/browse/GITOPSRVCE-702


cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -2842,6 +2845,121 @@ var _ = Describe("Operation Controller", func() {
...
+
+				mockClient.EXPECT().Get(ctx, objectKey, gomock.Any()).DoAndReturn(func(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+					return nil
+				})
+				mockDB.EXPECT().GetOperationById(ctx, gomock.Any()).DoAndReturn(func(ctx context.Context, dbOp *db.Operation) error {
+					// Set the necessary fields in 'dbOp' to simulate the fetched DB operation
+					dbOp.Instance_id = "test-instance-id"
+					dbOp.Resource_id = "test-resource-id"
+					dbOp.Resource_type = ""
+					return nil
```
jparsai, line 2886: Mocking is working fine here, but am curious why mocking is required while you can create physical entry in DB? I understand mocking is done to simulate errors, but here you are not expecting error. 
Am I missing something?

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -2842,6 +2845,121 @@ var _ = Describe("Operation Controller", func() {
...
+				})
+
+				newEvent := operationEventLoopEvent{
+					request: newRequest(operationNamespace, operationName),
+					client:  mockClient,
+				}
+				result, err := getDBOperationForEvent(ctx, newEvent, mockDB, logger)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result).ToNot(BeNil())
```
jparsai, line 2895: I think it would be better to verify if result object has expected data in fields.

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
```diff
@@ -2842,6 +2845,121 @@ var _ = Describe("Operation Controller", func() {
...
+			})
+
+			It("should handle error when fetching CR", func() {
+				objectKey := client.ObjectKey{Name: operationName, Namespace: operationNamespace}
+				notFoundError := apierr.NewNotFound(schema.GroupResource{}, operationName)
+
+				// Set up the expectation for the Get call to return an error
+				mockClient.EXPECT().Get(gomock.Any(), objectKey, gomock.Any()).Return(notFoundError)
```
jparsai, line 2903: Cant "NotFoundError" be produced by calling function without creating CR? Again mocking is also doing the same job, but I am unable to understand why mocking is done when error can be produced without it?

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
Rizwana777: Yes, right, we can create physical entry in DB

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
Rizwana777: Agreed, updated this

Rizwana777:
@jparsai Updated changes, PTAL

cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
jparsai: No, I didn't ask you to remove it yet, just wanted to understand the reason for using this approach.

jparsai:
> @jparsai Updated changes, PTAL

You didn't have to remove mocking yet, just want to know the reason, maybe I am missing some discussion happened in stand up for using mocks instead of using direct approach.

jgwest:
FYI @Rizwana777 @jparsai 

Great question, I can definitely see both sides of the argument here. :thinking: 

The log mock is definitely fine to use, but for the database mock, @Rizwana777, what do you think: how difficult would it be to handle this without mocking the DatabaseQueries? My only concern is that the database interface is very large, and is updated frequently, and we don't yet have https://issues.redhat.com/browse/GITOPSRVCE-732 in place to automate the mocking. (I suspect it is inevitable that we will have a DatabaseQueries mock, but we may not be at that bridge yet.)

So WDYT Rizwana, is it reasonable to test the function without mock, or is it not realistic?

Rizwana777:
> FYI @Rizwana777 @jparsai
> 
> Great question, I can definitely see both sides of the argument here. 🤔
> 
> The log mock is definitely fine to use, but for the database mock, @Rizwana777, what do you think: how difficult would it be to handle this without mocking the DatabaseQueries? My only concern is that the database interface is very large, and is updated frequently, and we don't yet have https://issues.redhat.com/browse/GITOPSRVCE-732 in place to automate the mocking. (I suspect it is inevitable that we will have a DatabaseQueries mock, but we may not be at that bridge yet.)
> 
> So WDYT Rizwana, is it reasonable to test the function without mock, or is it not realistic?

There are 2 tests which we cannot cover without using mocks, that is  `It("should handle error when fetching DB", func()`  and `It("should handle error when db values are empty", func()` this ```
if db.IsResultNotFoundError(err) { this part cannot be tested without mocks
				return taskCompleteTrue, nil
```

jgwest (changes requested):
Sounds good, thanks for investigating @Rizwana777 ! Let's keep the mocks in, then. I think they will be useful, and once [GITOPSRVCE-732](https://issues.redhat.com/browse/GITOPSRVCE-732) is done, we should have an easy way to regenerate them.

One remaining request: 
- can we moved the mock files (log_sink_mock.go, mock_client.go, and mock_database_queries.bo) into the `backend-shared` component? For example into `backend-shared/util/mocks` (new directory). That way, these mocks can be used by unit tests in any component, rather than just in the `cluster-agent` component.

Once that's done, PR is good to go!

Rizwana777:
Hi @jgwest Moved mocks from cluster-agent to backend-shared and removed log_sink_mock.go file as it was already present in util/log/mocks files, so imported log/mocks in cluster-agent to use log sink mocks and also manifests were failing in CI , so I ran `make generate-manifests ` still it is failing and Unit test is failing in backend which is not related to my changes or may be related in one or other way(I didn't found any possibility here) , PTAL

Rizwana777:
> Hi @jgwest Moved mocks from cluster-agent to backend-shared and removed log_sink_mock.go file as it was already present in util/log/mocks files, so imported log/mocks in cluster-agent to use log sink mocks and also manifests were failing in CI , so I ran `make generate-manifests ` still it is failing and Unit test is failing in backend which is not related to my changes or may be related in one or other way(I didn't found any possibility here) , PTAL

the unit test failing in this PR as well https://github.com/redhat-appstudio/managed-gitops/pull/643

Rizwana777:
> > Hi @jgwest Moved mocks from cluster-agent to backend-shared and removed log_sink_mock.go file as it was already present in util/log/mocks files, so imported log/mocks in cluster-agent to use log sink mocks and also manifests were failing in CI , so I ran `make generate-manifests ` still it is failing and Unit test is failing in backend which is not related to my changes or may be related in one or other way(I didn't found any possibility here) , PTAL
> 
> the unit test failing in this PR as well #643

The above mentioned checks are passing now and PR is ready for re review

jgwest (approved):
LGTM, thanks again @Rizwana777!

PR #622, opened 2023-08-22
Add unit tests for the missing List DB queries

chetan-rns:
#### Description:
- Add unit tests for the missing list functions

https://issues.redhat.com/browse/GITOPSRVCE-692

### DIFF
--- a/backend-shared/db/app_project_managed_env_test.go
+++ b/backend-shared/db/app_project_managed_env_test.go
@@ -80,7 +80,7 @@ var _ = Describe("AppProjectManagedEnvironment Test", func() {
 		It("should return 0 if there are no clusterusers", func() {
 			appProjEnv := &db.AppProjectManagedEnvironment{}
 			rows, err := dbq.CountAppProjectManagedEnvironmentByClusterUserID(ctx, appProjEnv)
-			Expect(err).To(BeNil())
+			Expect(err).ToNot(HaveOccurred())
 			Expect(rows).To(BeZero())
 		})
 
@@ -139,9 +139,16 @@ var _ = Describe("AppProjectManagedEnvironment Test", func() {
 			appProjectManagedEnvs := []db.AppProjectManagedEnvironment{}
 			err = dbq.ListAppProjectManagedEnvironmentByClusterUserId(ctx, clusterUser.Clusteruser_id, &appProjectManagedEnvs)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(len(appProjectManagedEnvs)).To(Equal(1))
+			Expect(appProjectManagedEnvs).To(HaveLen(1))
 			Expect(appProjectManagedEnvs[0]).To(Equal(appProjectManagedEnv))
 
+		})
+
+		It("should return an error if an empty clusterID is passed", func() {
+			appProjectManagedEnvs := []db.AppProjectManagedEnvironment{}
+			err := dbq.ListAppProjectManagedEnvironmentByClusterUserId(ctx, "", &appProjectManagedEnvs)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal("primary key is empty"))
 		})
 
 		It("should return an empty slice if there are no AppProjectManagedEnvironment with the given clusterUserID", func() {
--- a/backend-shared/db/app_project_managed_env_test.go
+++ b/backend-shared/db/app_project_managed_env_test.go
@@ -11,15 +11,25 @@ import (
 
 var _ = Describe("AppProjectManagedEnvironment Test", func() {
 	var seq = 101
-	It("Should Create, Get, Update and Delete an AppProjectManagedEnvironment", func() {
+
+	var (
+		ctx context.Context
+		dbq db.AllDatabaseQueries
+	)
+	BeforeEach(func() {
 		err := db.SetupForTestingDBGinkgo()
 		Expect(err).ToNot(HaveOccurred())
 
-		ctx := context.Background()
-		dbq, err := db.NewUnsafePostgresDBQueries(true, true)
+		ctx = context.Background()
+		dbq, err = db.NewUnsafePostgresDBQueries(true, true)
 		Expect(err).ToNot(HaveOccurred())
-		defer dbq.CloseDatabase()
+	})
 
+	AfterEach(func() {
+		dbq.CloseDatabase()
+	})
+
+	It("Should Create, Get, Update and Delete an AppProjectManagedEnvironment", func() {
 		_, managedEnvironment, _, _, _, err := db.CreateSampleData(dbq)
 		Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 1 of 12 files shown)


backend-shared/db/app_project_managed_env_test.go
```diff
@@ -66,4 +76,86 @@ var _ = Describe("AppProjectManagedEnvironment Test", func() {
...
+			appProjEnv := &db.AppProjectManagedEnvironment{}
+			rows, err := dbq.CountAppProjectManagedEnvironmentByClusterUserID(ctx, appProjEnv)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(rows).To(BeZero())
+		})
+
+		It("should return the number of rows of AppProjectManagedEnvironment for a given clusteruser", func() {
+			By("add sample AppProjectManagedEnvironment with a given clusteruser")
+			expectedRowCount := 1
+			_, managedEnvironment, _, _, _, err := db.CreateSampleData(dbq)
```
keithchong, line 90: Is it possible to common up line 90 to 108 somehow? It is common in the next test case, and it looks similar to what is in the previous existing test too. (Although, the test case above to test  "return 0" doesn't require this, it might not be applicable to BeforeEach.

backend-shared/db/app_project_managed_env_test.go
```diff
@@ -66,4 +76,86 @@ var _ = Describe("AppProjectManagedEnvironment Test", func() {
...
+			appProjectManagedEnv := db.AppProjectManagedEnvironment{
+				Clusteruser_id:         clusterUser.Clusteruser_id,
+				Managed_environment_id: managedEnvironment.Managedenvironment_id,
+				SeqID:                  int64(seq),
+			}
+
+			err = dbq.CreateAppProjectManagedEnvironment(ctx, &appProjectManagedEnv)
+			Expect(err).ToNot(HaveOccurred())
+
+			appProjectManagedEnvs := []db.AppProjectManagedEnvironment{}
```
keithchong, line 139: Use this instead?  `var appProjectManagedEnvs []db.AppProjectManagedEnvironment`

backend-shared/db/app_project_managed_env_test.go
```diff
@@ -66,4 +76,86 @@ var _ = Describe("AppProjectManagedEnvironment Test", func() {
 
 	})
 
+	Context("Test CountAppProjectManagedEnvironmentByClusterUserID function", func() {
```
keithchong, line 79: Not sure if this is much value other than to pad the coverage stats, I think we could cover the case

```
if err != nil {!
   return fmt.Errorf("unable to retrieve appProjectManagedEnvs with clusteruser_id: %v", err)
}
``` 
with something like (the main difference is the context):

```
		Context("Test error handling in ListAppProjectManagedEnvironmentByClusterUserId function", func() {
			It("should return no AppProjectManagedEnvironment", func() {
				_, managedEnvironment, _, _, _, err := db.CreateSampleData(dbq)
				Expect(err).ToNot(HaveOccurred())

				var clusterUser = &db.ClusterUser{
					Clusteruser_id: "test-user-application",
					User_name:      "test-user-application",
				}
				err = dbq.CreateClusterUser(ctx, clusterUser)
				Expect(err).ToNot(HaveOccurred())

				By("Verify whether AppProjectManagedEnvironment is created")
				appProjectManagedEnv := db.AppProjectManagedEnvironment{
					Clusteruser_id:         clusterUser.Clusteruser_id,
					Managed_environment_id: managedEnvironment.Managedenvironment_id,
					SeqID:                  int64(seq),
				}

				err = dbq.CreateAppProjectManagedEnvironment(ctx, &appProjectManagedEnv)
				Expect(err).ToNot(HaveOccurred())

				var appProjectManagedEnvs []db.AppProjectManagedEnvironment
				deadlineExceededContext, _ := context.WithTimeout(ctx, 1*time.Nanosecond)
				err = dbq.ListAppProjectManagedEnvironmentByClusterUserId(deadlineExceededContext, clusterUser.Clusteruser_id, &appProjectManagedEnvs)
				Expect(err).ToNot(HaveOccurred())
				Expect(appProjectManagedEnvs).To(HaveLen(0))
				Expect(appProjectManagedEnvs[0]).To(Equal(appProjectManagedEnv))

			})
		})

```

backend-shared/db/app_project_repository_test.go
```diff
@@ -106,4 +119,77 @@ var _ = Describe("AppProjectRepository Test", func() {
 
 	})
 
+	Context("Test ListAppProjectRepositoryByClusterUserId function", func() {
```
keithchong, line 122: See my other comment regarding the additional coverage for that if statement.

backend-shared/db/application_test.go
```diff
@@ -145,4 +158,43 @@ var _ = Describe("Application Test", func() {
 		})
 	})
 
+	Context("Test ListApplicationsForManagedEnvironment function", func() {
```
keithchong, line 161: See my earlier comment regarding the coverage of the if statement.  Similar solution should apply here.



```
if err != nil {
  return 0, fmt.Errorf("unable to retrieve applications with managed environment id: %v", err)
}
```

backend-shared/db/application_test.go
```diff
@@ -145,4 +158,43 @@ var _ = Describe("Application Test", func() {
...
+			}
+
+			apps := []db.Application{}
+			count, err := dbq.ListApplicationsForManagedEnvironment(ctx, managedEnvironment.Managedenvironment_id, &apps)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(count).To(Equal(expectedRows))
+		})
+
+		It("should return an error if the ManagedEnvironment ID is empty", func() {
+			apps := []db.Application{}
```
keithchong, line 185: Use instead?:  `var apps []db.Application`

backend-shared/db/application_test.go
```diff
@@ -145,4 +158,43 @@ var _ = Describe("Application Test", func() {
...
+			count, err := dbq.ListApplicationsForManagedEnvironment(ctx, "", &apps)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal("primary key is empty"))
+			Expect(count).To(BeZero())
+		})
+
+		It("should return an empty slice if there are no Applications for a given Environment", func() {
+			apps := []db.Application{}
```
keithchong, line 193: Use instead? var apps []db.Application

backend-shared/db/clusteraccess.go
```diff
@@ -117,8 +117,8 @@ func (dbq *PostgreSQLDatabaseQueries) DeleteClusterAccessById(ctx context.Contex
 
 func (dbq *PostgreSQLDatabaseQueries) ListClusterAccessesByManagedEnvironmentID(ctx context.Context, managedEnvironmentID string, clusterAccesses *[]ClusterAccess) error {
 
-	if err := validateQueryParamsEntity(clusterAccesses, dbq); err != nil {
-		return err
+	if clusterAccesses == nil {
```
keithchong, line 120: What's the reason for changing this?  Will there be a situation when dbq.dbConnection = nil ?

backend-shared/db/app_project_managed_env_test.go
chetan-rns: I've moved it to a separate function and reused it when required.

backend-shared/db/clusteraccess.go
chetan-rns: Thanks. I've reverted the change. I didn't realize we were checking the internal connection field `.dbConnection`. But the original reason I didn't want to use `validateQueryParamsEntity` is because we are converting the clusterAccesses a slice to an interface when we pass it as an argument. Even though the clusterAccesses can be `nil`, when passed to the `validateQueryParamsEntity` the interface will no longer be nil and the validation fails. Because in Go, an interface can be considered as a pair of <type,value> and an interface is `nil` only when both its type and value are `nil`. It doesn't affect the test coverage but we might need to update the function to check if the interface value is nill.

```

func (dbq *PostgreSQLDatabaseQueries) ListClusterAccessesByManagedEnvironmentID(ctx context.Context, managedEnvironmentID string, clusterAccesses *[]ClusterAccess) error {

        // clusterAccesses can be nil here. But in the below function, it is used as an interface 
        // and only the interface value will be nil and interface type still says that is a slice. 
	if err := validateQueryParamsEntity(clusterAccesses, dbq); err != nil {
		return err
	}

```

backend-shared/db/app_project_repository_test.go
```diff
@@ -106,4 +119,84 @@ var _ = Describe("AppProjectRepository Test", func() {
...
+			Expect(appProjectRepositories).To(HaveLen(expectedRows))
+
+			for _, appRepo := range appProjectRepositories {
+				expectedAppRepo, found := appProjectRepoMap[appRepo.AppprojectRepositoryID]
+				Expect(found).To(BeTrue())
+				Expect(expectedAppRepo).To(Equal(appRepo))
+			}
+		})
+
+		It("should return an error if the DB query fails", func() {
```
keithchong, line 180: This looks good

backend-shared/db/application_test.go
```diff
@@ -145,4 +158,51 @@ var _ = Describe("Application Test", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+			}
+
+			apps := []db.Application{}
+			count, err := dbq.ListApplicationsForManagedEnvironment(ctx, managedEnvironment.Managedenvironment_id, &apps)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(count).To(Equal(expectedRows))
+		})
+
+		It("should return an error if the DB query fails", func() {
```
keithchong, line 184: This looks good

keithchong (approved):
LGTM, over to you @jgwest

jgwest (changes requested):
LGTM too, thanks @chetan-rns  and thank @keithchong for reviewing! Hust needs a rebase than go

jgwest (approved):
Thanks again @chetan-rns!

PR #628, opened 2023-08-28
GITOPSRVCE-742: ensure operation is always created when appproject DB is updated

jgwest:
#### Description:
- See parent item for description of the bug
- Updated the GitOpsDeployment/RepositoryCredential reconcilers (runners) to ensure they create an Operation whenever AppProjectRepository is updated
- Adds unit/E2E tests to verify the discovered bug


Rizwana777:
Looks good to me @jgwest

jgwest:
Thanks @Rizwana777!

PR #631, opened 2023-08-28
chore: improve test coverage.

jparsai:
#### Description:
This PR is to cover deleteArgoCDApplicationOfDeletedApplicationRow function, of Cluster-agent component.

#### JIRA Ticket:
https://issues.redhat.com/browse/GITOPSRVCE-703

### DIFF
--- a/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
+++ b/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop_test.go
@@ -37,9 +37,10 @@ import (
 
 var _ = Describe("Operation Controller", func() {
 	const (
-		name      = "operation"
-		dbID      = "databaseID"
-		namespace = "argocd"
+		name                     = "operation"
+		dbID                     = "databaseID"
+		namespace                = "argocd"
+		argoCDResourcesFinalizer = "resources-finalizer.argocd.argoproj.io/background"
 	)
 
 	Context("Managed Environment Unit Tests", func() {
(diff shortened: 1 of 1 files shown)


Rizwana777:
@jparsai I think we can cover log errors as well in deleteArgoCDApplicationOfDeletedApplicationRow function by creating mocks similar to this https://github.com/redhat-appstudio/managed-gitops/pull/612 ?

jparsai:
> @jparsai I think we can cover log errors as well in deleteArgoCDApplicationOfDeletedApplicationRow function by creating mocks similar to this #612 ?

That is a good suggestion @Rizwana777, but unfortunately there will be a lot of effort required, 
for example here https://github.com/redhat-appstudio/managed-gitops/blob/main/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go#L967
log message will be printed only if I get an error, but there is no way I can cause labels.NewRequirement() to fail without mocking this function as well. If I go with mocking function then it has to be done for many other functions as well.

Rizwana777:
> > @jparsai I think we can cover log errors as well in deleteArgoCDApplicationOfDeletedApplicationRow function by creating mocks similar to this #612 ?
> 
> That is a good suggestion @Rizwana777, but unfortunately there will be a lot of effort required, for example here https://github.com/redhat-appstudio/managed-gitops/blob/main/cluster-agent/controllers/managed-gitops/eventloop/operation_event_loop.go#L967 log message will be printed only if I get an error, but there is no way I can cause labels.NewRequirement() to fail without mocking this function as well. If I go with mocking function then it has to be done for many other functions as well.

I agree @jparsai thank you

jgwest (changes requested):
Looks great, one final request: can we also verify that the `AppProject` is deleted, and/or not deleted (exists), for the test cases? (more info on cluster agent appproject behaviour is [here](https://github.com/redhat-appstudio/managed-gitops/blob/main/docs/sandboxing-via-appprojects.md#changes-to-cluster-agent-component) if useful).

jparsai:
> Looks great, one final request: can we also verify that the `AppProject` is deleted, and/or not deleted (exists), for the test cases? (more info on cluster agent appproject behaviour is [here](https://github.com/redhat-appstudio/managed-gitops/blob/main/docs/sandboxing-via-appprojects.md#changes-to-cluster-agent-component) if useful).

It is done, PTAL.

jparsai:
/test managed-gitops-e2e-tests

Failing in `operation_namespace_test.go`, which has no changes in this PR, hence triggering once again.

jgwest (approved):
LGTM, thanks @jparsai!

PR #632, opened 2023-08-28
Fix intermittent binding controller e2e test

chetan-rns:
#### Description:
- The DeploymentTarget reclaimer will unset the claimRef if the specified DTC is not found. In the e2e test, we were creating a fake DT pointing to a DTC which didn't exist and the claimRef field was unset by the reclaimer controller. Hence the binding controller had two matching DTs(the correct DT and the fake DT whose claimRef was unset) available for a single DTC. So, when the binding controller picked the fake DT it resulted in an intermittent failure.
- In order to solve this we create a DTC that binds with the fake DT so that its claimRef field will not be unset.

https://issues.redhat.com/browse/GITOPSRVCE-711


jgwest (approved):
Looks perfect, thanks for investigating @chetan-rns!

PR #633, opened 2023-08-28
RHTAPBUGS-735: Delete orphaned SpaceRequests

jgwest:
#### Description:
- The main fix is the call to `deleteSpaceRequestIfOrphaned`
- Added a new label to Space Request:
    - Currently we have an existing label `appstudio.openshift.io/dtc` which is added to SpaceRequests that are created by the DeploymentTarget controller (and points to DTC).
    - I've added a new (optional) label `appstudio.openshift.io/dt` which is likewise added to SpaceRequest, and which points to the DeploymentTarget (rather than DTC). This makes it easier to locate the DT for a SpaceRequest. However, we should fall back to using  `appstudio.openshift.io/dtc` in cases where `/dt` isn't found.
    - This logic is then used in findMatchingDTforSpaceRequest/ findMatchingSpaceRequestForDT
- Everything else in the PR is just unit/E2E tests or minor refactoring


jgwest:
cluster ci failure
/retest

chetan-rns (commented):
Looks good @jgwest. Just added a question regarding what to do when the DTClass is missing.

jgwest:
Thanks @chetan-rns!

PR #635, opened 2023-08-31
fix: Invalid cluster credentials error is logged with ERROR.

jparsai:
#### Description:
This PR is change log level of message printed when user provided wrong credentials and service is unable to connect with provided cluster.

#### JIRA Ticket:
https://issues.redhat.com/browse/GITOPSRVCE-741


backend/eventloop/workspace_resource_event_loop_test.go
```diff
@@ -21,7 +21,7 @@ var _ = Describe("isManagedEnvironmentConnectionUserError function Test", func()
 			rootErr := fmt.Errorf("Get \"https://api.ireland.burr-on-aws.com:6443/api?timeout=32s\": dial tcp: lookup api.ireland.burr-on-aws.com on 172.30.0.10:53: no such host")
 			err := fmt.Errorf("%s: %w", shared_resource_loop.UnableToCreateRestConfigError, rootErr)
 
-			result := isManagedEnvironmentConnectionUserError(err, log)
+			result := application_event_loop.IsManagedEnvironmentConnectionUserError(err, log)
```
chetan-rns, line 24: Should we move the tests to the same package as the function? It's not significant but makes it easier to find it. Everything else looks great.

jparsai: Ah, yes I forgot to move unit tests after I moved the function definition.

backend/eventloop/application_event_loop/application_event_runner_deployments.go
```diff
@@ -248,6 +249,11 @@ func (a applicationEventLoopRunner_Action) handleNewGitOpsDeplEvent(ctx context.
 		userError := "Unable to reconcile the ManagedEnvironment. Verify that the ManagedEnvironment and Secret are correctly defined, and have valid credentials"
 		devError := fmt.Errorf("unable to get or create managed environment, isworkspacetarget:%v: %v", isWorkspaceTarget, err)
 
+		// If we recognize this error is a connection error due to the user providing us invalid credentials, don't bother to log it.
```
jgwest, line 252: I think this will not only block the log, but also block the update of the `GitOpsDeployment`'s .status field, which is where the error is communicated back to the user? We definitely still want to log the error to the .status field of the GitOpsDeployment

jgwest (changes requested):

backend/eventloop/application_event_loop/application_event_runner_deployments.go
jparsai: Yes it will, I have changed it and now error is ignored after status is updated here:
https://github.com/redhat-appstudio/managed-gitops/pull/635/files#diff-08e1558d59743d0033fba6e63425c93a27f78f971a4dca44b4bd386ae3214f4cR392
PTAL.

jgwest (approved):
LGTM, thanks @jparsai!

PR #638, opened 2023-09-06
Recreate missing Secrets for ManagedEnvironments.

jparsai:
## Description:
This PR is to add a logic to recreate Secrets in Argo CD namespace if they are required by ManagedEnvironments but missing in Cluster for some reason.

## JIRA Ticket:
https://issues.redhat.com/browse/GITOPSRVCE-739

### DIFF
--- a/cluster-agent/controllers/argoproj.io/namespace_reconciler_test.go
+++ b/cluster-agent/controllers/argoproj.io/namespace_reconciler_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/redhat-appstudio/managed-gitops/backend-shared/db"
 	dbutil "github.com/redhat-appstudio/managed-gitops/backend-shared/db/util"
 	sharedutil "github.com/redhat-appstudio/managed-gitops/backend-shared/util"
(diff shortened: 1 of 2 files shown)


cluster-agent/controllers/argoproj.io/namespace_reconciler_test.go
```diff
@@ -748,4 +749,211 @@ var _ = Describe("Namespace Reconciler Tests.", func() {
...
+					Namespace: gitopsEngineInstance.Namespace_name,
+				},
+			}
+		})
+
+		It("Should not create Operation for Secret, even if it is missing in Cluster, since ManagedEnvironment is created recently.", func() {
+
+			defer dbq.CloseDatabase()
+
+			By("Create Secret CR in cluster.")
```
jopit, line 872: Shouldn't we not create the secret on the cluster, since from the description of the test, it should be missing?

jopit (changes requested):

jgwest (approved):
LGTM, thanks @jparsai, and thanks @jopit for reviewing!

PR #651, opened 2023-09-20
Refactor log statements to better use log context, reduce verbosity

jgwest:
#### Description:
- Move some INFO-level log statements down to DEBUG, to reduce verbosity
- Move log text to constants to in log util package
- Remove spaces from log constant names
- Remove log context from log statements when that same log context is already being logged as part of the log object via `log.WithValues(...)`
- Change the param order of `deleteDbEntry`, to move the least specific params (dbqueries, log) to the end of the function signature


jgwest:
CI failure
/retest

jparsai:
LGTM @jgwest 
/lgtm

PR #654, opened 2023-09-22
Restrict repo server parallelism and git/kustomize run time

jgwest:
#### Description:
- Limit the number of parallel `git clone`/`kustomize` calls that can be running at any given time, so as to avoid OOM
- Limit `git clone`/`kustomize` calls to 30 seconds
- Raise the memory limit as we are close in some scenarios


chetan-rns (approved):
Thanks @jgwest 

/lgtm

jgwest:
CI failure
/retest

jgwest:
Thanks @chetan-rns!

PR #658, opened 2023-09-26
fix: Change naming convention for RepoCred secret created in Argo CD namespace.

jparsai:
#### Description:
This PR is to change naming convention for Secrets created for Repository Credential in Argo CD namespace.

#### JIRA Ticket:
https://issues.redhat.com/browse/GITOPSRVCE-755


jgwest (changes requested):
LGTM, thanks @jparsai! One small ask, can you also ensure we have E2E test that verifies that the Argo CD repository secret is created, similar to what the unit test is doing in the PR?

jgwest (approved):
LGTM, thanks again @jparsai

PR #660, opened 2023-10-02
Delete SEB when associated Application is deleted

jopit:
#### Description:
- Delete the SnapshotEnvironmentBinding when the associated application is deleted.  Add unit and e2e tests for this.  Update existing tests which used to set the .spec.application field in the SEB without creating an application.


jgwest (approved):
Looks perfect, thanks @jopit, and thanks @Rizwana777 for reviewing!

PR #661, opened 2023-10-03
Enable AppProject isolation for Makefile run targets

jopit:
#### Description:
- A change was made recently (PR #643) to enable AppProject isolation for `make test-e2e`.  But it was not enabled for the run targets in the various Makefiles, so running locally using `make start-e2e` would cause some of the tests to fail


jgwest (approved):
Looks like I missed this when I updated the other locations, thanks for investigating/fixing @jopit!

PR #663, opened 2023-10-06
Only call k8s client to update SnapshotEnvironmentBinding status when .status field has changed

jgwest:
#### Description:
- I happened to notice that on RHTAP cluster we were making way too many status updates to `SnapshotEnvironmentBinding` CR, despite no obvious changes. 
    - This code change should help reduce that, by only updating the status when it has changed, and using a more deterministic generation algorithm that does not rely on Go map key ordering
- Only call `k8sClient.Update` to update the `SnapshotEnvironmentBinding` status IF the contents of the .status field actually changed (otherwise skip)
- Process the component names in alphabetical order, so that the generated status is consistent (deterministic) from run to run
- Misc tweaks:
    - Tweak the pointer usage on `SnapshotEnvironmentBinding`
    - `Can not` -> `Cannot`

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/snapshotenvironmentbinding_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jopit (changes requested):
Looks good.  Maybe it would make sense to add a unit test to ensure that the status is indeed in lexical order?

jgwest:
Thanks @jopit!

PR #665, opened 2023-10-11
Fix failing of DT e2e test

Rizwana777:
#### Description:
- Fixed failing of e2e test by adding time interval in Eventually block


jgwest (approved):
LGTM, thanks @Rizwana777!

PR #676, opened 2023-10-12
GITOPSRVCE-767: fix interimittent failure in workspace_resource_event_loop_test.go

jgwest:
#### Description:
- Intermittent, a different a error string is returned, due to where the context cancelled is recognized. 
- This PR updates the error string to handle both cases.


chetan-rns (approved):
Thanks @jgwest 

/lgtm

PR #678, opened 2023-10-12
GITOPSRVCE-772: fix intermittent E2E test failure in gitopsdeployment_test.go

jgwest:
#### Description:
- Based on the failure logs, the error appears to be that the order is different between the expect and actual values
- So just compare in any order


keithchong (approved):
LGTM. 

Comparing arrays whose elements might not be in the same order.

PR #679, opened 2023-10-13
GITOPSRVCE-771: re-enable environment webhook

jgwest:
#### Description:
- Re-enable environment webhook after it was previously disabled, see 771 for details.


jgwest:
YAML change only.

PR #685, opened 2023-10-18
GITOPSRVCE-723: Better secure dev postgresql instance

jgwest:
#### Description:
- Postgresql dev container should not run as root, and should use a readonly FS


chetan-rns (approved):
/lgtm

PR #690, opened 2023-10-23
Fix invalid SEVERE log entry by only checking field (rather than pointer) of event.Message.Event in app event loop

jgwest:
#### Description:
- As part of [GITOPSRVCE-704](https://issues.redhat.com//browse/GITOPSRVCE-704) I converted some pointer fields to objects, however there were a couple SEVERE error logs that were checking the pointer.
- Fix invalid SEVERE log entry by only checking field (rather than pointer) of event.Message.Event in app event loop


jgwest:
Merging w/o reviews as this is a simple log fix, and this is currently blocking release to staging/prod.

PR #691, opened 2023-10-23
GITOPSRVCE-783: use Component CR .metadata.Name, not ComponentName

jgwest:
#### Description:
- As discussed on Slack (see link in JIRA), when we reference a Component CR, we should use the `.metadata.Name` of the component, rather than the `.spec.ComponentName`


chetan-rns (approved):
Thanks @jgwest 

/lgtm

PR #692, opened 2023-10-24
Fix log keys in application_event_loop.go

jgwest:
#### Description:
- Log statement keys on https://github.com/redhat-appstudio/managed-gitops/pull/690/files were incorrectly paired


jgwest:
Merging w/o reviews as this is a log statement change only, and this is currently blocking deployment to staging/prod.

PR #693, opened 2023-10-24
GITOPSRVCE-771: remove temporary overlays added for debugging

jgwest:
#### Description:
- As part of 771 I added temporary overlays to allow us to debug Environment.
- This removes that temporary overlay, as it is no longer needed.


jgwest:
Merging w/o reviews, as it's a YAML cleanup change only.

PR #694, opened 2023-10-24
GITOPSRVCE-776: dont re-reconcile on user errors

jgwest:
#### Description:

- When a user specifies an invalid value in a `GitOpsDeploymentManagedEnvironment` -- for example, they reference a Secret that doesn't exist -- we don't need to keep re-reconciling that resource.
- In this case, the user should fix the `GitOpsDeploymentManagedEnvironment` -- by adding a proper Secret -- before we reconcile that resource again.
- These types of errors are 'user errors': cases where a user is specifying invalid API input, and there is no reason for us to keep trying to reconcile the resource.

- So, as part of this PR, the `ReconcileSharedManagedEnv` will now detect user errors, and if found, return true.
- Then in the workspace resource event loop, we stop reconciling when a user error is detected.

### DIFF
--- a/backend/eventloop/shared_resource_loop/sharedresourceloop.go
+++ b/backend/eventloop/shared_resource_loop/sharedresourceloop.go
@@ -313,7 +313,7 @@ type sharedResourceLoopMessage_getOrCreateSharedResourcesResponse struct {
 	// A user error is case where the user has specified an invalid value in the GitOpsDeploymentManagedEnvironment, or the Secret
 	//
 	// - An example of a user error: user specified a Secret (in the GitOpsDeploymentManagedEnvironment) that doesn't exist.
-	// - An exampe of a non-user error: unable to connect to the database
+	// - An example of a non-user error: unable to connect to the database
 	//
 	// We do not need to continue to reconcile a resource that has a user error: a fix is required to the resource (for example, creating a missing Secret)
 	isUserError bool
--- a/backend/eventloop/shared_resource_loop/sharedresourceloop.go
+++ b/backend/eventloop/shared_resource_loop/sharedresourceloop.go
@@ -163,16 +163,20 @@ func (srEventLoop *SharedResourceEventLoop) GetGitopsEngineInstanceById(ctx cont
 
 // Ensure the user's workspace is configured, ensure a GitOpsEngineInstance exists that will target it, and ensure
 // a cluster access exists the give the user permission to target them from the engine.
+// Return values:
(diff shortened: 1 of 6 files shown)


backend/eventloop/shared_resource_loop/sharedresourceloop.go
```diff
@@ -301,7 +305,19 @@ type sharedResourceLoopMessage_getOrCreateSharedResourceManagedEnvRequest struct
...
-	err               error
+
+	// err is whether an error occurred while reconciling the managed env
+	err error
+
+	// if 'err' is non-nil, the 'isUserError' field will indicate whether the error is a user error.
+	// A user error is case where the user has specified an invalid value in the GitOpsDeploymentManagedEnvironment, or the Secret
+	//
+	// - An example of a user error: user specified a Secret (in the GitOpsDeploymentManagedEnvironment) that doesn't exist.
+	// - An exampe of a non-user error: unable to connect to the database
```
keithchong, line 316: example

backend/eventloop/workspace_resource_event_loop.go
```diff
@@ -288,11 +293,20 @@ func handleResourceLoopManagedEnvironment(ctx context.Context, msg workspaceReso
...
 		false, k8sClientFactory, log)
 
 	if err != nil {
-		return retry, fmt.Errorf("unable to reconcile shared managed env: %v", err)
+
+		if isUserErr {
+			log.Info("user error: user specified invalid managed environment parameters", "error", err)
+			// A user error indicates that the user specified invalid parameters, for example, they specified a Secret that doesn't exist.
+			// We do not need to retry in this case, as the user needs to make a change before their resource is valid.
+			return noRetry, nil
```
keithchong, line 305: Question, at this point, we add a message to the log that it is a user error, and then we return false and nil here.  Should we propagate this logic up further to the task level to explain why there was no retry?  At that point, it is treated as though there was no error so there is no need to retry the task.

jgwest: @keithchong My thinking was that the log was sufficient in this case: at this point we know it's just a user error (missing Secret, etc), and so there is not much benefit to continuing to log the issue in additional places in the code. 

In these cases, we need the user to fix the resource (create/update the Secret, etc), which will trigger a Re-reconcile, and cause the code to run again.

keithchong: Ok, seems reasonable.

jgwest:
Thanks @keithchong!

PR #695, opened 2023-10-25
Move up to latest kube-rbac-proxy, and add -http2-disable=true

jgwest:
#### Description:
- In response to rapid reset CVE, move up to latest kube-rbac-proxy, and add -http2-disable=true


jgwest:
Merging w/o reviews as tests as passing, YAML changes only.

PR #696, opened 2023-10-25
GITOPSRVCE-731: Add Amazon CA root certs, re-enable TLS verification

jgwest:
#### Description:
- Automatically acquire the latest Amazon CA root certs, and add them to the system store within the container
- Re-enable TLS verification, to valid RDS instances using those certificates


jgwest:
Merging w/o reviews so I can try it out on staging.

PR #698, opened 2023-10-25
Refactor cluster-agent namespace reconciler to return first error after processing resources, then verify no errors occur in unit tests

jgwest:
#### Description:
- If an error occurs in one of the cluster-agent namespace reconciler functions, continue to iterate through the remaining items in the list, but return the first error that occurred.
- Then, in the unit test, verify that no errors occurred.
- This keeps the behaviour of the cluster-agent namespace reconciler the same as before, but allows unit tests to catch any errors that occur.
- This will allow our unit tests to better catch any new errors that occur.


jgwest:
Thanks @jparsai!

PR #706, opened 2023-10-30
Refactor backend db reconciler to return first error after processing resources, then verify no errors occur in unit tests

jgwest:
#### Description:
- Similar to https://github.com/redhat-appstudio/managed-gitops/pull/698, we want to avoid burying errors (not returning an error whenever an error occurs)
- In most case, even after an error occurs, we still want to keep processing the database resources, but, we should still return the _first_ error that occured
- This allows us to verify from the unit tests that no errors are occurring

### DIFF
--- a/backend/eventloop/db_reconciler.go
+++ b/backend/eventloop/db_reconciler.go
@@ -75,25 +75,39 @@ func (r *DatabaseReconciler) startTimerForNextCycle(ctx context.Context, databas
 		_, _ = sharedutil.CatchPanic(func() error {
 
 			// Clean orphaned entries from DTAM table and other table they relate to (i.e ApplicationState, Application).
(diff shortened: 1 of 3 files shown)


jparsai:
LGTM
@jgwest should I `unhold` it?

jgwest:
Don't worry @jparsai,  I'll un`/hold` it.

 I use `/hold` for my PRs (and only mine), so I can control when they get merged (which sometimes effects when they get released to infra-deployment and therefore RHTAP staging/prod). 

For my PRs, as soon as someone `/lgtm`s the PT, then the Tide bot will automatically merge it. In some cases, I don't want tide to automatically merge it, so I add `/hold` to prevent Tide from automerging. 

For other folks' PRs, Tide won't automatically merge, so I don't need to add it there. It just affects my PRs. :shrug:

jgwest:
And thanks for reviewing @jparsai!

PR #707, opened 2023-10-30
GITOPSRVCE-769: Switch to namespace-scoped gitops-service-argocd

jgwest:
#### Description:
- Platform team would like use to avoid using very permissive cluster roles in our `gitops-service-argocd` instance, so the easiest way to do that is to move GitOps Service' Argo CD instance to be strictly namespace-scoped.

### DIFF
--- a/manifests/base/gitops-service-argocd/base/argo-cd-application-controller-clusterrolebinding.yaml
+++ b/manifests/base/gitops-service-argocd/base/argo-cd-application-controller-clusterrolebinding.yaml
@@ -1,12 +1,15 @@
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: appstudio-gitops-service-argocd-argocd-application-controller
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: appstudio-gitops-service-argocd-argocd-application-controller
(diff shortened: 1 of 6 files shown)


manifests/base/gitops-service-argocd/base/argo-cd-application-controller-clusterrolebinding.yaml
```diff
@@ -1,12 +1,15 @@
...
-  kind: ClusterRole
-  name: appstudio-gitops-service-argocd-argocd-application-controller
-subjects:
-  - kind: ServiceAccount
-    name: gitops-service-argocd-argocd-application-controller
-    namespace: gitops-service-argocd
\ No newline at end of file
+
+# To enable 'gitops-service-argocd' instance as cluster-scoped Argo CD, uncomment the lines below.
+
```
chetan-rns, line 3: Can we also remove the ClusterRoles since they will not be used with the namespace-scoped instance? We can either comment the resource or its reference in the kustomization.yaml

 https://github.com/redhat-appstudio/managed-gitops/blob/main/manifests/base/gitops-service-argocd/base/kustomization.yaml

jgwest: Done, thanks @chetan-rns!

PR #708, opened 2023-10-31
GITOPSRVCE-787: wait X minutes before deleting orphaned SEB

jgwest:
#### Description:

There is a theoretical race condition (not seen in practice, to my knowledge) where a SnapshotEnvironmentBinding (SEB) is reconciled before the Application exists, which causes the SEB to be deleted by our orphaned SEB detection login.

For example:
1) SEB is created
2) SEB is reconciled. SEB `Reconcile()` logic detects that the parent Application doesn't exist, and thus the Reconcile logic will automatically delete the SEB.
3) Application is created.

You could imagine a case like this might happen if you were kubectl apply-ing a bunch of resources at once, and the SnapshotEnvironmentBinding was reconciled before the Application was created.

In any case, the fix is just to wait a few minutes (via requeue) before deletion.


chetan-rns (approved):
Thanks @jgwest 
/lgtm

PR #711, opened 2023-11-06
GITOPSRVCE-788: reconcile changes to repository Secrets

jgwest:
#### Description:
- At present, the GitOpsRepositoryCredential reconciler is not watching Secret resources for changes. 
- This means that if a user changes the private Git repository credentials defined in a Secret (for example, changing their password), this will not be detected by the GitOpsRepositoryCredential reconciler.
- The fix is to add a watch for Secret resources to the controller, plus unit/E2E tests to verify

- Other PR Changes:
    - Fixed a bug where Operation was created in the Namespace of the GitOpsDeploymentReploymentCredential, rather than in the Argo CD namespace
    - Add a new function interface which allow us to mock the validation of Git repos from unit tests
    - Fix constant typos
    - Improved pointer hygiene
    - Change Secret watch type from `predicate.GenerationChangedPredicate` to `predicate.ResourceVersionChangedPredicate`, since Secrets don't have a generation
    - Refactor function signatures parameter order, so that `context` is first (which is a K8s pattern), followed by parameters relevant to what the function is doing, followed by generic parameters (database obj, log, etc). In general, parameters should be ordered in most important to least important.
    -  Move `managed-gitops.redhat.com/*` strings into a constant, and update references
    - Update E2E tests, and allow them to run if not all private repo credentials are defined


jgwest:
Thanks @Rizwana777!

PR #715, opened 2023-11-15
GITOPSRVCE-792: Light refactoring of DT/DTC/DTClaim controllers

jgwest:
#### Description:
- The DeploymentTarget/Claim/Class controller implementations are a bit tangled together, due to them being implemented by 3 different people across 2 different teams.
- This unifies the implementations by ensuring that only one controller is reconciling a particular CR at a time, and also renames filenames/struct names, to be consistent with which controller is doing what.

#### Notes for review:
- There are/should be no logic changes: just moving around code.
    - A couple controllers were both reconciling on the same CR, I merged them together and removed the old code (these are the deleted files, including `*predicates*.go`)
    - In a few cases I cleaned up the code by moving large blocks of code into their own function


chetan-rns (approved):
Looks great. Thank you @jgwest 

/lgtm

PR #722, opened 2023-11-22
Use updated Environment API .

jparsai:
## Description
This PR is to use latest version of of v1alha1 Environment API which will be updated by [this PR](https://github.com/redhat-appstudio/application-api/pull/79). 
Now managed-gitops controllers and tests will use `Target` field instead of `UnstableConfigurationFields` and `Target.Claim.DeploymentTargetClaim.ClaimName` instead of `Configuration.Target.DeploymentTargetClaim.ClaimName`.
Also there are few changes that were suggested by SonarCloud.

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-664

### DIFF
--- a/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
+++ b/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go
@@ -189,7 +189,7 @@ func reconcileEnvironment(ctx context.Context, environment appstudioshared.Envir
 		errorConditionSet_false = false
 	)
 
-	if environment.GetDeploymentTargetClaimName() != "" && environment.Spec.UnstableConfigurationFields != nil {
+	if environment.Spec.Target != nil && !reflect.ValueOf(environment.Spec.Target.KubernetesClusterCredentials).IsZero() && environment.GetDeploymentTargetClaimNamev2() != "" {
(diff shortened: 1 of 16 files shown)


jgwest (commented):
Since we no longer have the ability to use conversion webhooks within RHTAP, this means that changes that we make to the GitOps Service API need to be backwards compatible. This means:
- Backwards compatibility in the API (e.g. we don't remove old fields)
- Backwards compatibility in the behaviour (e.g. we don't break existing controllers which rely on existing fields).

We thus need to support consumers of both the old and new fields of the Environment API.

So, all changes we are making to GitOps Service' Environment logic need to be backwards compatible. That means that if another RHTAP service is using the old API, our service will continue to work for them (will maintain the previous behaviour that that service expected.)

application-api/79 - is backward compatible, AFAICT.	

HOWEVER, the changes made as part of this PR (correct me if I'm wrong!) appear to change (break) the behaviour of our existing API, and thus are not backwards compatible.
- If this PR were to be merged, the existing RHTAP components would need to be updated to use the new API, and those components would be broken until they did.

Since the GitOps Service within internal RHTAP is likely deprecated, and since the changes required to this PR to allow it to work with both old and new versions of the API are significant, I recommend we close this PR and close the story as 'won't do' (rather than updating this PR to maintain backwards compatibility). WDYT?

PR #726, opened 2023-12-12
Error reporting in DeploymentTarget/DeploymentTargetClaim.

jparsai:
## Description
This PR is to add logic in DeploymentTarget and DeploymentTargetClaim controllers to update new `Conditions` filed in `status`.

## Jira Ticket
https://issues.redhat.com/browse/GITOPSRVCE-627


jgwest:
I pushed a new commit which:
- Refactor the logic of the controller to update the status condition via a condition returned from child function
- Removal of status condition update in a few cases where I think it isn't needed (internal errors)
- Condition status of true -> condition status of false, in a few cases

After making these changes, the E2E and unit tests pass (at least locally)

jgwest (approved):
LGTM, thanks @jparsai!

